PR #5, opened 2022-04-18
Initial draft specification for age plugins

str4d:
Migrated from https://hackmd.io/@str4d/age-plugin-spec so this is a relatively complete draft.

### DIFF
--- a/age-plugin.md
+++ b/age-plugin.md
@@ -126,9 +126,9 @@ notations (optional fields indicated with `[brackets]`):
 - `(COMMAND, [METADATA] STANZA...)` - `STANZA` is a complete stanza.
 - `(COMMAND)` - a command with no metadata or data.
 
-Stanzas are serialized exactly as in age headers, using the explicit encoding
-of stanza bodies. For example, if three commands are sent, and the second
-command has no associated data:
+Stanzas are serialized exactly as in the [v1 age header][], using the explicit
+encoding of stanza bodies. For example, if three commands are sent, and the
+second command has no associated data:
 ```
 -> command-1 metadata
 Base64(data)
@@ -548,3 +548,4 @@ this is to have a 1:1 relationship between plugins and recipient types.
   various platforms, as well as be safe across OS upgrades.
 
 [RFC 4648]: https://www.rfc-editor.org/rfc/rfc4648.html
+[v1 age header]: https://c2sp.org/age#abnf-definition-of-file-header
--- a/age-plugin.md
+++ b/age-plugin.md
@@ -1,5 +1,3 @@
-{%hackmd theme-dark %}
-
 # Plugin system for age
 
 ## Introduction
--- a/age-plugin.md
+++ b/age-plugin.md
@@ -1,0 +1,541 @@
+{%hackmd theme-dark %}
+
+# Plugin system for age
+
+## Introduction
+
+The age file encryption format follows the "one well-oiled joint" design
+philosophy. The mechanism for extensibility (within a particular format version)
+is the recipient stanzas within the age header: file keys can be wrapped in any
+number of ways, and age clients are required to ignore stanzas that they do not
+understand. The core APIs that exercise this mechanism are:
+
+- A recipient that wraps a file key and returns a stanza.
+- An identity that unwraps a stanza and returns a file key.
+
+Custom age clients can bundle support for the exact recipient or identity types
+required for a particular environment. However, a general plugin system will
+enable an ecosystem of common third-party recipient types to grow.
+
+The plugin system specified in this document provides a mechanism for exposing
+the core APIs across process boundaries. It has two main components:
+
+- A map from recipients and identities to plugin binaries.
+- State machines for wrapping and unwrapping file keys.
+
+With this composable design, developers can implement a recipient or identity
+that they would use directly with an age library, and then also deploy it as a
+plugin binary.
+
+## Mapping recipients and identities to plugin binaries
+
+age plugins are identified by an arbitrary case-insensitive string `NAME`. This
+string is used in three places:
+
+- Plugin-compatible recipients are encoded using Bech32 with the HRP `age1name`
+  (lowercase).
+- Plugin-compatible identities are encoded using Bech32 with the HRP
+  `AGE-PLUGIN-NAME-` (uppercase).
+- Plugin binaries (to be started by age clients) are named `age-plugin-name`.
+
+Users interact with age clients by providing either recipients for file
+encryption, or identities for file decryption. When a plugin recipient or
+identity is provided, the age client searches the `PATH` for a binary with the
+corresponding plugin name.
+
+Recipient stanza types are not required to be correlated to specific plugin
+names. When decrypting, age clients will pass all recipient stanzas to every
+connected plugin. Plugins MUST ignore stanzas that they do not know about.
+
+A plugin binary may handle multiple recipient or identity types by being present
+in the `PATH` under multiple names. This can be implemented with symlinks or
+aliases to the canonical binary.
+
+Multiple plugin binaries can support the same recipient and identity types; the
+first binary found in the `PATH` will be used by age clients. Some Unix OSs
+support "alternatives", which plugin binaries should leverage if they provide
+support for a common recipient or identity type.
+
+Note that the identity specified by a user doesn't need to point to a specific
+decryption key, or indeed contain any key material at all. It only needs to
+contain sufficient information for the plugin to locate the necessary key
+material.
+
+### Standard age keys
+
+A plugin MAY support decrypting files encrypted to native age recipients, by
+including support for the `x25519` recipient stanza. Such plugins will pick
+their own name, and users will use identity files containing identities that
+specify that plugin name.
+
+### Agents for age identities
+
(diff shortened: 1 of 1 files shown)


FiloSottile:
TODO: add a note about not implicitly executing plugins from the local directory. See https://go.dev/blog/path-security.

age-plugin.md
```diff
@@ -0,0 +1,541 @@
...
+These may be interleaved with other requests, and the client must cache the
+unwrapped file keys until this phase is complete. The plugin sends no more than
+one file key per file index in the original set of stanzas.
+
+- TODO: Should IPC errors (e.g. invalid command structure or argument types) be
+  separated from state machine errors (e.g. invalid identities)?
+  - Probably not; it should be possible to run the IPC protocol over e.g. TCP.
+- TODO: What about identities that are not unknown or invalid, but for which we
+  get an error while trying to use them? (e.g. YubiKey missing, invalid PIN,
+  cloud KMS rejection)
```
FiloSottile, line 432: Some of these errors, such as the user deciding to skip the YubiKey through a `confirm` dialog, or a KMS plugin that encounters a transient error but wants to give a chance to a fallback identity, should be soft errors that let the client continue on to the next identity. The plugin can implement these by optionally sending a `msg` and then returning without sending anything, like when no stanzas are expected to be unwrappable, as described above.

age-plugin.md
```diff
@@ -0,0 +1,541 @@
...
+
+The following commands and responses are defined for this phase:
+
+- `(msg; MESSAGE)` - a message that should be displayed to the user, for
+  example to prompt them to interact with a hardware key.
+  - Response is `(ok)` (if the message can be displayed) or `(fail)` (if, for
+    example, there is no UI for displaying messages).
+  - The response MAY be sent by the client before the message has been displayed
+    to the user.
+- `(confirm, Base64(YES_STRING) [Base64(NO_STRING)]; MESSAGE)` - a request for
```
FiloSottile, line 236: Define Base64 as the standard alphabet and no padding.

age-plugin.md
```diff
@@ -0,0 +1,541 @@
...
+identity is provided, the age client searches the `PATH` for a binary with the
+corresponding plugin name.
+
+Recipient stanza types are not required to be correlated to specific plugin
+names. When decrypting, age clients will pass all recipient stanzas to every
+connected plugin. Plugins MUST ignore stanzas that they do not know about.
+
+A plugin binary may handle multiple recipient or identity types by being present
+in the `PATH` under multiple names. This can be implemented with symlinks or
+aliases to the canonical binary.
```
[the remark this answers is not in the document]
str4d: This does mean shell aliases.

emilazy:
> A plugin MUST refuse to start if it does not know about the requested state machine

It's unclear to me what "refuse to start" looks like here; does it involve replying with a protocol-structured error, or just unceremoniously dumping whatever to stderr and exiting with an error status code? The text implies the latter, but rage at least reacts to that with a rather unhelpful "Error: incomplete response", though I realize this could be an implementation infelicity.

age-plugin.md
```diff
@@ -0,0 +1,541 @@
...
+A plugin MUST refuse to start if it does not know about the requested state
+machine; in this situation, or if a plugin otherwise terminates early with an
+error, the age client:
+
+- MUST propagate the failure to the user if it occurs during file encryption
+  (which would mean that the file could not be encrypted to one of the requested
+  recipients). This may include displaying the contents of the plugin's standard
+  error to the user.
+- MAY ignore the failure if it occurs during file decryption (and try another
+  identity or plugin).
```
emilazy, line 101: It's unclear to me what "refuse to start" looks like here; does it involve replying with a protocol-structured error, or just unceremoniously dumping whatever to stderr and exiting with an error status code? The text implies the latter, but rage at least reacts to that with a rather unhelpful "Error: incomplete response", though I realize this could be an implementation infelicity.

age-plugin.md
```diff
@@ -0,0 +1,541 @@
...
+- A map from recipients and identities to plugin binaries.
+- State machines for wrapping and unwrapping file keys.
+
+With this composable design, developers can implement a recipient or identity
+that they would use directly with an age library, and then also deploy it as a
+plugin binary.
+
+## Mapping recipients and identities to plugin binaries
+
+age plugins are identified by an arbitrary case-insensitive string `NAME`. This
```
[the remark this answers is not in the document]
str4d: "arbitrary string" has a specific meaning here: a string composed of RFC 5234 `VCHAR`. The word "arbitrary" came from the original age design document, but we should clarify it here with ABNF in the same way we now do in https://c2sp.org/age.

age-plugin.md
```diff
@@ -0,0 +1,541 @@
+{%hackmd theme-dark %}
```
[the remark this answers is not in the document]
str4d: I think this is just left over, but I will check.

age-plugin.md
str4d: The latter is intended here; if the requested state machine is unknown then there is no guarantee that the IPC protocol it uses is compatible with any plugin IPC that the client speaks. Note that the stanza-based IPC protocol is specified below this point, and is documented as for "v1 state machines".

> rage at least reacts to that with a rather unhelpful "Error: incomplete response", though I realize this could be an implementation infelicity.

Indeed, this is something I should fix in `rage` 😅

emilazy: That makes sense, thanks! Part of my uncertainty was that `age_plugin::run_state_machine` doesn't offer any way to handle only `identity-v1` (you can't `panic!` in the other branch because you still need a concrete `R: RecipientPluginV1` to type it with) other than implementing the whole interface and returning protocol errors inside; I suppose there is a distinction to be made between *unknown* state machines and *known but inapplicable* ones?

str4d: That's indeed the case. It's usually the case that you'll want to implement both (for entirely custom recipients), but indeed for a plugin that handles native age identities you wouldn't need the recipient state machine. Maybe there are cases where you might only want to implement the recipient state machine, but I can't think of any at present.

In any case, this should be fixed in the `age-plugin` crate (probably with the addition of a `run_identity_state_machine` function).

age-plugin.md
```diff
@@ -0,0 +1,541 @@
...
+## Mapping recipients and identities to plugin binaries
+
+age plugins are identified by an arbitrary case-insensitive string `NAME`. This
+string is used in three places:
+
+- Plugin-compatible recipients are encoded using Bech32 with the HRP `age1name`
+  (lowercase).
+- Plugin-compatible identities are encoded using Bech32 with the HRP
+  `AGE-PLUGIN-NAME-` (uppercase).
+- Plugin binaries (to be started by age clients) are named `age-plugin-name`.
```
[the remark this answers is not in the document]
str4d: Custom stanzas are not required to use the same `name` as the plugin, so that consideration is separate. For example, my `age-plugin-yubikey` plugin defined the stanza `piv-p256` (named without FQDN because it predates the age spec recommendation).

As for having both 1 and 2, that's a usability decision: hyphenated binary names are easier to read, and having `plugin` in the name helps with domain separation in the binary name space. But they aren't selectable with a double-click, which is something we wanted for recipient strings.

age-plugin.md
```diff
@@ -0,0 +1,541 @@
...
+for a particular request. Instead, the client returns `(fail)` to indicate this;
+the plugin is expected to try another recipient stanza or identity.
+
+When the plugin is able to determine whether a given file key can be unwrapped
+or not, it takes one of three actions:
+- If a stanza cannot be unwrapped that detectably should be unwrappable (e.g.
+  the stanza specifically identifies the recipient), the plugin sends
+  `error stanza`.
+  - TODO: Should this be a hard error (preventing that file from being unwrapped)?
+    Probably yes (we already assume we don't get 32-bit collisions for SSH tags).
```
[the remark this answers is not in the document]
str4d: Opened #32 for subsequent work.

age-plugin.md
str4d: Opened #32 for subsequent work.

age-plugin.md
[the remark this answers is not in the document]
str4d: Opened #33 for confirming whether or not shell aliases are broadly usable for this.

age-plugin.md
str4d: Opened https://github.com/str4d/rage/issues/387.

age-plugin.md
```diff
@@ -0,0 +1,541 @@
...
+
+## Mapping recipients and identities to plugin binaries
+
+age plugins are identified by an arbitrary case-insensitive string `NAME`. This
+string is used in three places:
+
+- Plugin-compatible recipients are encoded using Bech32 with the HRP `age1name`
+  (lowercase).
+- Plugin-compatible identities are encoded using Bech32 with the HRP
+  `AGE-PLUGIN-NAME-` (uppercase).
```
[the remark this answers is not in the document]
str4d: Opened #34 for deciding this later.

str4d:
PR comments have either been directly addressed, or have open issues for resolution before the final spec version. Merging this PR so we have a draft version we can reference as 0.1.

PR #15, opened 2022-06-20
age: Fix byte value typo.

msparks:
Noticed this while reading through the age specification and the draft plugin spec. I didn't see any precedent in this doc for uppercase vs. lowercase hex; went with lowercase, but happy to change if you prefer.

In any case, thank you for all your work on age and open source!

### DIFF
--- a/age.md
+++ b/age.md
@@ -67,7 +67,7 @@ or at the first line shorter than 64 columns (for stanzas).
 #### Version line
 
 The version line always starts with "age-encryption.org/", is followed by an
-arbitrary version string, and ends with a line feed (`0x0a`).
+arbitrary version string, and ends with a line feed (`0x0A`).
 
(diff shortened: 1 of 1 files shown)


FiloSottile:
Oh, good catch, thank you. I think I'd prefer uppercase, please.

age.md
```diff
@@ -67,7 +67,7 @@ or at the first line shorter than 64 columns (for stanzas).
 #### Version line
 
 The version line always starts with "age-encryption.org/", is followed by an
-arbitrary version string, and ends with a line feed (`0x20`).
+arbitrary version string, and ends with a line feed (`0x0a`).
```
FiloSottile, line 70:
```suggestion
arbitrary version string, and ends with a line feed (`0x0A`).
```

PR #16, opened 2022-07-02
age: require scrypt work factor to be canonical

FiloSottile:


str4d (approved):
ACK, this matches the testkit impl, and removes another non-canonicity possibility.

PR #18, opened 2022-08-29
WIP: Batched Merkle Audit Paths

rozbb:
[Rendered](https://github.com/rozbb/C2SP/blob/bpath/bpath.md)

This small standard (and Python 3 ref impl and test vectors) defines a batched version of [Certificate Transparency-compatible Merkle membership proofs](https://datatracker.ietf.org/doc/html/rfc6962#section-2.1.1). The TL;DR is that it's a more space-efficient way of proving membership of multiple elements in a Merkle tree. As an example: the size of a membership proof of the first 100 elements in a 1000-element tree is reduced by over 99% by using a batched proof, rather than using 100 individual membership proofs.

This is compatible with the original CT proofs in that it produces identical proofs whenever `n = 1`.

I've also got a Rust impl that's passing known-answer tests [here](https://github.com/rozbb/ct-merkle/blob/batch-inclusion/src/batch_inclusion.rs).

Unresolved items:

* Should figure out whether I want to define a verification algorithm. The original spec doesn't define one, probably because you'd have to introduce a lot more math to describe it in sufficient details
* **PLEASE GIVE INPUT:** Should decide whether or not to make a normative statement on valid vs invalid proofs. It is very likely that there are CT verification algorithms out there that will verify on proofs that are too long (i.e., the verifier will have computed the root, and not even looked at the remaining bytes in the proof). Is an overly long proof still a valid proof? It breaks canonicity, and we know that bad things happen when you break canonicity.
    * Another example of ambiguity is whether or not empty proofs are allowed in the `PROOF` function (proving Merkle tree consistency). Technically, the CT spec doesn't define `PROOF` for trivial inputs (subtree == main tree), but [some implementations](https://github.com/transparency-dev/merkle/blob/e739733c8ca2b89cfc49f35b161e81e16b5957d7/testonly/reference_test.go#L129) will happily produce and verify these kinds of proofs. I think it's wrong, but I have to decide how much I should care.
* Need to figure out the correct place to put the multiple JSON test vectors and Python 3 reference impl. @FiloSottile do you have opinions here?

Open to any and all comments!

### DIFF
--- a/bpath.md
+++ b/bpath.md
@@ -1,0 +1,79 @@
+# Batched Merkle Inclusion Proofs
+
+This document defines the `BPATH` algorithm for proving inclusion of multiple leaves in a Merkle tree. This standard extends the `PATH` function defined in the Certificate Transparency standard, [RFC 6962][]. `BPATH` is equivalent to `PATH` whenever just one leaf's membership is being proved.
+
+## Conventions used in this document
+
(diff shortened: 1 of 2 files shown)


rozbb:
Bumping this. Anyone have opinions on the above open questions? It'd be nice to have this standardized

PR #20, opened 2022-10-17
jq255 specification

pornin:
This document specifies the jq255e and jq255s prime-order groups. These groups are built on double-odd elliptic curves interpreted as Jacobi quartics, as described in: https://eprint.iacr.org/2022/1052

Hash-to-group, key exchange (ECDH) and signatures (Schnorr signatures) are also specified.

There are already four implementations of these groups, in [Python](https://github.com/doubleodd/py-jq255), [Go](https://github.com/doubleodd/go-jq255), [Rust](https://github.com/pornin/crrl) and [C](https://github.com/doubleodd/c-jq255). The Python implementation contains/generates ample test vectors, which I suppose should ultimately be pushed into C2TV.

(Maybe the spec itself should contain a couple test vectors?)


dconnolly:
Thank you for this PR! I'll be able to give a review later today.

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+    mapping of arbitrary data into group elements with purely
+    constant-time implementations.
+
+The underlying curves are specific double-odd curves (curves whose order
+is twice an odd integer) interpreted with a coordinate system such that
+the curve equation is a Jacobi quartic, as described in the
+[double-odd Jacobi quartic whitepaper][]. More information on double-odd
+curves can be found on the [double-odd elliptic curves Web site][].
+
+## Notations
```
dconnolly, line 55:
```suggestion
## Conventions and notation used in this document
```

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+in the same sense as, for instance, Curve25519 is argued to offer
+"128-bit security" with a subgroup order of "only" about `2^252`
+elements). The use of a half-width challenge `c` was already suggested
+by Schnorr in his seminal paper; the security properties and requirements
+on the hash function (here, BLAKE2s) were thoroughly analyzed by [Neven,
+Smart and Warinschi][].
+
+It has been [noted][] that while the construction with a half-width
+challenge is safe *as a signature scheme*, some advanced protocols try
+to also use signatures as commiments on a unique message: these
```
dconnolly, line 1234:
```suggestion
to also use signatures as commitments on a unique message: these
```

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+On top of these groups, the following primitives are specified by this
+document:
+
+  - A digital signature scheme. Signatures have size 48 bytes, which is
+    shorter than the more usual 64-byte signatures associated with
+    elliptic curves of similar security (e.g. EdDSA signatures on
+    Curve25519, or ECDSA over P-256 or secp256k1 curves), but still
+    provides the security level expected from the curve. Signature
+    verification is also faster than with the common 64-byte signatures.
+
```
dconnolly, line 42: The half-size challenge signature scheme defined below is interesting - would it be useful to break it out into its own group and hash agnostic scheme, as long as the group and hash properties are constrained?

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+  - A key exchange protocol (Diffie-Hellman).
```
dconnolly, line 43: 🔑 🔀 🗝️

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+
+  - Map-to-curve and hash-to-curve procedures, that allow one-way
+    mapping of arbitrary data into group elements with purely
+    constant-time implementations.
```
dconnolly, line 47: 👍

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+
+The underlying curves are specific double-odd curves (curves whose order
+is twice an odd integer) interpreted with a coordinate system such that
+the curve equation is a Jacobi quartic, as described in the
+[double-odd Jacobi quartic whitepaper][]. More information on double-odd
+curves can be found on the [double-odd elliptic curves Web site][].
```
dconnolly, line 53:
```suggestion
[double-odd Jacobi quartic whitepaper]. More information on double-odd
curves can be found on the [double-odd elliptic curves Web site].
```

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+    with `r` being a prime integer. The group itself (jq255e or jq255s)
+    has order `r`. For both groups, `r` is close to `2^254`.
+
+  - The elliptic curve is the set of points `(e, u)`, with both
+    coordinates being elements of `GF(q)` that fulfill the curve
+    equation `e^2 = (a^2 - 4*b)*u^4 - 2*a*u^2 + 1`, for two constants
+    `a` and `b` that are part of the curve definition. The curve neutral
+    point (the "point-at-infinity" which is not actually "at infinity"
+    in this coordinate system) is `I = (1,0)`. The curve also contains a
+    single point of order 2 denoted `N = (-1,0)`.
```
dconnolly, line 87:
```suggestion
  - The elliptic curve is the set of points `(e, u)`, with both
    coordinates being elements of `GF(q)` that fulfill the curve
    equation `e^2 = (a^2 - 4*b)*u^4 - 2*a*u^2 + 1`, for two constants
    `a` and `b` that are part of the curve definition. The curve neutral
    point (the "point-at-infinity" which is not actually "at infinity"
    in this coordinate system) is `I = (1,0)`. The curve also contains a
    single point of order 2 denoted `N = (-1,0)`.
    
  - The underlying elliptic curve contains exactly `n = 2*r` points,
    with `r` being a prime integer. The group itself (jq255e or jq255s)
    has order `r`. For both groups, `r` is close to `2^254`.
```

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+  - The underlying elliptic curve contains exactly `n = 2*r` points,
+    with `r` being a prime integer. The group itself (jq255e or jq255s)
+    has order `r`. For both groups, `r` is close to `2^254`.
+
+  - The elliptic curve is the set of points `(e, u)`, with both
+    coordinates being elements of `GF(q)` that fulfill the curve
+    equation `e^2 = (a^2 - 4*b)*u^4 - 2*a*u^2 + 1`, for two constants
+    `a` and `b` that are part of the curve definition. The curve neutral
+    point (the "point-at-infinity" which is not actually "at infinity"
+    in this coordinate system) is `I = (1,0)`. The curve also contains a
```
dconnolly, line 86: 💖

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+    `e^2 = b'*u^4 + a'*u^2 + 1`
+
+  - A *group element* is defined as a pair `{ P, P+N }` for any curve
+    point `P`. The points `P` and `P+N` are the two possible
+    *representants* of the group element. The group law is denoted
+    additively: for any two group elements, their sum is computed by
+    simply adding together their representant points over the curve;
+    since `N` has order 2, it does not matter which representant is used
+    for each group element. The encoding process ensures that the same
+    output is obtained for both representants of the same group element.
```
dconnolly, line 100: I would tweak this wording to indicate that while the author's/encoding's choice of which of the two points to represent the group element doesn't matter, there _is_ one correct canonical representation of the group element, and doing funny stuff with the other point is a bad idea and abstraction violation that may lead to tears

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+
+  - Integers modulo `r` are called *scalars*. Since `r` is prime, the
+    scalars constitute a finite field. A group element can be multiplied
+    by a scalar by using any integer representant of the scalar.
+
```
dconnolly, line 105: Need to specify that 'multiply by scalar' is 'applying the group law to a group element value-of-scalar number of times'

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+jq255e:
+
+  q = 2^255 - 18651
+  a = 0
+  b = -2
+  a' = -2*a      = 0
+  b' = a^2 - 4*b = 8
+  equation: e^2 = 8*u^4 + 1
+  curve order: n = 2*r
+     r = 2^254 - 131528281291764213006042413802501683931
```
dconnolly, line 118:
```suggestion
  group order: r = 2^254 - 131528281291764213006042413802501683931
  curve order: n = 2*r
```

We may want to elide the full curve order unless strictly needed, as the _group_ constructed from the curve, and that group's order, is of primary interest.

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+jq255e:
+
+  q = 2^255 - 3957
+  a = -1
+  b = 1/2 = 2^254 - 1978
+  a' = -2*a      = 2
+  b' = a^2 - 4*b = -1
+  equation: e^2 = -u^4 + 2*u^2 + 1
+  curve order: n = 2*r
+     r = 2^254 + 56904135270672826811114353017034461895
```
dconnolly, line 136:
```suggestion
  group order: r = 2^254 + 56904135270672826811114353017034461895
  curve order: n = 2*r
```

Similar as above, we may want to elide the full curve order unless strictly needed, as the group constructed from the curve, and that group's order, is of primary interest.

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+  conventional generator: G = (eg, ug)
+     eg = 6929650852805837546485348833751579670837850621479164143703164723313568683024
+     ug = 3
+```
+
+## Group Element Operations
+
+In this section, practical procedures for computing point operations are
+presented using pseudocode in a syntax compatible with the Python language
+and the [SageMath][] mathematics system. This syntax closely follows the
```
dconnolly, line 146:
```suggestion
and the [SageMath] mathematics system. This syntax closely follows the
```

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+
+In this section, practical procedures for computing point operations are
+presented using pseudocode in a syntax compatible with the Python language
+and the [SageMath] mathematics system. This syntax closely follows the
+abstract mathematical notation otherwise used in this document, with the
+following exceptions:
+
+  - Exponentiation uses the `**` operator (instead of `^`).
+  - Points with extended coordinates `(E:Z:U:T)` become
+    tuples `(E,Z,U,T)`.
```
dconnolly, line 152: 👍

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+
+The pseudocode assumes that the following functions are available:
```
dconnolly, line 154: "any endianness concerns?"

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+    `j` modulo `q`.
+
+  - `int(x)` converts the field element `x` into an integer by using its
+    unique representant in the `0` to `q-1` range.
+
+  - `half(x)` divides the field element `x` by two. It is usually
+    implemented by a right-shift by 1 bit, then the conditional addition
+    of the constant `(q+1)/2` if the dropped bit was non-zero. It is
+    considered to be a "fast" operation (i.e. at least as fast as an
+    addition in the field).
```
dconnolly, line 166:
```suggestion
  - `half(x)` divides the field element `x` by two. It is usually
    implemented by a right-shift by 1 bit, then the conditional addition
    of the constant `(q+1)/2` if the dropped bit was non-zero. Such an 
    implementation is considered to be a "fast" operation (i.e. at least
    as fast as an addition in the field).
```

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+
+  - `invert(x)` returns the multiplicative inverse of `x` in the field.
+    Zero has no multiplicative inverse, but it is sometimes convenient
+    for constant-time implementations to return `0` as a (formal) inverse
+    of `0`; this is obtained naturally if inversion is implemented using
+    Fermat's Little Theorem (i.e. raising `x` to the power `q-2`). This
+    feature is not needed for the formulas described in this document.
+
+  - `sgn(x)` returns the "sign" of `x`, as defined previously: the
+    sign is the value of the least significant bit of `int(x)`.
```
dconnolly, line 176: Aha yes, endianness

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+    a square in the finite field), or `False` (if `x` is not a square
+    in the finite field). **Important:** when `x` is a square, the
+    returned square root is non-negative.
+
+  - `is_square(x)` return `True` if `x` is a square in the finite field,
+    or `False` otherwise.
+
+  - `scalar_to_int(s)` and `int_to_scalar(j)` compute the same
+    operations as `int()` and `GFq()`, respectively, but for scalars,
+    which are integers modulo the prime `r`.
```
dconnolly, line 188: 💖

jq255.md
```diff
@@ -0,0 +1,1405 @@
...
+
+  - `secure_random(m)`, for an integer `m >= 1`, returns a sequence
+    of `m` random bytes obtained from a cryptographically secure source,
+    with a uniform probability distribution. This is used only for
+    private key generation in the specification of high-level protocols.
```
dconnolly, line 193: Does this need more detail, hm

jq255.md
pornin: Technically you can define such a scheme over any prime-order group (that's basically what Schnorr did in the first place!) and you could do that over, for instance, Ristretto. But here I am trying to have something self-contained that does not spill over several spec documents, since such dispersion seems to lead only to the spec being ignored (conversely, I always found that the great advantage of RFC 8032 is that it did everything in a single document). Also, there are some shortcuts that I can take here that are interesting for implementations, especially on small microcontrollers, namely that I can take a 32-byte random blob and reduce it modulo r, and it's unbiased enough; I cannot do that generically, but for jq255e and jq255s it works because r is very close to a power of 2. A generic definition for all groups would have to "play safe" and use a 48-byte output and that would mean BLAKE2b or a BLAKE2s-based XOF with several invocations, and more generic modular reduction, and that would be inconvenient for implementers (and increase code footprint).

jq255.md
pornin: You mean adding a link to RFC 4086? I am not entirely sure this is a good idea. I am tempted to say that if an implementer generates private keys, sees the words "cryptographically secure source", and goes for Mersenne Twister, then that calls for some educational session that will exceed the scope of the spec...

jq255.md
pornin: Wording has been tweaked.

jq255.md
pornin: I have expanded a bit on the words, hopefully it is clearer now.

jq255.md
dconnolly: Not necessarily a link to RFC 4086, perhaps an example ("such as sourcing from /dev/urandom") may suffice

pornin: Seems fine. Done.

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+  - `scalar_to_int(s)` and `int_to_scalar(j)` compute the same
+    operations as `int()` and `GFq()`, respectively, but for scalars,
+    which are integers modulo the prime `r`.
+
+  - `secure_random(m)`, for an integer `m >= 1`, returns a sequence of
+    `m` random bytes obtained from a cryptographically secure source,
+    with a uniform probability distribution (e.g. `/dev/urandom` or the
+    `getrandom()` system call on many Unix-like operating syustems).
+    This is used only for private key generation in the specification of
+    high-level protocols.
```
dconnolly, line 200: 👍

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+ 1. Compute: `c = (2*x)^((q-5)/8)`
+ 2. Compute: `d = 2*x*c^2`
+ 3. Compute the square root candidate: `z = x*c*(d - 1)`
+
+The obtained candidate MUST be checked (by comparing `z^2` with `x`).
+Moreover, if `z` is indeed a square root of `x`, then it MUST be
+normalized to a non-negative value (i.e. if `sgn(z) = 1`, then `z` must
+be replaced with `-z`). The use of this fixed convention for square
+roots is needed for proper interoperability of the map-to-group and
+hash-to-group operations.
```
dconnolly, line 224: Should we bring in RFC 2119 for 'MUST' etc?

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+the field is such that `q = 3 mod 4`, which allows a simple square root
+implementation by computing the square root candidate `z = x^((q+1)/4)`.
+For jq255e, the field is such that `q = 5 mod 8`, which can be handled
+with Atkin's algorithm:
+
+ 1. Compute: `c = (2*x)^((q-5)/8)`
+ 2. Compute: `d = 2*x*c^2`
+ 3. Compute the square root candidate: `z = x*c*(d - 1)`
+
+The obtained candidate MUST be checked (by comparing `z^2` with `x`).
```
dconnolly, line 219:
```suggestion
The obtained candidate MUST be checked (by comparing `z^2` with `x` and finding them equivalent).
```

jq255.md
dconnolly, line 224: 👍

jq255.md
dconnolly: @str4d @FiloSottile thoughts?

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+Moreover, if `z` is indeed a square root of `x`, then it MUST be
+normalized to a non-negative value (i.e. if `sgn(z) = 1`, then `z` must
+be replaced with `-z`). The use of this fixed convention for square
+roots is needed for proper interoperability of the map-to-group and
+hash-to-group operations.
+
+`is_square(x)` can be implemented as a simple wrapper around `sqrt(x)`,
+though faster implementations are possible with the [Legendre/Jacobi
+symbol][]. A fast, constant-time implementation of that operation can be
```
dconnolly, line 228:
```suggestion
`is_square(x)` can be implemented as a simple wrapper around `sqrt(x)`,
though faster implementations are possible with the [Legendre/Jacobi
symbol]. A fast, constant-time implementation of that operation can be
```

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+obtained as a straightforward derivative of the [optimized binary
+GCD][]; on some small microcontrollers, this `is_square()`
```
dconnolly, line 230:
```suggestion
obtained as a straightforward derivative of the [optimized binary
GCD]; on some small microcontrollers, this `is_square()`
```

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+implementation can be about six times faster than the full square root
+computation.
```
dconnolly, line 232: 👍

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+
+### Affine Formulas
+
+We recall here the affine formulas applicable to points on double-odd
+Jacobi quartic curves. Implementations are expected to use the extended
+representation described later on in this document; the affine formulas
+are potentially useful if some alternate point representations are
+explored.
```
dconnolly, line 240: Thank you for this; it will be useful to have the intended implementation in extended coordinate formulas and then a slower implementation in affine formulas which can then be checked against each other for correctness or other testing

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+
+For any point `P = (e, u)` on the curve, the point `P+N` (i.e. the other
+representant of the same group element) is `P+N = (-e, -u)`. The
+*opposite* of the point `P` is `-P = (e, -u)`. The neutral element of
+the group is `{ I, N }` (i.e. `I = (1, 0)` and `N = (-1, 0)` are the two
+representants of the group neutral).
```
dconnolly, line 246: Is the *opposite* the inverse?

dconnolly: If `N = (-1, 0)` and `P=(e, u)`, how does `P+N = (-e, -u)`?

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+equivalent to the curve with Weierstraß equation `y^2 = x*(x^2 + a*x + b)`
+with the following mappings (defined for all points excepts `I` and `N`):
+
+```
+x = (e + 1 - a*u^2)/(2*u^2)
+y = x/u
+
+e = (x^2 - b)/(x^2 + a*x + b)
+u = x/y
+```
```
dconnolly, line 274: 💖

jq255.md
pornin: It depends on how much you want C2SP to mimic or fight the IETF.

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+
+On the Weierstraß curve, `N` has coordinates `(0, 0)`, and `I` has no
+defined coordinates (it is then truly "at infinity"). The Weierstraß
```
dconnolly, line 277: ♾️

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+X = x + a/3
+Y = y
+```
+
+Converting to Weierstraß coordinates and back is not required for
+implementing the jq255e and jq255s groups; such coordinates have
+furthermore special cases for the points `I` and `N`. These conversions
+may be useful in some specific contexts, e.g. to leverage preexisting
+generic implementations of (short) Weierstraß curves, if only for
+testing purposes.
```
dconnolly, line 292: 👍

jq255.md
pornin: I think "opposite" is the correct term if we talk about "adding" points; "inverse" relates to multiplication, but here I am denoting the group law additively.

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+In order to reduce the number of field element divisions (which are
+usually quite expensive in software implementations, relative to
+additions and multiplications), the following internal representation is
+recommended: a point `P = (e, u)` is represented by the quadruplet
+`(E:Z:U:T)` of field elements, with the following rules:
+
+  - `Z != 0` (in all cases)
+  - `e = E/Z`
+  - `u = U/Z`
+  - `u^2 = T/Z` (i.e. `U^2 = T*Z`)
```
dconnolly, line 305: 👍

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+
+These coordinates are called *EZUT coordinates*.
+
+The points `I` and `N`, which represent the neutral element in the group,
+have coordinates `(Z:Z:0:0)` and `(-Z:Z:0:0)`, respectively, for any
+field element `Z != 0`. It does not matter, for correct computations, which
+of `I` or `N` is used to represent the neutral; similarly, any non-zero `Z`
+can be used.
```
dconnolly, line 313: Hm

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+
+*Extended affine coordinates* are a special case of extended coordinates
+where `Z = 1` (in which case, `E = e`, `U = u` and `T = u^2`). Such
+coordinates are appropriate for use in fixed precomputed tables of
+points, e.g. small multiples of the conventional generator; setting `Z = 1`
+for such points reduces the table size in memory, and allows for some
+shortcuts in operations on these points.
```
dconnolly, line 320: 👍

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+
+The points `I` and `N` can be represented on XWJ coordinates with the
+triplets `(W^2:W:0)` and `(0:W:0)`, respectively, for any `W != 0`. For
+all curve points other than `I` and `N`, the coordinate `J` is non-zero.
+In general, Jacobian coordinates do not lead to efficient and complete
+point addition formulas, but they offer fast and complete formulas for
+point doublings. Thus, a strategy for computing a sequence of several
+successive point doublings is to convert the source point from EZUT to
+XWJ coordinates, compute the doublings in XWJ, then convert the result
+back into EZUT coordinates.
```
dconnolly, line 340:
```suggestion
point doublings. Thus, a strategy for computing a sequence of several
successive point doublings is to convert the source point from EZUT to
XWJ coordinates, compute the doublings in XWJ, then convert the result
back into EZUT coordinates. Such a strategy is used later in doubling formulas.
```

jq255.md
pornin: If you start with `P = (e,u)`, you have `-P = (e, -u)`, `P+N = (-e, -u)` and `-P+N = (-e, u)`. The point `N` itself is `(-1, 0)`, hence `N+N = I = (1, 0)` (since `N` has order 2, adding it to itself yields the point-that-was-at-infinity).

In the Weierstraß curve, `u` is the inverse of the slope of the line that goes from `N` to `P`, and that line (by construction) also intersects the curve at point `-(P+N) = -P+N`, so it is expected that `P` and `-P+N` share the same `u`, while point negation is symmetry relatively to the horizontal, hence `-P` must have the opposite slope (hence `-u`). Geometric interpretation of `e` is a bit harder in the Weierstraß domain, because `e` is "unscaled" (curve isomorphisms are obtained by multiplying `x` by `d^2`, `y` by `d^3`, `a` by `d^2` and `b` by `d^4` for any non-zero `d`; but `e` is unchanged through such isomorphisms).

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+
+### Encoding and Decoding
```
dconnolly, line 342: We have this encoding/decoding section and a subsection later under High Level Protocols that is also about the same, perhaps we should rename this section or merge it into the later section? These subroutines are only invoked there

pornin: Good question. I'd say that from an engineering point of view, there should be an implementation of operations on finite fields, that includes encoding and decoding operations of finite field elements, and then a separate implementation of operations in the group, where group elements are encoded and decoded (and that relies on the encoding and decoding of field elements). The split of the encoding and decoding into separate subsections relates to that code structure (and it's how I implemented it in various places, e.g. in crrl). But within a _specification_, we can potentially do differently.

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+the group is `{ I, N }` (i.e. `I = (1, 0)` and `N = (-1, 0)` are the two
+representants of the group neutral).
+
+Let `P1 = (e1, u1)` and `P2 = (e2, u2)` be two points on the curve. Their
+sum `P3 = P1 + P2 = (e3, u3)` can be computed as follows:
+
+```
+     (1 + b'*u1^2*u2^2)*(e1*e2 + a'*u1*u2) + 2*b'*u1*u2*(u1^2 + u2^2)
+e3 = ----------------------------------------------------------------
+                          (1 - b'*u1^2*u2^2)^2
```
FiloSottile, line 254: I think conventions and notations doesn't specify that `/` means multiplication by the inverse.

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+  conventional generator: G = (eg, ug)
+     eg = 6929650852805837546485348833751579670837850621479164143703164723313568683024
+     ug = 3
+```
+
+## Group Element Operations
+
+In this section, practical procedures for computing point operations are
+presented using pseudocode in a syntax compatible with the Python language
+and the [SageMath] mathematics system. This syntax closely follows the
```
FiloSottile, line 151: It's unclear which procedures this refers to, as there's plenty of them using `^`.

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+generic implementations of (short) Weierstraß curves, if only for
+testing purposes.
+
+### Extended Coordinates
+
+In order to reduce the number of field element divisions (which are
+usually quite expensive in software implementations, relative to
+additions and multiplications), the following internal representation is
+recommended: a point `P = (e, u)` is represented by the quadruplet
+`(E:Z:U:T)` of field elements, with the following rules:
```
FiloSottile, line 300: This is mildly surprising, as I am used to X:Y:Z:T coordinates. Is the inversion of U and Z intentional?

jq255.md
FiloSottile: Hmm, it's an interesting question. In my mind, the fundamental implementation of a prime order group interface is everything up to encoding and decoding of elements to/from _bytes_, as well as mapping bytes to an element, so all of this section and half the section below.

Signatures, key exchange and hash-to-group operations are the higher-level stuff.

The Ristretto document has all of the former and none of the latter, for example.

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+
+This document specifies the **jq255e** and **jq255s** groups. These are
+prime order groups based on elliptic curves, with characteristics
+suitable for building cryptographic protocols:
+
+  - The order of each group is a prime integer close to `2^254`, hence
+    formally offering "127-bit security" against discrete logarithm
+    attacks.
+
+  - Each group element (including the neutral element) can be encoded as
```
FiloSottile, line 16: I think we call the neutral element the identity in Ristretto and I've been trying to use that across groups. Opinions?

jq255.md
FiloSottile: On the one hand, I'd like C2SP to define reusable components across clean abstractions like prime-order groups. On the other hand, I really dislike the "least common multiple" effect of having to bring in the complexity required to handle all possible concrete implementations. It's a tension. I feel like a way to optimize that is not aiming for runtime composition (aka cipher suites) but "design-time" composition.

This specific shortcut though feels like it can be addressed as part of the abstraction: you define a map from a fixed number of uniform bytes to a uniform scalar, and for these groups the fixed number of bytes is 32.

I tend to agree having everything in a document is advantageous, but maybe that can be achieved by reference, like "signatures over these groups are constructed by applying style X of the prime-order group signatures document with these parameters" followed by test vectors that lock them in.

FiloSottile (commented):
Thank you, this is great!

(Very partial review, I got to _Encoding and Decoding_. I'll continue later.)

jq255.md
FiloSottile: Heh, neither is the goal, really. There are some things of the IETF culture that are good. RFC 2119, I'd argue, is at least good by being universally known by the folks who implement specs. There's an argument that maybe there shouldn't be that many SHOULD in a robust spec, but in general I am not against using MUST/SHOULD/etc and referencing RFC 2119 if that's how we are using them.

jq255.md
pornin: Actually it's all of them (or, if you prefer, all the ones which are in colour).

jq255.md
pornin: Yes, with the inversion it makes it sound like a French mild expletive (equivalent to something like "oh darn", nothing dire).

jq255.md
pornin: "Identity" works. In French we usually say _élément neutre_, hence my use of "neutral" by default, but this is certainly arguable and no big deal.

jq255.md
pornin: I have added a reference to RFC 2119 in the "conventions" section.

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+e3 = ----------------------------------------------------------------
+                          (1 - b'*u1^2*u2^2)^2
+
+      e1*u2 + e2*u1
+u3 = ----------------
+     1 - b'*u1^2*u2^2
+```
+
+These formulas are *complete*: they work on all pairs of curve points,
+including the points `I` and `N` (that represent the neutral element).
```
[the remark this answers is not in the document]
pornin: I added a sentence to recall that point.

jq255.md
pornin: I renamed the neutral into "identity".

jq255.md
pornin: I have added a sentence to that effect.

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+point addition formulas, but they offer fast and complete formulas for
+point doublings. Thus, a strategy for computing a sequence of several
+successive point doublings is to convert the source point from EZUT to
+XWJ coordinates, compute the doublings in XWJ, then convert the result
+back into EZUT coordinates.
+
+### Encoding and Decoding
+
+A group element, represented by the point `P = (E:Z:U:T)`, can be
+encoded unambiguously into a field element `v` as follows:
```
dconnolly, line 357: This might go in the later section, but if someone chose the other point of the two points per group element as the argument to group_to_field(), will that be a different field element?

FiloSottile: I think not, that's what `if sgn(iZ*E) == 1:` is doing.

It's also spelled out below:

> The group element is really encoded as the `u` coordinate of the
representant point whose `e` coordinate is non-negative.

This is the part that makes the prime order group, both here and in Ristretto and in Decaf: encoding will encode the two representatives to the same value.

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+        v = -v
+    return v
+```
+
+Here, `iZ*U` and `iZ*E` are the `u` and `e` coordinates, respectively.
+The group element is really encoded as the `u` coordinate of the
+representant point whose `e` coordinate is non-negative.
+
+The decoding process rebuilds either the original `P`, or the other
+representant `P+N` of the same group element:
```
dconnolly, line 374: Oh

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+    return (E, Z, U, T)
+```
+
+The `field_to_group(v)` function returns a representant of the (unique)
+group element that encodes to the value `v`, or `False` if there is no
+such element. Take care that `sqrt()` MUST return the non-negative root.
+
+**Note:** The decoding process naturally returns a point in (extended)
+affine coordinates (`Z = 1`).
+
```
dconnolly, line 395: 👍

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+    U3 = half((hd + eu)**2 - Z3 - T3)   # Alternatively: U3 = hd*eu
+    return (E3, Z3, U3, T3)
+```
+
+If the point `P2` is in extended affine coordinates, i.e. such that `Z2`
+is statically known to be equal to 1, then the multiplication `Z1*Z2`
+can be avoided, and the computation of `tz` simplifies to `Z1*T2 + T1`.
+This specific case is known as a *mixed addition* and is typically
+encountered when multiplying the conventional generator by a scalar,
+using precomputed tables of small multiples of the generator.
```
dconnolly, line 427: 👍

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+the following:
+
+  - Perform the first doubling with formulas that receive `P` in
+    EZUT coordinates, and output `2*P` in XWJ coordinates.
+  - Perform the other `k-1` doublings in XWJ coordinates.
+  - Convert the final result from XWJ to EZUT.
+
+Depending on the used curve, it can be beneficial to combine some of
+these doublings with a switch to the other representant (i.e. computing
+`2*P+N` instead of `2*P`); this does not change the output as a group
```
dconnolly, line 458:
```suggestion
`2*(P+N)` instead of `2*P`); this does not change the output as a group
```

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+        s3 = s2**2
+        X = s3**2
+        J = J*((W + s2)**2 - s1 - s3)   # Alternatively: J = 2*J*W*s2
+        W = s3 - 2*(s1**2)
+
+    # Conversion XWJ -> EZUT
+    Z = W**2
+    T = J**2
+    U = half((W + J)**2 - Z - T)        # Alternatively: U = J*W
+    E = 2*X - Z
```
dconnolly, line 495: Should this be its own subroutine?

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+        s3 = (W + J)**2 - 2*s1
+        J = 2*s1*(2*X - s3)
+        X = 8*(s2**2)
+        W = 2*s2 - s3**2
+
+    # Conversion XWJ -> EZUT
+    Z = W**2
+    T = J**2
+    U = half((W + J)**2 - Z - T)        # Alternatively: U = J*W
+    E = 2*X - Z - T
```
dconnolly, line 527: Same here

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+    (E1, Z1, U1, T1) = P1
+    (E2, Z2, U2, T2) = P2
+    return U1*E2 == U2*E1
+```
+
+This function handles all cases properly, including when one or both
+of the operands represent the group identity.
+
+A consequence is that a point `P = (E:Z:U:T)` represents the identity
+element in the group if and only if `U = 0`.
```
dconnolly, line 547: 💖

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+### Field to Point Map
+
+We define two field-to-point maps (for jq255e and jq255s, respectively)
+which turn a given field element into a group element (really, a curve
+point) such that the discrete logarithm of the output relatively to the
```
dconnolly, line 553:
```suggestion
point) such that the discrete logarithm of the output relative to the
```

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+This map uses the constant `sqrtm1 = sqrt(-1)` on the field. Since there are
+two such square roots modulo `q`, we conventionally use the non-negative
+root, i.e.:
+
+```
+sqrtm1 = 7656063742463026568679823572395325799027601838558345258426535816504372595438
+```
+
+The field element `f` is mapped to a point as follows:
```
dconnolly, line 572:
```suggestion
The field element `f` is mapped to a point in EZUT coordinates as follows:
```

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+in that case. Similarly, `is_square(z1)` and `is_square(z2)` would be
+systematically computed, and constant-time conditional move operations
+performed to get the appropriate operand to the (single) `sqrt()`
+operation, and also set the `x`, `xx`, `y` and `yy` variables
+appropriately.
+
+### On jq255s
+
+On jq255s, we use the Elligator2 map. The field element `f` is mapped to
+a point as follows:
```
dconnolly, line 609: Why Elligator2 for jq255s and not for jq255e?

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+        R = point_xdouble(R, 5)
+        R = point_add(R, Q)
+    return R
+```
+
+This process becomes constant-time if all the steps are constant-time.
+In particular, the lookup process from the window must then scan *all*
+elements from the window systematically, and select the correct one with
+constant-time conditional move operations that use bitwise Boolean
+operations.
```
dconnolly, line 734: 👍

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+This is the same value `sqrtm1` which was already used in the specification
+of the `map_to_jq255e()` function. In `(x, y)` coordinates on the
+corresponding Weierstraß equation, the `zeta()` function returns
+`(-x, sqrtm1*y)`.
+
+In EZUT coordinates, `zeta()` is computed as `zeta(E:Z:U:T) =
+(E:Z:sqrtm1*U:-T)`. It can be shown that applying `zeta()` to a group
+element yields exactly the same group element as multiplying by a given
+scalar `mu`, which happens to be a square root of -1 in the scalar field
+(integers modulo `r`). With the value of `sqrtm1` defined above, the
```
dconnolly, line 766: I now love endomorphisms

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+Any given scalar `s` can be split into two half-width signed integers
+`s0` and `s1`, such that `s = s0 + mu*s1`, and the absolute values of
+`s0` and `s1` are both at most equal to `sqrt(r)`. In particular, they
+both fit on 128 bits each, as signed integers, including the sign bit.
+The splitting process can be done efficiently with operations on plain
+integers; it involves some rounded divisions, but these can be
+simplified into only multiplications and shifts thanks to the specific
+value of the group order `r`. For details, refer to the [double-odd
+elliptic curves whitepaper][] (section 6.2). Once `s` has been split
```
dconnolly, line 785:
```suggestion
value of the group order `r`. For details, refer to the [double-odd
elliptic curves whitepaper] (section 6.2). Once `s` has been split
```

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+containing points `zeta(j*P)`, and only half the number of point
+doublings.
+
+## High-Level Protocols
+
+*High-level protocols* are signatures, key exchange and hash-to-group
+operations. We specify here how these operations are performed, and
+in particular all encoding rules and handling of invalid inputs.
+
+The primitives below use the [BLAKE2][] hash function, specifically
```
dconnolly, line 797:
```suggestion
The primitives below use the [BLAKE2] hash function, specifically
```

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+thus, the last byte of a 32-byte sequence has index 31.
+
+#### Field Elements
+
+A field element `x` is encoded as a fixed 32-byte sequence using the
+unsigned little-endian convention:
+
+```python
+def field_encode(x):
+    return int(x).to_bytes(32, byteorder='little')
```
dconnolly, line 830: 🙏

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+Since `q` is slightly below `2^255`, the last output byte (index 31)
+necessarily has value at most 127, which means that the most significant
+bit of that byte is necessarily equal to 0.
+
+When decoding, the following rules MUST be observed:
+
+  - The decode MUST reject as invalid any input sequence whose length
+    is not exactly 32 bytes.
+  - The integer value `z` must be recomputed by taking into account
+    all input bytes completely. In particular, the most significant
```
dconnolly, line 849:
```suggestion
  - The integer value `z` MUST be recomputed by taking into account
    all input bytes completely. In particular, the most significant
```

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+The following function implements these rules:
+
+```python
+def field_decode(bb):
+    if len(bb) != 32:
+        return False
+    z = int.from_bytes(bb, byteorder='little')
+    if z >= q:
+        return False
+    return GFq(z)
```
dconnolly, line 864: This is the first mention of this type/constructor

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+    implicit or explicit reduction modulo `q` is incorrect. If
+    `z >= q`, then the decoder MUST reject the input.
+
+The following function implements these rules:
+
+```python
+def field_decode(bb):
+    if len(bb) != 32:
+        return False
+    z = int.from_bytes(bb, byteorder='little')
```
dconnolly, line 861: 🙏

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+    if z >= q:
+        return False
+    return GFq(z)
+```
+
+Any application or outer protocol may reuse the most significant bit of
+the last byte to store some other data element, provided that the bit is
+restored to its proper value (0) before giving the byte sequence as
+argument to `field_decode()`.
```
dconnolly, line 870: 🤔

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+#### Scalars
+
+Scalars are integers modulo `r`. They follow the same rules as field
+elements, except that they work with the modulus `r` instead of `q`:
+
+```python
+def scalar_encode(x):
+    return scalar_to_int(x).to_bytes(32, byteorder='little')
```
dconnolly, line 879: 💖

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+def scalar_decode(bb):
+    if len(bb) != 32:
+        return False
+    z = int.from_bytes(bb, byteorder='little')
```
dconnolly, line 884: 💖

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+**Note:** This function generates a random 256-bit integer, then reduces
+it modulo `r` (through the `int_to_scalar()` function). Since the value
+of `r`, in both jq255e and jq255s, is very close to a power of two, then
+this process induces only negligible bias; in the general case, one
+would need to produce a source integer substantially larger than `r`
+(e.g. 384 bits), but this is not required for jq255e and jq255s. Using a
+source 256-bit integer means that the reduction modulo `r` can be
+implemented in a simple way, without requiring a generic big integer
+division routine.
```
dconnolly, line 961: 👍

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+### Signatures
+
+A **signature** is computed over a given input message, using a private
+key; the signature value has size exactly 48 bytes. The correctness of
+the signature value over a given message can be verified against the
+public key corresponding to the private key used for signing.
+
+Internally, a Schnorr signature process is used; the challenge value is
+computed with the [BLAKE2][] hash function. Specifically, the BLAKE2s
```
dconnolly, line 1002:
```suggestion
computed with the [BLAKE2] hash function. Specifically, the BLAKE2s
```

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+function is used, with a 32-byte output (as will be described below,
+that 32-byte output is then truncated to 16 bytes, but the hash output
+length 32 is still used in the BLAKE2 parameter block).
+
+In this section, we use the following notations:
+
+  - The signer's private key is `sk`.
```
dconnolly, line 1009: Signing key?

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+  - The signer's public key is the group element `Q = sk*G`.
```
dconnolly, line 1011: Verifying key?

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+      - A single byte of value `0x48`.
+
+      - The symbolic name of the hash function that was used to produce
+        `hv`, encoded in ASCII.
+
+      - A single byte of value `0x00` (which terminates the encoded hash
+        function name).
+
+      - The have value `hv`.
```
dconnolly, line 1033:
```suggestion
      - The hash value `hv`.
```

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+def prepare_message_prehashed(hv, hashname):
+    return b'\x48' + hashname.encode() + b'\x00' + hv
+```
+
+The **recommended usage** is to apply jq255e and jq255s signatures on
+input messages pre-hashed with BLAKE2s with a 32-byte (256-bit) output.
+The signature scheme shall be designated, formally, as follows:
+
+  - **jq255e** and **jq255s**: signatures are computed over the
```
dconnolly, line 1069: Are these not 'just' jq255e-blake2s and jq255s-blake2s ?

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+Thus, "jq255e", as a signature scheme name, is equivalent to
+"jq255e-blake2s".
+
+The following remarks apply:
+
+  - Regardless of the hash function used to pre-hash the message (if any),
+    the *internal* hash function used to compute the challenge is always
+    BLAKE2s. Within the scope of this specification, the use of BLAKE2s
+    internally is not configurable.
```
dconnolly, line 1090: oof that might get confusing

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+  - Using a raw input conceptually provides additional protection
+    against collision attacks on weak hash functions. On the other hand,
+    raw inputs are not compatible with some kinds of streamed processing,
+    forcing either the sender or the recipient of a signed message to
+    buffer the entirety of the signed data. This is why pre-hashing with
+    a properly collision-resistant hash function is recommended in
+    general.
```
dconnolly, line 1098: 👍

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+function to generate `k` in a safe and deterministic way; that process
+uses an extra optional parameter called `seed`. More on this function
+is detailed later on.
+
+```python
+def generate_nonce(sk, Q, M, seed):
+    sh = hashlib.blake2s()
+    sh.update(private_key_encode(sk))
+    sh.update(public_key_encode(Q))
+    sh.update(len(seed).to_bytes(8, byteorder='little'))
```
dconnolly, line 1131: interesting

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+message (`M`) as input. That function calls the `generate_nonce()`
+function to generate `k` in a safe and deterministic way; that process
+uses an extra optional parameter called `seed`. More on this function
+is detailed later on.
+
+```python
+def generate_nonce(sk, Q, M, seed):
+    sh = hashlib.blake2s()
+    sh.update(private_key_encode(sk))
+    sh.update(public_key_encode(Q))
```
dconnolly, line 1130: Why include the verifying key/message in the nonce generation when those are committed as part of the challenge?

dconnolly: Oh there's more at line 1163

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+The `make_challenge()` function returns a 16-byte value. The signature
+value is the concatenation of `c` and the encoded scalar `s`, for
+a total length of 48 bytes.
+
+The per-signature nonce `k` is a scalar. The nonce may conceptually have
+value zero (though this is very improbable); thus, the commitment `R`
+may be the identity element. This is why `R` must be encoded with
+`group_encode()` and not `public_key_encode()` (public keys *cannot* be
+the identity element).
```
dconnolly, line 1161: 👍

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+If a cryptographically secure random generator is available, then `k`
+may be obtained by simply getting 32 bytes out of that generator,
+interpreting them as a 256-bit integer, then reducing that integer
+modulo `r` (as in the private key generation process, this induces only
+negligible bias since `r` is close enough to a power of two). However,
+in practical implementations, a properly secure random generator is not
+necessarily available at signature generation time; the
+`generate_nonce()` function instead uses the private key, public key,
+message and optional extra seed to produce `k` in a safe way without
+requiring a random generator.
```
dconnolly, line 1176: You allow no attempt at sourcing randomness at all?

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+empty byte sequence, of length 0), then `generate_nonce()` is fully
+deterministic (the same signature value is always obtained for a given
+message and a given private key); this is safe. However, additional
+protection against some active attacks that introduce glitches in the
+computation through physical means can be obtained with some
+non-determinism. Such non-determinism is achieved by adding some varying
+bytes as `seed`. The security does not depend on the randomness of the
+seed; a randomly generated seed with a non-secure generator, or even a
+fully predictable seed, such as a sequence number or the current time,
+still yields a fully safe per-signature nonce.
```
dconnolly, line 1188: Hm.

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+The ability to provide the seed explicitly allows for testing the
+nonce generation process through test vectors. Since `generate_nonce()`
```
dconnolly, line 1191: 👍

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+encodes the seed length (in bytes) over 8 bytes, the seed may be
+an arbitrary sequence of up to `2^64-1` bytes.
+
+#### Signature Generation
```
dconnolly, line 1195:
```suggestion
#### Signature Verification
```

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+The scalar `s` can be split into a low and a high halves (`s0` and `s1`,
+respectively), with `s = s0 + s1*(2^128)`, which allows rewriting the
+computation of `R` as:
+```
+R = s0*G + s1*((2^128)*G) + c*(-Q)
+```
+which is a linear combination of three points (`G`, `(2^128)*G` and `Q`)
+with coefficients that all fit on 128 bits each. In a double-and-add
+alglorithm, the point doublings for the three multiplications can be
```
dconnolly, line 1228:
```suggestion
algorithm, the point doublings for the three multiplications can be
```

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+#### Signature Security
+
+The signatures described above formally provide the same security level
+as the group itself, i.e. "127 bits" (arguably equivalent to "128 bits"
+in the same sense as, for instance, Curve25519 is argued to offer
+"128-bit security" with a subgroup order of "only" about `2^252`
+elements). The use of a half-width challenge `c` was already suggested
+by Schnorr in his seminal paper; the security properties and requirements
+on the hash function (here, BLAKE2s) were thoroughly analyzed by [Neven,
+Smart and Warinschi][].
```
dconnolly, line 1249:
```suggestion
Smart and Warinschi].
```

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+It has been [noted][] that while the construction with a half-width
```
dconnolly, line 1251:
```suggestion
It has been [noted] that while the construction with a half-width
```

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+challenge is safe *as a signature scheme*, some advanced protocols try
+to also use signatures as commitments on a unique message: these
+protocols assume that it is infeasible for the signer to produce a
+public key `Q`, a signature `sig` and two distinct messages `M0` and
+`M1` such that `verify(Q, sig, M0)` and `verify(Q, sig, M1)` both return
+`True`. With the half-width challenges used in the signature scheme
+specified here, it is possible to produce such `Q`, `sig`, `M0` and `M1`
+with effort about `2^64` operations (i.e. a substantial but not
+infeasible computation). Formally speaking, the scheme achieves normal
```
dconnolly, line 1260: 👀

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+unforgeability but not semi-strong unforgeability. This does *not*
+contradict non-repudiation: in such a scenario, the signer truly signed
+both messages `M0` and `M1`, and can truthfully be held accountable to
+both. However, this lack of message unicity can complicate and even
+formally weaken some protocols such as distributed threshold signature
+schemes with some potentially actively dishonest signers. For "normal"
+usages of signatures (e.g. for authentication or certification
+purposes), this is a non-issue.
```
dconnolly, line 1268: 👀

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+  - While private and public keys used here have the same format as
+    private and public keys used for signatures, it is **not
+    recommended** to use the same private/public key pair for both key
+    exchange and signatures. In general, the two activities call for key
+    pairs with distinct, incompatible lifecycle properties (simply put,
+    keys used for encryption should normally be backed up to prevent
+    data loss in case of key loss, while keys used for signatures must
+    not be backed up, since their value relies in their exclusive
+    control by the purported signer). Moreover, possible interactions
+    between the two schemes are not entirely explored.
```
dconnolly, line 1303: 👍

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+The process below describes how a party uses its own key pair (`sk` and
+`Q`) along with the *encoded* public key received from the peer
+(`peer_pk`) to compute the shared secret. The function returns two
```
dconnolly, line 1307:
```suggestion
(`peer_PK`) to compute the shared secret. The function returns two
```

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+
+    # Compute the shared point.
+    P = point_mul(peer_Q, sk)
+    eP = group_encode(P)
+
+    # Get the two encoded public keys, in lexicographic order.
+    pk1 = public_key_encode(Q)
+    pk2 = peer_pk
+    ipk1 = int.from_bytes(pk1, byteorder='big')
+    ipk2 = int.from_bytes(pk2, byteorder='big')
```
dconnolly, line 1337: big???

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+outsiders. The point of producing such a fake secret key is to allow the
+use of the key exchange in some (rather contrived) scenarios in which
+attackers cannot *see*, but can *alter* public keys in transit, and may
+obtain some extra information from knowledge of whether the altered key
+was valid or not. Computing a fake but indistinguishable secret key can
+provide this extra security feature. In most normal uses of
+Diffie-Hellman, that feature is not needed.
+
+```python
+def ECDH(sk, Q, peer_pk):
```
dconnolly, line 1321: is this 'ecdh' or 'just' DH when over prime-order groups that just happen to use elliptic curves under the hood

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+    # Get the two encoded public keys, in lexicographic order.
+    pk1 = public_key_encode(Q)
+    pk2 = peer_pk
+    ipk1 = int.from_bytes(pk1, byteorder='big')
+    ipk2 = int.from_bytes(pk2, byteorder='big')
+    if ipk1 > ipk2:
+        (pk1, pk2) = (pk2, pk1)
+
+    # Compute the shared secret key.
+    sh = hashlib.blake2s()
```
dconnolly, line 1342: Any chance we want to add more domain separation on all these invocations of Blake2s

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+        sh.update(b'\x53')    # One byte of value 0x53
+        sh.update(eP)
+    else:
+        sh.update(b'\x46')    # One byte of value 0x46
+        sh.update(private_key_encode(sk))
+    return (sh.digest(), peer_pk_good)
+```
+
+The derivation of the shared secret group element `P` into a secret key
+appropriate for further symmetric cryptograhic operations uses an
```
dconnolly, line 1355:
```suggestion
appropriate for further symmetric cryptographic operations uses an
```

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+invocation of the BLAKE2s hash function. The two exchanged public keys
+are also used as input to the hashing process; since both parties should
+obtain the same key, the two public keys are inserted in lexicographic
+order (lowest key first). The pseudocode applies lexicographic ordering
```
dconnolly, line 1359: 👍

jq255.md
```diff
@@ -0,0 +1,1425 @@
...
+by interpreting both sequences of bytes as integers with the
+*big-endian* notation, and then comparing them as integers; when using
+languages without direct access to big integers, a simple byte-by-byte
+comparison may also be performed.
```
dconnolly, line 1363: 👍

jq255.md
dconnolly: @str4d @FiloSottile

jq255.md
pornin: I will detail here (I was not clear above): I use _mathematical notation_, in monospace font (with backquotes in the Markdown syntax), and I use _pseudocode_ (which is in fact real Python/Sage code), in code blocks (rendered in highlighted and coloured blocks). The pseudocode uses the `**` operator because Python somewhat requires it (the `^` operator being a XOR instead). Normally I did not use `**` outside of pseudocode. The `^` character can still occur in _code comments_ within code blocks.

jq255.md
pornin: It's really `(2*P)+N` that I meant here. But otherwise yes, as an implementer you have the choice between `2*P`, `2*(P+N)`, `(2*P)+N` and `(2*(P+N))+N`, which all lead to the same group element. In XWJ coordinates, the switch from `P` to `P+N` means negating `w`, and replacing `x` with `b/x`; the latter change moves expressions from numerator to denominator and vice versa, and thus leads to different formulas, which may be better or worse from a performance point of view.

Indeed, for jq255e, the best performance seems to be obtained by computing `2*P`, while on jq255s it's better to use `(2*P)+N` (it reduces per-doubling cost from 1M+6S to 2M+4S).

jq255.md
pornin: I kept it inlined for two (weak) reasons:
1. It avoids having (X,W,J) triplets moving around in the API. It can lead to confusion if there are several competing systems of coordinates, and I prefer to keep them contained inside the xdouble function.
2. In practice (and in more aggressively optimized languages), keeping them inline helps substantially with performance.

All of this is debatable.

jq255.md
pornin: Elligator2 works on curves that, in short Weierstraß form, have equation `Y^2 = X^3 + A*X + B` with both `A` and `B` being non-zero. For jq255e, the short Weierstraß form leads to `B = 0`, and Elligator2 does not work for that curve, hence the need for something else.

The generic case for curves with `A = 0` or `B = 0` can be covered by finding an isogeny to another curve with `A' != 0` and `B' != 0`, doing the map-to-curve on that other curve, and using the isogeny to get back to our original target curve. This is explained in https://eprint.iacr.org/2019/403 . However, in the case of jq255e, the lowest-degree isogeny is rather complicated, with lots of constants and many multiplications for its computation. The method I use, which I initially found "by hand" when jq255e was called "do255e", leads to simpler code and is also kinda faster, especially on small microcontrollers.

jq255.md
pornin: There is a `seed` and you can choose the seed contents arbitrarily, including randomly, and that even yields benefits in terms of resistance against fault attacks. However, being able to provide a non-random seed allows reproducibility of test vectors.

Basically you _can_ provide a random seed; you do not _need_ to provide a random seed. There are situations where you prefer to inject randomness through the seed, and there are situations where you prefer a fully deterministic operation.

jq255.md
pornin: Purists would say that Diffie and Hellman only talked about the multiplicative subgroup of integers modulo a big prime, and that anything else is a different algorithm. I am using here "ECDH" mostly because it's an already known acronym with a lot of visibility (e.g. you see it in the names of TLS cipher suites) and it is not, technically, _wrong_ in our case, because this is really done over an elliptic curve; the only difference with the "usual" ECDH is in how the resulting shared point is turned into a shared key (plain ECDH uses the x coordinate of the point, so it maps P and -P to the same output; here, the group encoding is used, so P and -P are mapped to distinct outputs, but, _a contrario_, P and P+N are mapped to the same value).

jq255.md
pornin: It's really about lexicographic ordering. In Python, this is most easily done by converting the two values to big integers (with big-endian convention) and then comparing them. In a language such as Rust or Go, you would simply use a loop over the bytes to know which of the two values should go first into the hash function.

jq255.md
pornin: _Normally_ (if I didn't goof up) they should already be separated (the values they are computed over are necessarily distinct). Moreover, as was explicitly said, you should really not use the same private key for ECDH and for signatures; as long as you use distinct private keys, then you live in different worlds and there's no need for more domain separation.

jq255.md
pornin: Injecting the message is necessary, otherwise two distinct messages might be signed with the same nonce (in the deterministic setup) and that would be Bad. Injecting the public key as well protects some weird "remote signer" implementations in which the caller can designate a private key and a public key independently, and confusion at that level can lead to some key recovery attacks (I know that EdDSA does not do that, and it has been pointed out in sternly worded tweets).

jq255.md
pornin: `GFq()` is defined near the start of the "Group Element Operations" section (line 171 currently):

> `GFq(j)` converts the integer `j` to a field element by reducing `j` modulo `q`.

jq255.md
pornin: This `peer_pk` is the name of the parameter to the function in the pseudocode. Are you suggesting it making it uppercase in the pseudocode as well?

jq255.md
pornin: I am trying to decouple the "outside" hash function for pre-hashing, on which I have little control, from the "inside" hash function, which is not visible in the API. The idea is that it should be possible to process the bulk of the data with anything that has good performance (e.g. BLAKE3, if that's your thing, or SHA3 if you are a die-hard NIST fan) without requiring the signature engine itself to support gazillions of possible hash functions. A useful model here is that of a smartcard or a HSM: the bulk hashing is done externally, in the calling machine, and the smartcard or HSM only receives the resulting hash value.

jq255.md
pornin: There's always some ambiguousness about the "sign" term, because it designates both the overall signature system, and the action of producing a signature. That's how you end up with functions called something like `sign_verify()` and `sign_sign()`, and that is ugly. Here I went for the traditional characterization of keys with their private/public status. If we want to name them for their intended use, we end up with "signature generation key" and "signature verification key", but then there's the unfortunate collision with key generation, and we have "signature generation key generation", which, in terms of ugliness, is even worse.

jq255.md
pornin: I have added a formal definition of `UNIFORM_SIZE(r)`, which is the number of random bytes you need so that the reduce-modulo-r process yields something which is indistinguishable from uniform selection. I have included some formulas that should be fine for every possible r, and make `UNIFORM_SIZE(r) = 32` for both jq255e and jq255s. Then I added references to that function where appropriate (private key generation, and signature nonce generation). Thus, I get the optimization for my case, but I still document the general case so that extension to larger curves is possible.

More generally, I think that _in the long term_ there should be a generic "sign over a prime-order group" specification that works for everything (with some interactions to unravel with both RFC 8032 and the FROST draft spec); when such a document exist, the jq255 spec can be modified to reference that document. In the short term, though, the jq255 spec may include the high-level algorithms (as it does presently), if only to be able to gather experience and comments from outsiders. Isn't it why C2SP specs are versioned, after all?

jq255.md
pornin: So basically there are _three_ layers here? Finite field at the bottom (with its encoding/decoding rules), group on top of it (and using the field encoding/decoding for its own encoding/decoding format), and then high-level protocols on top of it (using the group encoding/decoding, _and_ the scalar encoding/decoding process, which may be different).

(And then there's the hash-to-group process, which is "high-level" but needs to fiddle with converting input bytes into a field element, not a scalar...)

jq255.md
dconnolly: OK; since this may, in the worst case of bad seed randomness, fall back to a full deterministicly generated nonce, I'm fine with this. // @FiloSottile

jq255.md
```diff
@@ -1165,15 +1182,18 @@ The `generate_nonce()` function incarnates a process known as
...
-`generate_nonce()` function instead uses the private key, public key,
-message and optional extra seed to produce `k` in a safe way without
-requiring a random generator.
+`generate_nonce()` function shown above instead uses the private key,
+public key, message and optional extra seed to produce `k` in a safe way
+without requiring a random generator. Any cryptographically secure hash
+function with an output size of at least `UNIFORM_SIZE(r)` bytes can be
+used; in this specification, we use BLAKE2s, whose 32-byte output
+matches the value of `UNIFORM_SIZE(r)` for both jq255e and jq255s. For
+reproducibility of test vectors, it is RECOMMENDED to use BLAKE2s.
```
dconnolly, line 1196: 👍

jq255.md
```diff
@@ -154,6 +154,28 @@ jq255s:
...
+usl = max(lr, log2(abs(r - 2^lr)) + lr/2)
+UNIFORM_SIZE(r) = ceil(usl/8)
+```
+
+For both jq255e and jq255s, we obtain that `UNIFORM_SIZE(r) = 32`. In
+all generality, for groups with order less than `2^256`,
+`UNIFORM_SIZE()` may range up to 48, but the respective orders of jq255e
+and jq255s are sufficiently close to a power of two (namely `2^254`)
+that 32 bytes are enough to ensure the required indistinguishability
+from uniform selection.
```
dconnolly, line 177: 👍

jq255.md
dconnolly: 'and yet'

jq255.md
dconnolly: Gotcha

jq255.md
dconnolly: Yes, if it's a group element/point, uppercase; if it's a scalar/field element, lowercase

dconnolly: `PEER_PK` is shouty but

jq255.md
dconnolly: @FiloSottile thoughts?

jq255.md
dconnolly: 🙏✔️

jq255.md
dconnolly: Oh sorry, nevermind then

jq255.md
dconnolly: Very cool

jq255.md
pornin: Also: the XWJ -> EZUT conversions are distinct between jq255e and jq255s. The computation of `E` is really `E = 2*X - Z + a*T` and the two curves don't use the same `a` (`a = 0` for jq255e, and `a = 1` for jq255s), so that's an extra reason not to try to merge them into a single shared function.

jq255.md
pornin: Well technically it is an _encoded_ group element (or purported to be, at least) so maybe I should call it `peer_encoded_pk`.

jq255.md
```diff
@@ -0,0 +1,1412 @@
...
+with Atkin's algorithm:
+
+ 1. Compute: `c = (2*x)^((q-5)/8)`
+ 2. Compute: `d = 2*x*c^2`
+ 3. Compute the square root candidate: `z = x*c*(d - 1)`
+
+The obtained candidate MUST be checked (by comparing `z^2` with `x` and finding them equivalent).
+Moreover, if `z` is indeed a square root of `x`, then it MUST be
+normalized to a non-negative value (i.e. if `sgn(z) = 1`, then `z` must
+be replaced with `-z`). The use of this fixed convention for square
```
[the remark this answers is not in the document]
pornin: I have added a note on constant-time code in the "conventions" section, and a few references to it (e.g. right below here) using the CT_SELECT() notation (lifted from the Ristretto draft).

jq255.md
pornin: I have moved the field and group element encodings earlier in the document, while the scalar and key encoding rules remain in the "high-level" section, so as to follow the logical succession of these layers.
(Moving the code confused the GitHub UI about this conversation; it is now marked as "outdated".)

jq255.md
```diff
@@ -176,6 +176,24 @@ and jq255s are sufficiently close to a power of two (namely `2^254`)
...
+practice use a constant-time implementation strategy: the clause
+contents should be computed systematically, then their results either
+kept or discarded through a constant-time selection primitive that
+typically uses bitwise Boolean operations on machine words. In the
+text below, we will use the following to denote a constant-time selection
+of value `v` (if `cond` is true) and value `u` (if `cond` is false):
+```
+CT_SELECT(v IF cond ELSE u)
+```
+(This notation is imported from the [Ristretto255 draft].)
```
dconnolly, line 195: 👍

jq255.md
```diff
@@ -253,7 +271,10 @@ Moreover, if `z` is indeed a square root of `x`, then it MUST be
 normalized to a non-negative value (i.e. if `sgn(z) = 1`, then `z` must
 be replaced with `-z`). The use of this fixed convention for square
 roots is needed for proper interoperability of the map-to-group and
-hash-to-group operations.
+hash-to-group operations. If the source data is secret, then these
+operations should be implemented in constant-time; notably, the selection
+of the non-negative root should use systematic evaluation of `-z`,
+and a constant-time selection, as in: `CT_SELECT(-z IF sgn(z) = 1 ELSE z)`.
```
dconnolly, line 277: 👍

jq255.md
```diff
@@ -390,7 +467,10 @@ def group_to_field(P):
 
 Here, `iZ*U` and `iZ*E` are the `u` and `e` coordinates, respectively.
 The group element is really encoded as the `u` coordinate of the
-representant point whose `e` coordinate is non-negative.
+representant point whose `e` coordinate is non-negative. If the point to
+encode is secret (e.g. it is the "shared secret" from a Diffie-Hellman
+key exchange process) then the conditional negation of `v` SHOULD be
+implemented in constant-time: `v = CT_SELECT(-v IF sgn(iZ*E) == 1 ELSE v)`.
```
dconnolly, line 473: 👍

jq255.md
```diff
@@ -415,6 +495,30 @@ such element. Take care that `sqrt()` MUST return the non-negative root.
...
+    return field_to_group(v)
+```
+
+**Note:** When decoding, both the decoding of the bytes to a field
+element, and the decoding of the field element to a group element, may
+fail (i.e. return `False`). If a strictly constant-time process is
+required, that does not even leak whether the process succeeded or not,
+or the cause of the failure, then the test on the output of the
+`field_decode()` call can be delayed until the end of the function,
+invoking `field_to_group()` on some conventional value in that case.
```
dconnolly, line 520: 👍

jq255.md
```diff
@@ -753,7 +857,17 @@ This process becomes constant-time if all the steps are constant-time.
...
+`Q` in the loop can be expressed as follows:
+```python
+        Q = (GFq(1), GFq(1), GFq(0), GFq(0))
+        abs_digit = CT_SELECT(-sd[i] IF sd[i] < 0 ELSE sd[i])
+        for j in range(0, 16):
+            Qj = win[j]
+            Q = CT_SELECT(Qj IF abs_digit == j + 1 ELSE Q)
+        Qn = point_neg(Q)
+        Q = CT_SELECT(Qn IF sd[i] < 0 ELSE Q)
+```
```
dconnolly, line 870: 👍

jq255.md
```diff
@@ -830,66 +944,12 @@ expects the following API (which is compatible with Python's standard
...
-```
-
-Any application or outer protocol may reuse the most significant bit of
-the last byte to store some other data element, provided that the bit is
-restored to its proper value (0) before giving the byte sequence as
-argument to `field_decode()`.
+Encoding rules for group elements were detailed previously
+(functions `group_encode()` and `group_decode()`). High-level
+operations use specific encoding rules for scalars, private keys,
+and public keys.
```
dconnolly, line 952: 👍

jq255.md
pornin: Well I have renamed it to `peer_encoded_Q` (the decoded point was already `peer_Q`).

jq255.md
pornin: I have expanded the wording a bit to make things clearer.

jq255.md
```diff
@@ -580,9 +580,11 @@ the following:
...
 
 Depending on the used curve, it can be beneficial to combine some of
-these doublings with a switch to the other representant (i.e. computing
-`2*P+N` instead of `2*P`); this does not change the output as a group
-element, since we are free to use either of the two representants.
+these doublings with a switch to the other representant for either the
+input or the output point (i.e. computing `2*(P+N)`, `(2*P)+N` or even
+`(2*(P+N))+N` instead of `2*P`); this does not change the output as a
+group element, since we are free to use either of the two representants,
+but it can reduce the cost when working in XWJ coordinates.
```
dconnolly, line 587: 👍

jq255.md
```diff
@@ -1360,7 +1362,7 @@ for further cryptographic operations, typically symmetric encryption.
 
 The process below describes how a party uses its own key pair (`sk` and
 `Q`) along with the *encoded* public key received from the peer
-(`peer_pk`) to compute the shared secret. The function returns two
+(`peer_encoded_Q`) to compute the shared secret. The function returns two
```
dconnolly, line 1365: 👍

jq255.md
dconnolly: @FiloSottile @str4d any opinions on this naming or am I just thinking too hard about this

FiloSottile (approved):
I have not finished reading through everything, but I think this would be a great C2SP spec, and I'm happy to send comments along with everyone else while the spec is at v0. Thank you!

PR #31, opened 2023-05-10
age-piv-p256 recipient stanza

remko:
This is an initial draft of a specification for the age piv-p256 recipient stanza, as used by [age-plugin-yubikey](https://github.com/str4d/age-plugin-yubikey) and [age-plugin-se](https://github.com/remko/age-plugin-se).

### DIFF
--- a/age-piv-p256.md
+++ b/age-piv-p256.md
@@ -1,0 +1,100 @@
+# age piv-p256 recipient stanza
+
+[c2sp.org/age-piv-p256](https://c2sp.org/age-piv-p256)
+
+## Introduction
+
+This document specifies the age piv-p256 recipient stanza. This stanza
+is used for encryption using NIST P-256 keys stored on hardware tokens
+or PIV devices.
+
+Apart from using the NIST P-256 curve, the piv-p256 recipient stanza
+differs from the native [X25519 recipient
+stanza](https://age-encryption.org/v1#x25519-recipient-stanza) in that
+it includes a tag of the recipient, to allow the recipient to select the
+correct secret without the hardware token requesting user interaction.
+
+## Conventions used in this document
+
+ABNF syntax follows [RFC
+5234](https://www.rfc-editor.org/rfc/rfc5234.html) and [RFC
+7405](https://www.rfc-editor.org/rfc/rfc7405.html) and references the
+core rules in RFC 5234, Appendix B.1.
+
+The base64 encoding used throughout is the standard Base 64 encoding
+specified in [RFC 4648](https://www.rfc-editor.org/rfc/rfc4648.html),
+Section 4, without `=` padding characters (sometimes referred to as
+"raw" or "unpadded" base64). Encoders MUST generate canonical base64
+according to RFC 4648, Section 3.5, and decoders MUST reject
+non-canonical encodings and encodings ending with `=` padding
+characters.
+
+`||` denotes concatenation. `0x` followed by two hexadecimal characters
+denotes a byte value in the 0-255 range.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" in this document are to be interpreted as described in [BCP
+14](https://www.rfc-editor.org/info/bcp14) [RFC
+2119](https://www.rfc-editor.org/rfc/rfc2119.html) [RFC
+8174](https://www.rfc-editor.org/rfc/rfc8174.html) when, and only when,
+they appear in all capitals, as shown here.
+
+## Recipient stanza
+
+A piv-p256 recipient stanza has three arguments.
+
+    -> piv-p256 8u8pZg Aqa5k2Tm/BJGTzCno1W29dWaWBHkFchOztcIZ7aQo0Ca
+    CXE0N0KMI0BUE5WaYTLOxYv7aHB/5IQpAZKsk2C/yjU
+
+1.  The first argument is the fixed string `piv-p256`.
+
+2.  The second argument is the base-64 encdoded tag of the recipient,
+    consisting of the first 4 bytes of the SHA-256 hash of the
+    recipient:
+
+         tag = SHA-256(recipient)[0..4]
+
+3.  The third argument is the base64-encoded ephemeral share computed by
+    the recipient implementation as follows:
+
+         ephemeral secret = read(CSPRNG, 32)
+         ephemeral share = P256(ephemeral secret, basepoint)
+
+    A new ephemeral secret MUST be generated for each stanza and each
+    file.
+
+The body of the recipient stanza is computed by the recipient
+implementation as
+
+    salt = ephemeral share || recipient
+    info = "piv-p256"
+    shared secret = P256(ephemeral secret, recipient)
+    wrap key = HKDF-SHA-256(ikm = shared secret, salt, info)
+    body = ChaCha20-Poly1305(key = wrap key, plaintext = file key)
+
+where the ChaCha20-Poly1305 nonce is fixed as 12 0x00 bytes.
+
+The identity implementation MUST ignore any stanza that does not have
+`piv-p256` as the first argument, and MUST otherwise reject any stanza
+that has more or less than three arguments, or where the second argument
+is not a canonical encoding of a 4-byte value, or the third argument is
+not a canonical encoding of a 32-byte value. It MUST check that the body
+length is exactly 32 bytes before attempting to decrypt it.
+
+The identity implementation MUST ignore any stanza with a tag for which
+it does not have any corresponding secret.
+
+The identity implementation computes the shared secret as follows:
+
+    shared secret = P256(identity, ephemeral share)
+
+If the shared secret is all 0x00 bytes, the identity implementation MUST
+abort.
+
+Finally, it derives the key as above and decrypts the file key in the
+body.
+
+## Changelog
+
+- **v0.0.1**: Initial draft


age-piv-p256.md
```diff
@@ -0,0 +1,100 @@
...
+    -> piv-p256 8u8pZg Aqa5k2Tm/BJGTzCno1W29dWaWBHkFchOztcIZ7aQo0Ca
+    CXE0N0KMI0BUE5WaYTLOxYv7aHB/5IQpAZKsk2C/yjU
+
+1.  The first argument is the fixed string `piv-p256`.
+
+2.  The second argument is the base-64 encdoded tag of the recipient,
+    consisting of the first 4 bytes of the SHA-256 hash of the
+    recipient:
+
+         tag = SHA-256(recipient)[0..4]
```
remko, line 56: I was wondering if the tag should be salted, to hide relations between stanzas (in different files) for the same recipient

age-piv-p256.md
```diff
@@ -0,0 +1,100 @@
...
+
+The identity implementation MUST ignore any stanza with a tag for which
+it does not have any corresponding secret.
+
+The identity implementation computes the shared secret as follows:
+
+    shared secret = P256(identity, ephemeral share)
+
+If the shared secret is all 0x00 bytes, the identity implementation MUST
+abort.
```
FiloSottile, line 93:
```suggestion
If the shared secret is the point at infinity, the identity implementation
MUST abort.
```

The identity is going to be represented differently by different libraries, and often not as a long string of 0x00.

age-piv-p256.md
```diff
@@ -0,0 +1,100 @@
...
+    A new ephemeral secret MUST be generated for each stanza and each
+    file.
+
+The body of the recipient stanza is computed by the recipient
+implementation as
+
+    salt = ephemeral share || recipient
+    info = "piv-p256"
+    shared secret = P256(ephemeral secret, recipient)
+    wrap key = HKDF-SHA-256(ikm = shared secret, salt, info)
```
FiloSottile, line 73: If we make other backward-incompatible changes, we should swap salt and info here. HKDF is... very theoretically documented, and this is a mistake I made in the original design. It doesn't have practical consequences, and I think one can prove the security of the current design, but it makes applying more sophisticated proofs annoying.

age-piv-p256.md
```diff
@@ -0,0 +1,100 @@
...
+    consisting of the first 4 bytes of the SHA-256 hash of the
+    recipient:
+
+         tag = SHA-256(recipient)[0..4]
+
+3.  The third argument is the base64-encoded ephemeral share computed by
+    the recipient implementation as follows:
+
+         ephemeral secret = read(CSPRNG, 32)
+         ephemeral share = P256(ephemeral secret, basepoint)
```
FiloSottile, line 62: "P256" is not a well-defined ECDH operation the way X25519 is.

We should define it somewhere as scalar multiplication over P-256, with SEC 1 point decoding and encoding. (Assuming that's what we are doing? But then why is the third argument 32 bytes?)

age-piv-p256.md
FiloSottile: Hmm what do you have in mind?

Maybe `x = CSPRNG(4); tag = x || HMAC(recipient, x)[0..4]` so that only someone who knows the recipient can link files?

That is indeed a nice step forward in privacy! Might be worth making a backwards-incompatible change. @str4d, what do you think?

age-piv-p256.md
```diff
@@ -0,0 +1,100 @@
...
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" in this document are to be interpreted as described in [BCP
+14](https://www.rfc-editor.org/info/bcp14) [RFC
+2119](https://www.rfc-editor.org/rfc/rfc2119.html) [RFC
+8174](https://www.rfc-editor.org/rfc/rfc8174.html) when, and only when,
+they appear in all capitals, as shown here.
+
+## Recipient stanza
+
+A piv-p256 recipient stanza has three arguments.
```
FiloSottile, line 45: If we make other changes, we should drop "piv" from the name.

Maybe tagged-p256? Or hw-p256?

I assume @Foxboron could use it for age-plugin-tpm, too.

I wonder if we should standardize the recipient (not identity) format, and add native support to age(1) so that one doesn't need the plugin to encrypt to PIV/TPM/SE/any other P-256 identity.

/cc @str4d

age-piv-p256.md
tarcieri: I think it's pretty common to just use the x-coordinate, similar to X25519

FiloSottile: > I think it's pretty common to just use the x-coordinate, similar to X25519

For the shared ECDH output yes, but the input is usually either a compressed or an uncompressed point, no? x-coordinate-only ECDH like in X25519 is very uncommon on P-256. (I recently surveyed a bunch of ECDH and ECIES implementations while designing the crypto/ecdh.)

tarcieri: Which spec uses a SEC1-encoded point as the input?

TLS 1.3 (by way of IEEE1363) uses the x-coordinate, for example:

https://datatracker.ietf.org/doc/html/rfc8446#section-7.4.2

> [7.4.2](https://datatracker.ietf.org/doc/html/rfc8446#section-7.4.2).  Elliptic Curve Diffie-Hellman
>
>   For secp256r1, secp384r1, and secp521r1, ECDH calculations (including
>   parameter and key generation as well as the shared secret
>   calculation) are performed according to [[IEEE1363](https://datatracker.ietf.org/doc/html/rfc8446#ref-IEEE1363)] using the
>   ECKAS-DH1 scheme with the identity map as the key derivation function
>   (KDF), so that the shared secret is the x-coordinate of the ECDH
>   shared secret elliptic curve point represented as an octet string.

FiloSottile:
That describes the output not the input. I’m from mobile so I can’t look now but you’d find it described under the key share section probably.

tarcieri:
@FiloSottile whoops, I think I originally meant to reply [here](https://github.com/C2SP/C2SP/pull/31/files/4c6c0ad4a6993a83d635a2e74602d4128c3e9ce3#diff-20a22797eec96149dd04f811a5965cc76b5ec3f6958525fc5206b7fd2ac5979a) instead (and I'll go ahead and leave a comment there as well)

age-piv-p256.md
tarcieri: My comments about using the x-coordinate for ECDH with P-256 (and really all the NIST P-curves) were supposed to be in reply to this comment. I guess I must've clicked the wrong one in my previous replies.

age-piv-p256.md
tarcieri: My bad, I was originally trying to reply to [this comment](https://github.com/C2SP/C2SP/pull/31/files/4c6c0ad4a6993a83d635a2e74602d4128c3e9ce3#r1305437492)

age-piv-p256.md
Foxboron: I can probably use this, I wasn't aware of this PR work when I originally wrote `age-plugin-tpm`.

age-piv-p256.md
FiloSottile: Replying to https://github.com/C2SP/C2SP/pull/31#discussion_r1305720851.

The ECDH output is the x-coordinate. The ECDH input (that is, the third stanza argument here) is `0x04 || x || y`, which is SEC 1. 

https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8.2

age-piv-p256.md
remko: Native support sounds neat.

- You don't need to install a plugin to encrypt to these plugins. For encrypting to a different person, it feels odd that you need to install `age-plugin-se` even if you don't have an Apple device. It also avoids the plugin discovery problem where you need to explain which plugin people have to download to encrypt a message to you.
- For `age-plugin-se`, I could drop all cross-platform support, since the parts that don't involve encryption are only useful on macOS. I can imagine there are other hardware encryption plugins that don't make sense on all platforms.
- It hides information about which type of token corresponds to a certain recipient
- If there's a backwards-incompatible change now, only half of the existing state machines need to be updated (the easy half)

age-piv-p256.md
remko: > Maybe `x = CSPRNG(4); tag = x || HMAC(recipient, x)[0..4]` so that only someone who knows the recipient can link files?

That's indeed what I had in mind, except I was thinking of putting `x` in a separate stanza argument instead of concatenating it (to avoid a custom split)

erincandescent: Is there a reason to prefer this construction instead of `HMAC(recipient, ephemeral share)[0..4]`?

Foxboron:
How can we move this forward? It would be nice to have a generic implementation on the `age` side of things :)

I'm contemplating implementing the spec as is into `age-plugin-tpm` but if the naming isn't settled that would maybe be a bit annoying?

age-piv-p256.md
remko: IANAC, but sounds nice and simple to me.

remko:
> How can we move this forward? It would be nice to have a generic implementation on the `age` side of things :)

I was hoping @str4d could give some input on the document and the change suggestions.

The following backwards-incompatible changes for the recipient stanza format are on the table:

- Change the tag to offer better privacy (add randomness to the tag)
- Swap salt and info in HKDF-SHA-256

One possible way forward is to keep piv-p256 as-is, use the current version of this document (with the suggested editorial changes) as a historical description of the spec, and incorporate the changes into a different recipient format under a different recipient name (tagged-p256, hw-p256, p256), which new implementations should use.

Orthogonally(?), there is the suggestion to standardise the recipient format and implement it in age.

Foxboron:
I think it would make sense to leave the current `piv-p256` as-is and maybe just standardize this as `p256`?

FiloSottile:
Thank you for getting this started. I think it's becoming clear that a standard tagged P-256 recipient type would be valuable across multiple plugins. I propose we

1. change the tag to `HMAC(recipient, ephemeral share)[0..4]`;
2. swap HKDF salt and info;
3. name the new stanza `p256-tag`;
4. also specify a `p256tag` recipient plugin type `age1p256tag1...`;
5. implement native support for `p256tag` recipients in age(1) and maybe rage(1);
6. migrate plugins to the new stanza and recipient, but have them support decrypting the old one stanza ~forever.

Plugins might choose to provide a recipient-only `age-plugin-p256tag` binary for compatibility with older versions of age.

How does that sound @str4d, @Foxboron, @remko?

Foxboron:
I think this sounds great!

FiloSottile:
Actually, do we need a new spec for this? Why not add a section to c2sp.org/age?

Foxboron:
Considering we have two other "native recipient types" I don't think a separate standard is required really. It might be nicer to ensure the age spec doesn't become super long?

remko:
> Actually, do we need a new spec for this? Why not add a section to c2sp.org/age?

Works for me. How do you suggest to proceed with this spec: It can be released as a 'historical' spec for documenting existing behavior of plugins, with a note forwarding to the new recipient type and recommend to implement that. Or do you suggest we drop it altogether, and not document anything?

remko:
> I propose we

FYI, the current HEAD of age-plugin-se has an implementation of this proposal (behind a hidden `--recipient-type=p256tag` flag).

Foxboron:
@FiloSottile are you going to propose the addition to the age spec or would it be useful if someone else finds the time to draft it?

FiloSottile:
> @FiloSottile are you going to propose the addition to the age spec or would it be useful if someone else finds the time to draft it?

If anyone finds the time to make a PR against the age spec, that would be awesome!

FiloSottile:
> How do you suggest to proceed with this spec: It can be released as a 'historical' spec for documenting existing behavior of plugins, with a note forwarding to the new recipient type and recommend to implement that. Or is it not worth the trouble (and the risk of people implementing the wrong spec), and we drop it altogether, and not document anything?

That's a good question, it would be nice for future-proofing to have the spec somewhere I guess. @C2SP/stewards how do you feel about "historical" specs?

remko-bw:
> If anyone finds the time to make a PR against the age spec, that would be awesome!

The PR can be found here: #96

remko:
> If anyone finds the time to make a PR against the age spec, that would be awesome!

The PR can be found here: #96

PR #35, opened 2023-06-21
Switch around A & P instead of changing arrow direction

quite:
Maybe it's just me, but I find it much more readable this way.


str4d (approved):

str4d:
I don't particularly mind either way, so if the feedback is that this is more readable then let's merge it. And if this makes readability worse for other people, we can address that when a corresponding issue gets opened.

PR #36, opened 2023-07-08
XAES-256-GCM: preliminary draft

FiloSottile:


XAES-256-GCM.md
```diff
@@ -0,0 +1,45 @@
+This document specifies the XAES-256-GCM authenticated encryption with additional data algorithm, based on the composition of a standard NIST SP 800-108r1 KDF and the standard NIST AES-256-GCM AEAD.
```
str4d, line 1:
```suggestion
This document specifies the XAES-256-GCM authenticated encryption with additional data algorithm, based on the composition of a standard [NIST SP 800-108r1] KDF and the standard NIST AES-256-GCM AEAD.
```

XAES-256-GCM.md
```diff
@@ -0,0 +1,45 @@
...
+
+The XAES-256-GCM inputs are a 256-bit key, a 192-bit nonce, a plaintext of up to approximately 64GiB, and additional data of up to 2 EiB.
+
+Unlike AES-256-GCM, the XAES-256-GCM nonce can be randomly generated for a virtually unlimited number of messages (2⁸⁰ messages with collision risk 2⁻³²). Only a 256-bit key version is specified, which provides a comfortable multi-user security margin. Like AES-256-GCM, XAES-256-GCM is not nonce misuse-resistant, nor is it key-committing. Compared to AES-256-GCM, XAES-256-GCM requires three extra invocations of the AES-256 function, but has otherwise the same performance profile.
+
+## Overview
+
+XAES-256-GCM derives a subkey for use with AES-256-GCM from the input key and half the input nonce using a NIST SP 800-108r1 KDF.
```
str4d, line 9:
```suggestion
XAES-256-GCM derives a subkey for use with AES-256-GCM from the input key and half the input nonce using a [NIST SP 800-108r1] KDF.
```

XAES-256-GCM.md
```diff
@@ -0,0 +1,45 @@
...
+
+The counter-based KDF (NIST SP 800-108r1, Section 4.1) is instantiated with CMAC-AES256 (NIST SP 800-38B) and the input key as *Kin*, the ASCII letter `X` (0x58) as *Label*, and the first 96 bits of the input nonce as *Context* (as recommended by NIST SP 800-38B, Section 4, point 4), to produce a 256-bit derived key.
```
str4d, line 11:
```suggestion
The counter-based KDF ([NIST SP 800-108r1, Section 4.1]) is instantiated with CMAC-AES256 ([NIST SP 800-38B]) and the input key as *Kin*, the ASCII letter `X` (0x58) as *Label*, and the first 96 bits of the input nonce as *Context* (as recommended by [NIST SP 800-38B], Section 4, point 4), to produce a 256-bit derived key.
```

I also cannot find the Section 4, point 4 reference in NIST SP 800-38B.

XAES-256-GCM.md
```diff
@@ -0,0 +1,45 @@
...
+
+Note that with a counter (*i*) size of 16 bits and omitting the optional *L* field, the AES-CMAC input totals 128 bits, which fits into a single block, mitigating the key control security issue described in NIST SP 800-38B, Section 6.7. It would in theory be possible to shrink *i* to 8 bits to fit a 8 bits *L*, but some implementations unfortunately fix the *L* size to 32 bits.
```
str4d, line 13: NIST SP 800-38B does not contain a Section 6.7 (that part only goes up to 6.3), and I cannot find an alternative mention of a key control security issue.

XAES-256-GCM.md
```diff
@@ -0,0 +1,45 @@
...
+
+1. *L* = AES-256ₖ(0¹²⁸)
+2. If MSB₁(*L*) = 0, then *K1* = *L* << 1;
+   Else *K1* = (*L* << 1) ⊕ 0¹²⁰10000111
+3. *M1* = 0x00 || 0x01 || `X` || 0x00 || *N*[:12]
+4. *M2* = 0x00 || 0x02 || `X` || 0x00 || *N*[:12]
+5. *Kₓ* = AES-256ₖ(*M1* ⊕ *K1*) || AES-256ₖ(*M2* ⊕ *K1*)
+6. *Nₓ* = *N*[12:]
+
+Steps 1 and 2 reproduce the CMAC subkey generation specified in NIST SP 800-38B, Section 6.1. Note that only *K1* is needed.
```
str4d, line 39:
```suggestion
Steps 1 and 2 reproduce the CMAC subkey generation specified in [NIST SP 800-38B, Section 6.1]. Note that only *K1* is needed.
```

XAES-256-GCM.md
```diff
@@ -0,0 +1,45 @@
...
+
+Steps 2 and 3 compose the PRF inputs for *i* = 1, 2 according to NIST SP 800-108r1, Section 4.1.
```
str4d, line 41:
```suggestion
Steps 3 and 4 compose the PRF inputs for *i* = 1, 2 according to [NIST SP 800-108r1, Section 4.1].
```

[NIST SP 800-108r1, Section 4.1](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-108r1.pdf#%5B%7B%22num%22%3A79%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C70%2C300%2C0%5D) specifies the PRF inputs with a 2-byte suffix committing to the output length in bits. In this case the "output" here has constant length, so the length commitment is unnecessary (and dropping it lets the input fit into an AES-256 block); it might be worth mentioning this here.

XAES-256-GCM.md
```diff
@@ -0,0 +1,45 @@
...
+
+Step 5 applies CMAC twice to the two single-block messages to derive the KDF output.
+
+*Kₓ* and *Nₓ* are then used as the AES-256-GCM key and nonce, respectively.
```
str4d, line 45:
```suggestion
*Kₓ* and *Nₓ* are then used as the AES-256-GCM key and nonce, respectively.

[NIST SP 800-38B]: https://csrc.nist.gov/publications/detail/sp/800-38b/final
[NIST SP 800-108r1]: https://csrc.nist.gov/publications/detail/sp/800-108/rev-1/final
```

FiloSottile:
Oh my god PDF deep links, mind blown.

XAES-256-GCM.md
FiloSottile: It was supposed to be NIST SP 800-108r1, thanks.

XAES-256-GCM.md
FiloSottile: Ditto.

XAES-256-GCM.md
FiloSottile: That's the *L* field mentioned in the third Overview paragraph, now expanded.

XAES-256-GCM/reference.go
```diff
@@ -0,0 +1,84 @@
...
+		return nil, errors.New("xaes256gcm: bad key length")
+	}
+
+	x := new(xaes256gcm)
+
+	x.c, _ = aes.NewCipher(key)
+	x.c.Encrypt(x.k1[:], x.k1[:])
+
+	var msb byte
+	for i := len(x.k1) - 1; i >= 0; i-- {
```
bleichenbacher-daniel, line 32: This loop looks wrong to me assuming that is should implement the same shift left operation as for example used in EAX. the msb for k1[i] is used to compute the new value of k[i]. I would expect here that the msb should be used in the next iteration in the computation of k[i-1].

FiloSottile: Good catch! Thank you, I'll fix this. Part of why I didn't merge this yet is that I wanted to finish writing the OpenSSL implementation based on the high-level CMAC API, to find any mistakes in the low-level Go.

bleichenbacher-daniel: I also have an independent implementation based on the spec.

FiloSottile: Fixed the bug, finished the OpenSSL implementation, and added test vectors. Could you check that the vectors agree with your implementation? Thank you!

PR #37, opened 2023-08-05
age-plugin: Add labels extension to `recipient-v1`

str4d:
Designed with @FiloSottile.

This extension to `recipient-v1` enables clients to require that plugins provide a set of labels at encryption time. These labels are used to restrict which stanzas can appear together in the same file header.

We don't make any changes to `identity-v1` because decryption time restrictions are pointless: an attacker can drop stanzas or reuse the single stanza.

This can be used for
- Authentication (auth X25519, scrypt) by setting a random label.
- PQC by setting a well-known label.
- Corp approval by setting a private but potentially shared label.

### DIFF
--- a/age-plugin.md
+++ b/age-plugin.md
@@ -216,8 +216,8 @@ Four commands are defined for this phase:
   wrap all the file keys to.
   - `IDENTITY` is the Bech32 encoding of an identity.
 - `(wrap-file-key; FILE_KEY)` - a file key to be wrapped.
-- `(extension-labels)` - indicates that the client is using "labels" to constrain
-  the recipient stanzas that it will accept.
+- `(extension-labels)` - indicates that the client supports "labels" to constrain
+  the recipient stanzas and will accept a `labels` command in phase 2.
   - A label is defined in ABNF by the `argument` type from the [v1 age header][].
   - Clients MUST check that all recipient stanzas wrapping a given file key have
     the exact same label set. Clients MUST NOT permit partial overlapping sets.
@@ -225,7 +225,7 @@ Four commands are defined for this phase:
     treat the plugin's recipient stanzas has having an empty label set.
   - The `x25519` recipient stanza has an implicit empty label set.
   - The `scrypt` recipient stanza has an implicit label set containing a single
-    random label.
+    random label. In other words, it can't be combined with any other stanza.
 
 The plugin indexes recipients, identities, and file keys in the order received
 (starting from 0). The two may be interleaved by the client, with no semantic
@@ -285,12 +285,12 @@ The following commands and responses are defined for this phase:
   recipient stanzas it generates in this phase.
   - Labels can have any value, but usually take one of several forms:
     - *Common public label* - used by multiple plugins to permit their recipient
-      stanzas to be used together. Examples include:
+      stanzas to be used only together. Examples include:
       - `postquantum` - indicates that the recipient stanzas being generated are
         postquantum-secure, and that they can only be combined with other stanzas
         that are also postquantum-secure.
     - *Common private label* - used by plugins created by the same private entity
-      to permit their recipient stanzas to be used together. For example, private
+      to permit their recipient stanzas to be used only together. For example, private
       plugins used in a corporate environment could all send the same private label
       in order to prevent compliant age clients from simultaneously wrapping file
       keys with other plugins.
@@ -299,9 +299,9 @@ The following commands and responses are defined for this phase:
       that is only encrypted to a single recipient stanza.
   - Plugins MUST NOT send duplicate labels, and MUST NOT send a `labels` command
     more than once.
-  - The order of labels in the command is undefined. Clients MUST treat them as a
-    set.
-  - Response is `(ok)`.
+  - The order of labels in the command is irrelevant. Clients MUST treat them as an
+    unordered set.
+  - Response is `(ok)` or an error.
   - The plugin SHOULD send this command as early as it can, to avoid any expensive
     computations required to generate the recipient stanzas if the client is going
     to raise an error (because the label set does not match those associated with
--- a/age-plugin.md
+++ b/age-plugin.md
@@ -221,6 +221,8 @@ Four commands are defined for this phase:
   - A label is defined in ABNF by the `argument` type from the [v1 age header][].
   - Clients MUST check that all recipient stanzas wrapping a given file key have
     the exact same label set. Clients MUST NOT permit partial overlapping sets.
+  - If the plugin does not respond with a `labels` command in phase 2, clients MUST
+    treat the plugin's recipient stanzas has having an empty label set.
   - The `x25519` recipient stanza has an implicit empty label set.
   - The `scrypt` recipient stanza has an implicit label set containing a single
     random label.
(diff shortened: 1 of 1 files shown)


age-plugin.md
```diff
@@ -216,6 +216,16 @@ Three commands are defined for this phase:
...
+- `(extension-labels)` - indicates that the client is using "labels" to constrain
+  the recipient stanzas that it will accept.
+  - A label is defined in ABNF by the `argument` type from the [v1 age header][].
+  - Clients MUST check that all recipient stanzas wrapping a given file key have
+    the exact same label set. Clients MUST NOT permit partial overlapping sets.
+  - If the plugin does not respond with a `labels` command in phase 2, clients MUST
+    treat the plugin's recipient stanzas has having an empty label set.
+  - The `x25519` recipient stanza has an implicit empty label set.
+  - The `scrypt` recipient stanza has an implicit label set containing a single
+    random label.
```
str4d, line 228: In https://github.com/str4d/rage/pull/365#issuecomment-1666825131 I noted that we also need to decide on how file header grease (currently unspecified) interacts with the label set, because what it results in is extra stanzas.

The simplest approach, and the one compatible with the behaviour of pre-labels clients, is that grease has an implicit empty label set. This lets it be added alongside `x25519` and any pre-labels plugins, but not alongside `scrypt`. However, it means that plugins which start using labels would as a side-effect never have their headers greased. This might be fine (and in fact would probably be a nice side-effect for corp usage, where you probably don't want random extra stanzas), but would be annoying for e.g. postquantum-only headers.

FiloSottile: It's indeed a greasing loss, but probably worth it to keep the implementation general. Empty label set sounds right.

age-plugin.md
```diff
@@ -216,6 +216,16 @@ Three commands are defined for this phase:
   wrap all the file keys to.
   - `IDENTITY` is the Bech32 encoding of an identity.
 - `(wrap-file-key; FILE_KEY)` - a file key to be wrapped.
+- `(extension-labels)` - indicates that the client is using "labels" to constrain
+  the recipient stanzas that it will accept.
```
FiloSottile, line 220:
```suggestion
- `(extension-labels)` - indicates that the client supports "labels" to constrain
  the recipient stanzas and will accept a `labels` command in phase 2.
```

age-plugin.md
```diff
@@ -270,6 +280,32 @@ The following commands and responses are defined for this phase:
...
+      in order to prevent compliant age clients from simultaneously wrapping file
+      keys with other plugins.
+    - *Random label* - used by plugins that want to ensure their stanzas are not
+      used with any other recipient stanzas. This can be used to produce a file key
+      that is only encrypted to a single recipient stanza.
+  - Plugins MUST NOT send duplicate labels, and MUST NOT send a `labels` command
+    more than once.
+  - The order of labels in the command is undefined. Clients MUST treat them as a
+    set.
+  - Response is `(ok)`.
```
FiloSottile, line 304:
```suggestion
  - Response is `(ok)` or an error.
```

age-plugin.md
```diff
@@ -270,6 +280,32 @@ The following commands and responses are defined for this phase:
...
+  in phase 1, the plugin sends the set of labels it associates with the set of
+  recipient stanzas it generates in this phase.
+  - Labels can have any value, but usually take one of several forms:
+    - *Common public label* - used by multiple plugins to permit their recipient
+      stanzas to be used together. Examples include:
+      - `postquantum` - indicates that the recipient stanzas being generated are
+        postquantum-secure, and that they can only be combined with other stanzas
+        that are also postquantum-secure.
+    - *Common private label* - used by plugins created by the same private entity
+      to permit their recipient stanzas to be used together. For example, private
```
FiloSottile, line 293:
```suggestion
      to permit their recipient stanzas to be used only together. For example, private
```

age-plugin.md
```diff
@@ -270,6 +280,32 @@ The following commands and responses are defined for this phase:
     be mapping the stanzas to specific recipients, and will cache stanzas for
     separate files until the state machine completes.
   - Response is `(ok)`.
+- `(labels, [LABEL [LABEL...]])` - if the client sent a `extension-labels` command
+  in phase 1, the plugin sends the set of labels it associates with the set of
+  recipient stanzas it generates in this phase.
+  - Labels can have any value, but usually take one of several forms:
+    - *Common public label* - used by multiple plugins to permit their recipient
+      stanzas to be used together. Examples include:
```
FiloSottile, line 288:
```suggestion
      stanzas to be used only together. Examples include:
```

FiloSottile (dismissed):

age-plugin.md
```diff
@@ -270,6 +280,32 @@ The following commands and responses are defined for this phase:
...
+      in order to prevent compliant age clients from simultaneously wrapping file
+      keys with other plugins.
+    - *Random label* - used by plugins that want to ensure their stanzas are not
+      used with any other recipient stanzas. This can be used to produce a file key
+      that is only encrypted to a single recipient stanza.
+  - Plugins MUST NOT send duplicate labels, and MUST NOT send a `labels` command
+    more than once.
+  - The order of labels in the command is irrelevant. Clients MUST treat them as an
+    unordered set.
+  - Response is `(ok)` or an error.
```
str4d, line 304: @FiloSottile you suggested the "or an error" here. What exactly did you intend to be returned to the plugin? In the rest of the plugin protocol we only define `(ok)` and `(fail)`, and the latter is _not_ an error resulting in a protocol abort. Did you mean that the client should abort the protocol and return an error _to the user_ if a MUST / MUST NOT is violated? Or did you intend to convey an error to the plugin?

str4d: I agree with dropping the error condition (reverting your suggestion that I merged before noticing this issue).

I think the next bullet point was about avoiding the more intensive plugin operations if they were going to be pointless because the client was about to abort the entire encryption process. Labels are a bit different than user interactions in that it is unambiguous to a label-aware client that the subsequent plugin operations are going to be pointless. But the client dropping the connection on its own isn't necessarily going to halt them; only the client closing the plugin subprocess will do so, and that won't necessarily actually halt the operations (if the plugin called out to a long-running service and was just waiting on a response). And in my client plugin implementation I also didn't have a great way to halt the process early, indeed due to designing my APIs to enforce the plugin being the controller of that phase. So yeah, I think dropping the next bullet point would also be fine.

str4d: I went to implement this and had a look at the commit history. Turns out we did originally allow returning `(fail)` from here if certain `labels` invariants weren't upheld, but replaced that with `MUST NOT` language in f2c4053fa4921ae72b89be45e925b187807fedcd.

Something we _could_ therefore do is return `(fail)` if the client knows that the plugin stanzas will not be compatible, but not _require_ that the client inform the plugin at this stage. That would be compatible with `WrapWithLabels` (the client would not ever tell the plugin what the existing label set it is working to is, just that it's incompatible with the plugin's label set, so `WrapWithLabels` would be used with a client impl that never returns `(fail)`), while clients that want to short-circuit could go around that abstraction (by not implementing plugins on the client side using the `Recipient` interface).

However, this would also be a backwards-compatible change: existing clients would never return `(fail)` to new plugins, and new clients returning `(fail)` would cause an error with existing plugins which is what we want to happen (it would maybe just be an unexpected error message). So I think on balance it's fine to drop `(fail)`.

str4d: @FiloSottile I've pushed a commit making the change, with an extensive comment. Please review the comment and resolve this conversation if you're satisfied.

FiloSottile: Yep, thank you for documenting this.

PR #38, opened 2023-08-24
vrf-r255: update references to RFC 9381

FiloSottile:


str4d (approved):
ACK. I confirmed that there were no [editorial changes between draft 15 and RFC 9381](https://author-tools.ietf.org/iddiff?url1=draft-irtf-cfrg-vrf-15&url2=rfc9381&difftype=--html) that interfere with this spec.

PR #41, opened 2023-10-26
chacha8rand: add spec for the new Go global CSPRNG

FiloSottile:
/cc @rsc


chacha8rand.md
```diff
@@ -0,0 +1,85 @@
+# ChaCha8Rand
+
+ChaCha8Rand is a ChaCha8-based key-erasure CSPRNG with performance similar to non-cryptographic random number generators, designed to be [the default source for the `math/rand{,v2}` and `runtime` Go packages][].
```
dconnolly, line 3: Missing link?

dconnolly, line 3: Does key-erasure need to be defined or a link to a definition added?

rsc:
I think this has converged. Can it be submitted?

chacha8rand.md
```diff
@@ -0,0 +1,245 @@
...
+
+**Why sixteen blocks at a time?** All amd64 processors have 128-bit vector registries, so we can target them without CPU feature detection. Other implementers might find it preferable to pay the complexity and buffer size cost to parallelize eight or sixteen blocks, with 256-bit and 512-bit vector instructions respectively. We produce sixteen blocks before rekeying but interlace four to optimize for a smaller buffer (of four block at a time) but still future-proof (by allowing up to sixteen blocks to be generated at a time with 512-bit vector instructions) and improve throughput.
+
+**Key erasure vs state compression.** Traditional key erasure involves immediately overwriting the key, and deleting buffered output bytes as soon as they are produced. An alternative strategy is deferring the overwriting until the next iteration starts, in which case the state can be serialized to just the last iteration input (from which the current buffer and the next iteration key can be regenerated) and a counter.
```
str4d, line 43: This, combined with the 33 byte length specified above, indicates that this is a 1-byte counter. But nowhere is it specified what this is counting.

My guess is that this is meant to be the ChaCha8 counter (which is guaranteed to fit into 1 byte), rather than a count of how many of the 992 bytes per rekey have been output (which would require a 2-byte counter). But the unspecified implication is that when randomness is sampled that only consumes part of a block, the remainder of the block is discarded. That in turn means that the sample output is only going to match for continuous reads (or reads of a multiple of 992 bytes).

chacha8rand.md
```diff
@@ -0,0 +1,245 @@
...
+
+**Why the subtractions?** ChaCha8 needs to adds the key back to the output, or the block function would be invertible. However, adding back the constants and the counter is unnecessary (as they are public), and is done maybe to allow adding the state as a whole when non-interlaced. (The security importance of adding back the key is mentioned in passing in the [XSalsa20 paper](https://cr.yp.to/snuffle/xsalsa-20110204.pdf), at page 5, when defining HSalsa20.) This is only slowing down SIMD implementations, so it is skipped.
```
str4d, line 45:
```suggestion
**Why the subtractions?** ChaCha8 needs to add the key back to the output, or the block function would be invertible. However, adding back the constants and the counter is unnecessary (as they are public), and is done maybe to allow adding the state as a whole when non-interlaced. (The security importance of adding back the key is mentioned in passing in the [XSalsa20 paper](https://cr.yp.to/snuffle/xsalsa-20110204.pdf), at page 5, when defining HSalsa20.) This is only slowing down SIMD implementations, so it is skipped.
```

str4d:
Generally LGTM. I'll wait until the end of the week (in case any of the unresolved conversations get resolved in this PR), and then I'll move any remaining unresolved conversations to issues and merge the draft spec.

chacha8rand.md
str4d: Opened #46.

chacha8rand.md
str4d: Opened #47.

chacha8rand.md
str4d: Opened #48.

str4d (approved):
Spec is suitable for inclusion. @FiloSottile and @rsc are the assigned maintainers.

The `CODEOWNERS` file contains errors, and I don't know if that will prevent GitHub from enforcing the rest of it correctly; @FiloSottile I'm leaving it in your hands to determine whether there is an issue and fix it if so.

PR #54, opened 2024-03-08
Add signed-note, tlog-checkpoint, and tlog-cosignature specs

FiloSottile:


FiloSottile:
@rgdd added tlog-cosignature!

rgdd (dismissed):
These specs read as very polished now, nice! Just dropped a few minor editorial suggestions -- take what you like, leave what you dislike.

signed-note.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Signatures
+
+    — <key name> base64(32-bit key ID || signature)
+
+The tuple of key name and key ID uniquely identifies the key that performed the
+signature and the algorithm it used. Verifiers MUST ignore signatures from
+unknown keys, even if they share a name or ID (but not both) with a known key.
+If the key name and ID match a known key, verifiers MUST verify the signature
+and MAY treat a verification failure as an error.
```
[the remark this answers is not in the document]
AlCutter: I read it as you've described - hinting that there is likely to be policy around signature verification at play which is outside the scope of this spec.

tlog-checkpoint.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Note text
+
+The note text is a sequence of at least three non-empty lines, separated by
+newlines (U+000A).
+
+ 1. The first line is the **origin**, a unique identifier for the log identity
+    which issued the checkpoint. Like note key names, it MUST be non-empty,
+    well-formed UTF-8 containing neither Unicode spaces nor plus (U+002B), and
+    it SHOULD be a schema-less URL, such as `example.com/log42`. This is only a
```
AlCutter, line 47: This is going make all the existing checkpoint-using logs, including sumdb, non-compliant.

Separately, the language here could do with a massage - the 2nd sentence is a "MUST", but the 3rd sentence says it's just a recommendation. Was the final sentence intended to apply only to the schema-less URL part? Either way it would be good to make this more crisp.

tlog-checkpoint.md
```diff
@@ -0,0 +1,72 @@
...
+    origin is following this format.
+
+ 2. The second line is the **tree size**, the ASCII decimal representation of
+    the number of leaves in the tree, with no leading zeroes (unless the tree is
+    empty, in which case the tree size is `0`).
+
+ 3. The third line is the **root hash**, the base64 encoding of the root of the
+    [RFC 6962] Merkle hash tree at the specified tree size.
+
+ 4. Any following lines are **extension lines**, opaque and optional. Extension
```
AlCutter, line 58: I wonder if we should add some text _somewhere_ about not adding extension lines lightly?

tlog-checkpoint.md
```diff
@@ -0,0 +1,72 @@
...
+    lines, if any, must be non-empty.
+
+## Signatures
+
+Logs MUST NOT sign inconsistent checkpoints, where a consistency proof can't be
+constructed from one size and root hash to the other.
```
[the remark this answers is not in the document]
AlCutter: I think this could be tighter, as-written it would be ok for a log to sign checkpoints along multiple inconsistent views:
```
  0
 / \
A   A'
|   |
B   B'
```

`0->A` and `0->A'` are both pairwise consistent with `0` since everything is consistent with an empty tree, and `A->B` and `A'->B'` are both also pairwise consistent.

It could be more along the lines of `Logs MUST not sign any checkpoint which is inconsistent with any checkpoint it previously signed`?

tlog-cosignature.md
```diff
@@ -0,0 +1,100 @@
+# Transparency Log Cosignatures
+
+A cosignature is a statement by a transparency log [witness][] that it ensured
```
AlCutter, line 3: nit: /ensured/verified ?

A witness has no power to _ensure_ that the checkpoint was consistent before it was issued by the log, only to check after the fact that it is.

tlog-checkpoint.md
mhutchinson: Could be worthwhile, but I like how tight the spec is otherwise. FWIW we wrote https://github.com/transparency-dev/formats/tree/main/log#merging-checkpoints to dissuade people from using this (or at least, educating them on the risks).

One option to _consider_ (not sure that I love this idea myself) is a (potentially empty) non-canonical section at the bottom or beside each spec where we can discuss practical usage such as this, and potentially link to implementations. This would allow the actual spec to stay small, but give readers access to information on how not to shoot themselves in the foot.

tlog-checkpoint.md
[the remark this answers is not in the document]
mhutchinson: I don't see the possibility of misinterpretation in the original text that Al describes, though I do think Al's suggestion is clean and worth considering. Though would we need a definition of consistency/inconsistency? https://www.rfc-editor.org/rfc/rfc6962.html#section-2.1.2 is an option, but the language in this spec is quite different.

My bigger issue with this sentence is that `Logs MUST` introduces the concept of the Log having responsibilities. This is very reasonable, but what/who is a Log? Statements from a Log can be identified by the pair of {origin, signing key}. But as this spec mentions "key rotation" right at the end, we must assume that signing key can't be in this tuple. And origin on its own means that anyone could make checkpoints that could be inconsistent, so we must need some form of identity.

I realise that this tangent has distilled itself down to a key/identity distribution problem. But if the spec relies on there being a concept of a Log with responsibilities, do we need a definition?

In case an example helps make this real:
1. Log performs key rotation and issues checkpoint at size N with both signatures
2. At some later point it issues two checkpoints with the same origin and tree size:
  a. One signed with only the original signatures
  b. Another signed with only the new signature

If these two checkpoints have different root hashes, has the Log invalidated this rule? Or at some point did a new Log get introduced that just happened to have the same prefix as the original Log?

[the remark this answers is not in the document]
mhutchinson: Would it be reasonable to say that unless communicated via some other mechanism, a log is identified by its {origin, signing key}. And then to have an extension to this spec (or another spec) that proposes how key rotation is performed. One way that just came to mind is to use one of the extension lines for the old identity to pass the baton to the new identity. Once that is stated, then any checkpoint with the same origin signed by _either_ of these identities is considered to come from the same Log.

tlog-checkpoint.md
```diff
@@ -0,0 +1,72 @@
...
+## Note text
+
+The note text is a sequence of at least three non-empty lines, separated by
+newlines (U+000A).
+
+ 1. The first line is the **origin**, a unique identifier for the log identity
+    which issued the checkpoint. Like note key names, it MUST be non-empty,
+    well-formed UTF-8 containing neither Unicode spaces nor plus (U+002B), and
+    it SHOULD be a schema-less URL, such as `example.com/log42`. This is only a
+    recommendation to avoid collisions, and clients MUST NOT assume that the
```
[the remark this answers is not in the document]
FiloSottile: I like "assume" a little better when paired with the endpoint thing, otherwise it sounds like clients should check if it's an endpoint and be ok if not.

tlog-checkpoint.md
FiloSottile: The conversation about how much non-normative guidance to include and where is a very interesting one, because we generally have non-normative opinions we want to communicate. It's a bit out of scope for this PR though.

I went a bit hard on this and made extension lines NOT RECOMMENDED. I think we agree on "should not used them unless you really know what you're doing" which is close enough.

tlog-checkpoint.md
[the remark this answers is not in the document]
FiloSottile: Adopted Al's language, and defined consistency in terms of it being possible to construct consistency proofs, avoiding speaking separately about sizes and hashes, to Rasmus's point.

I think how to _authenticate_ a log is out of scope for this spec, and belongs in the log policy of a specific ecosystem. In the abstract sense, a log that signs inconsistent checkpoints is violating this spec. This is why I insisted on using origins to identify logs rather than keys. Whether the log violated the spec or an attacker is trying to fake a checkpoint is down to authentication.

We can definitely also write other specs for key rotation that can be optionally adopted by the ecosystems for whom it makes sense, or to interact with the witness ecosystem.

FiloSottile (commented):
I think I addressed all review comments, PTAL and approve!

tlog-cosignature.md
```diff
@@ -0,0 +1,100 @@
...
+
+Concretely, a cosignature is a [note signature][] applied to a [checkpoint][].
+
+Per the signed note format, a note signature line is
+
+    — <key name> base64(32-bit key ID || signature)
+
+The key name SHOULD be a schema-less URL that identifies the witness. Like the
+checkpoint origin line, this is for disambiguation, and MAY NOT match a publicly
+reachable endpoint (if any).
```
rgdd, line 45: Suggestion: the "(if any)" can probably be removed.

tlog-checkpoint.md
```diff
@@ -0,0 +1,74 @@
+# Transparency Log Checkpoints
+
+A checkpoint is a [signed note][] where the body is precisely formatted for use
+in transparency log applications.  The mandatory note text includes the three
+essential parts of a log's Merkle tree head at a given size.
+
+```
+example.com/behind-the-sofa
```
rgdd, line 8: FWIW: I think this gives unnecessary CT vibes. I'd rather have a more neutral example, e.g., "example.org/some-log" or similar. Like you did with the example witness in cosignature.md. If you update here, please also update the log's example origin line in cosignature.md.

rgdd (approved):
Still looks good to me on a second look! Just dropped two minor editorial suggestions.

signed-note.md
```diff
@@ -0,0 +1,138 @@
...
+they MAY require specific content structures of the note text.
+
+Although multiple signature types are specified, implementations SHOULD select
+only the one(s) required for their design, and avoid supporting multiple ones at
+runtime if possible.
+
+* `0x01` — Ed25519 signatures as specified by this document.
+
+* `0x02` — ECDSA signatures as implemented by github.com/transparency-dev/witness.
+
```
haydentherapper, line 114: Could RSA be supported? While Sigstore defaults to ECDSA, this would be helpful for private deployments of Sigstore infrastructure.

tlog-checkpoint.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Signatures
+
+Logs MUST not sign any checkpoint which is inconsistent with any checkpoint it
+previously signed. Two checkpoints are inconsistent if a consistency proof can't
+be constructed from one to the other.
+
+The log’s key name in its signature line SHOULD match the origin line.
+
+Logs SHOULD use Ed25519 signatures to sign the checkpoint, but MAY use any note
```
haydentherapper, line 70: Why SHOULD for Ed25519, rather than SHOULD for any signed note signature algorithm?

mhutchinson (approved):

tlog-checkpoint.md
FiloSottile: Only a few folks will catch the reference :) and even then, I think amalgamating CT into the checkpoint ecosystem is a good thing overall.

signed-note.md
FiloSottile: We can specify an RSA note signature format (although we'll want to pick specific parameters, like RSA-PSS, hash, key size requirements, etc.) but every new log signature algorithm needs to be supported by every witness, and like everything with cryptography, adding more complexity adds more things that can go wrong.

What makes private deployments unable to use ECDSA? It's supported by every HSM, KMS, and compliance regime I am aware of by now.

tlog-checkpoint.md
FiloSottile: Like above, it's better if the ecosystem doesn't fragment too much on signature algorithms, because then every client and witness and monitor needs to support all of them.

The SHOULD here says that you should have a good reason not to use Ed25519, and if you don't have a strong preference you should use Ed25519, both because it's a decent signature algorithm (small keys, ubiquitous, kinda hard to screw up in this setting) and because having one (whichever it is) helps converging.

tlog-cosignature.md
FiloSottile: Agreed, but since it's the only remaining editorial comment and GitHub annoyingly requires approval even when the author is an owner, punting on it to avoid another cycle of approvals.

See #58 for improving this.

signed-note.md
FiloSottile: (Merging this PR now, so feel free to pull this out into an issue!)

haydentherapper: Opened https://github.com/C2SP/C2SP/issues/63 to discuss more, thanks!

PR #56, opened 2024-03-08
https-bastion: import spec

FiloSottile:
From https://git.glasklar.is/sigsum/project/documentation/-/blob/5d5951ff4cd200b2573b36df9aa6c6c3eb4c3ebb/bastion.md

/cc @rgdd

### DIFF
--- a/https-bastion.md
+++ b/https-bastion.md
@@ -106,12 +106,12 @@ analyzed for domain separation.
 
 ## Appendix B — Example Go adapter
 
-This Go adapter turns a regular HTTP Handler into a bastion backend.
+This Go adapter turns a regular `http.Server` into a bastion backend.
 
 ```go
-func connectAndServe(host string, handler http.Handler, key ed25519.PrivateKey) {
+func connectAndServe(ctx context.Context, host string, srv *http.Server, key ed25519.PrivateKey) {
 	log.Printf("Connecting to bastion...")
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	dialCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 	conn, err := (&tls.Dialer{
 		Config: &tls.Config{
@@ -121,16 +121,17 @@ func connectAndServe(host string, handler http.Handler, key ed25519.PrivateKey) 
 			}},
 			MinVersion: tls.VersionTLS13,
 			MaxVersion: tls.VersionTLS13,
+			NextProtos: []string{"bastion/0"},
 		},
-	}).DialContext(ctx, "tcp", host)
+	}).DialContext(dialCtx, "tcp", host)
 	if err != nil {
-		log.Printf("Failed to connect to bastion: %v", err)
-		return
+		log.Fatalf("Failed to connect to bastion: %v", err)
 	}
-
 	log.Printf("Connected to bastion. Serving connection...")
 	(&http2.Server{}).ServeConn(conn, &http2.ServeConnOpts{
-		Handler: handler,
+		Context:    ctx,
+		BaseConfig: srv,
+		Handler:    srv.Handler,
 	})
 }
 
--- a/https-bastion.md
+++ b/https-bastion.md
@@ -1,0 +1,138 @@
+# HTTPS Bastion
+
+This document specifies an HTTPS bastion, a service that acts as a reverse proxy
+accepting connections from both clients and backends and routing requests from
+the former to the latter.
+
+Bastions allow backends to run in an environment without a publicly reachable
+address, while still being reachable synchronously over a public HTTP endpoint.
+
+The design requires no changes to HTTPS clients, and can be implemented as an
+abstraction or even a separate service on the backend side with no changes to
+most HTTP servers.
+
+Backends are authenticated by their Ed25519 key, and the hash of that public key
+becomes the first segment of the path of client requests. To prevent abuse, it
+is RECOMMENDED that bastions apply an allowlisting or registration policy to the
+backends they will serve.
+
+Like regular HTTPS reverse proxies, the bastion is in the position of blocking,
+observing, or manipulating the backend’s traffic. Bastions SHOULD only be used
+for self-authenticating protocols, that don't rely exclusively on TLS for
+authentication or confidentiality. Alternatively, backends MAY use bastions they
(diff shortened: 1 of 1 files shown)


https-bastion.md
```diff
@@ -0,0 +1,138 @@
...
+### Client to bastion requests
+
+The bastion accepts HTTP requests at
+
+```
+https://<bastion host>/<key hash>/<path>
+```
+
+where "key hash" MUST be a hex-encoded SHA-256 hash of a 32-byte Ed25519
+public key.
```
rgdd, line 62: Consider referencing https://datatracker.ietf.org/doc/html/rfc4648#section-8 for hex, or specify case-insensitive just to be less ambiguous.

https-bastion.md
```diff
@@ -0,0 +1,138 @@
+# HTTPS Bastion
+
+This document specifies an HTTPS bastion, a service that acts as a reverse proxy
+accepting connections from both clients and backends and routing requests from
+the former to the latter.
```
rgdd, line 5: Suggestion: break up into >1 sentence, a bit heavy as is.

rgdd (approved):
I have nothing of substance to add, looks good. Just dropped a few editorial suggestions -- take what you like, leave what you dislike.

https-bastion.md
```diff
@@ -0,0 +1,138 @@
...
+
+At a high level, backends connect to the bastion authenticating with their
+Ed25519 key and start serving HTTP/2 on that connection. The bastion then
+proxies to that connection all requests that include the backend's public key
+hash in the URL path.
+
+### Backend to bastion connection
+
+A backend connects to the bastion’s backend endpoint with TLS 1.3, specifying
+the ALPN protocol `bastion/0`. Other TLS versions MUST be rejected, as only TLS
```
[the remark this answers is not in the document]
FiloSottile: Let's do that while tagging v1.0.0.

https-bastion.md
```diff
@@ -0,0 +1,138 @@
...
+
+If the backend key doesn’t correspond to any open backend connection, the
+bastion MUST serve a 502 Bad Gateway response.
+
+Otherwise, the bastion MUST remove _all_ `X-Forwarded-For` headers from the
+request, add a single `X-Forwarded-For` header with the IP address of the
+client, and proxy the request as `/<path>` over the backend connection. Note
+that the `/<key hash>` portion of the path is trimmed.
+
+The bastion MUST ignore all caching headers is both request and response.
```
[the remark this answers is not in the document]
FiloSottile: The paragraph above uses singular, which feels natural there. (Fixed "in".)

https-bastion.md
FiloSottile: Good catch. Went for requiring lowercase not to create an infinite set of equivalent URLs. Let me know if you think that's an issue.

https-bastion.md
FiloSottile: Toyed with this a bit and failed to make it both lighter and not stuttering.

https-bastion.md
```diff
@@ -0,0 +1,137 @@
...
+that the `/<key hash>` portion of the path is trimmed.
+
+The bastion MUST ignore all caching headers in both request and response.
+
+## Appendix A — Security analysis of reusing a witness key
+
+Transparency log witnesses MAY choose to reuse the witness key as their bastion
+backend key.
+
+If domain separation between the TLS handshake signatures, the X.509 certificate
```
AlCutter, line 78: This section might be more readable if we re-flow the "arc" a bit, e.g.:
 
```
For this to be done safely, there must be domain separation between all protocols the key is used in to mitigate the risk of cross-protocol attacks.

This requirement holds true for the witness key use as described in this document since the various signed message types have no common prefix, guaranteeing domain separation:
* TLS 1.3 ...
* X.509 ...
* The witness protocol ...
```

There's a hanging implication that if you're using the witness key for yet something else, you should make sure that this domain separation holds there too - don't know if we should explicitly call that out or leave it as "obvious" and/or out of scope?
(I don't feel strongly either way, although perhaps being explicit about this here may save someone from a painful mis-step?)

FiloSottile: Much better, thanks!

Added a NOT RECOMMENDED / MUST be analyzed for further reuse.

https-bastion.md
```diff
@@ -0,0 +1,137 @@
...
+
+Appendix B presents an example Go adapter that turns a regular HTTP server into
+a bastion backend.
+
+### Client to bastion requests
+
+The bastion accepts HTTP requests at
+
+```
+https://<bastion host>/<key hash>/<path>
```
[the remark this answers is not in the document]
FiloSottile: I think encouraging a dedicated host is fine, also because bastions need to make host-wide configuration changes to accept backend connections. Hosting a bastion alongside a different site is not technically impossible but is awkward enough that I think we can ignore it.

FiloSottile (commented):
Address review comments, PTAL!

https-bastion.md
```diff
@@ -0,0 +1,138 @@
...
+
+Otherwise, the bastion MUST remove _all_ `X-Forwarded-For` headers from the
+request, add a single `X-Forwarded-For` header with the IP address of the
+client, and proxy the request as `/<path>` over the backend connection. Note
+that the `/<key hash>` portion of the path is trimmed.
+
+The bastion MUST ignore all caching headers is both request and response.
+
+## Appendix A — Example Go adapter
+
```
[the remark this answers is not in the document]
FiloSottile: Done!

https-bastion.md
```diff
@@ -0,0 +1,152 @@
...
+This Go adapter turns a regular HTTP Handler into a bastion backend.
+
+```go
+func connectAndServe(host string, handler http.Handler, key ed25519.PrivateKey) {
+	log.Printf("Connecting to bastion...")
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	conn, err := (&tls.Dialer{
+		Config: &tls.Config{
```
AlCutter, line 117: Does this need a `NextProtos` on here to specify the `bastion/0` protocol?

AlCutter (dismissed):
LGTM, would be great to get a bastion host or two up and running and get the Armored (and any other) Witnesses registering & serving to prove it out!

AlCutter (approved):

Issue #63, opened 2024-03-18
signed-note: Support RSA as a signature type


haydentherapper:
Sigstore will be using the tlog-cosignature and related specifications. Currently, our key hash

Private deployments have the option to use either ed25519, ecdsa or RSA for signature algorithms to maximize cryptographic agility. While the public deployment uses ECDSA, we don't provide any recommendations to private deployments.

The request is to include identifiers for RSA-PKCS#1-v1.5 and RSA-PSS. Within Sigstore deployments, the former is far more prevalent when RSA is used.

For key size and hash requirements, is it sufficient to include suggestions as SHOULDs? Ideally we don't need a signature type per combination of scheme, key size and hash. Tangential, but for ECDSA, should a preferred curve be included as a SHOULD?

Ref: https://github.com/C2SP/C2SP/pull/54#discussion_r1527628144

haydentherapper:
It's just to maximize compatibility for self-hosted instances where there are existing PKIs. We can shift towards PSS as the recommendation when RSA is used, though it would be helpful to have PKCS#1 v1.5 specified still, even with a warning against using it.

FiloSottile:
I am not familiar with self-hosted Sigstore. How does it reuse existing PKIs for the log signing key? (I can imagine how legacy signatures get logged *into* Sigstore, but why reuse an existing key as the signing key for a new log?)

Not trying to be difficult, but saddling the whole witness ecosystem with PKCS#1 v1.5 is something I would like to do only if strictly necessary.

Foxboron:
I'm a bit unclear on how private sigstore instances are used, but is it relevant for *private* sigstore instances to interact with the witness ecosystem at large?

Is there cases where it could be useful for them to do so? I'd sorta imagine they would be publicly available if they where.

AlCutter:
I share the reservations already expressed. 

> It's just to maximize compatibility for self-hosted instances where there are existing PKIs

I'm not aware of any log which relies on a PKI hierarchy for certifying its identity, for what I guess are obvious reasons given the need for CT, but perhaps you've come across rationale for why this would be necessary?

haydentherapper:
What I meant was requirements to use a certain signature type due to one's organization, not reuse of a key. For example, if an organization uses only PKCS RSA keys in GCP or AWS, we want to meet them where they're at ideally. Cloud providers haven't mandated PSS so we still see users using both signature types.

[...]

PR #66, opened 2024-03-20
tlog-witness: new spec imported from Sigsum

FiloSottile:
Previously at https://git.glasklar.is/sigsum/project/documentation/-/blob/main/witness.md

### DIFF
--- a/tlog-cosignature.md
+++ b/tlog-cosignature.md
@@ -41,8 +41,8 @@ Per the signed note format, a note signature line is
     — <key name> base64(32-bit key ID || signature)
 
 The key name SHOULD be a schema-less URL that identifies the witness. Like the
-checkpoint origin line, this is for disambiguation, and MAY NOT match a publicly
-reachable endpoint (if any).
+checkpoint origin line, this is for disambiguation, and MAY match a publicly
+reachable endpoint or not.
 
 The key ID MUST be
 
@@ -63,7 +63,7 @@ The signature MUST be a 72-byte `timestamped_signature` structure.
     }
 
 "timestamp" is the time at which the cosignature was generated, as seconds since
-the UNIX epoch.
+the UNIX epoch (January 1, 1970 00:00 UTC).
 
 "signature" is an [RFC 8032][] signature from the witness public key over the
 message defined in the next section.
@@ -78,9 +78,8 @@ The header line MUST be the fixed string `cosignature/v1`, and provides domain
 separation.
 
 The timestamp line MUST consist of the string `time`, a single space (0x20), and
-the number of seconds since the UNIX epoch (January 1, 1970 00:00 UTC) encoded
-as an ASCII decimal with no leading zeroes. This value MUST match the
-`timestamped_signature.timestamp` value.
+the number of seconds since the UNIX epoch encoded as an ASCII decimal with no
+leading zeroes. This value MUST match the `timestamped_signature.timestamp`.
 
     cosignature/v1
     time 1679315147
@@ -93,7 +92,7 @@ the consistent tree head with the largest size the witness has observed for the
 log identified by the origin line has the specified root hash.
 
 Extension lines MAY be included in the checkpoint by the log, and if present
-MUST be included the cosigned message. However, it's important to understand
+MUST be included in the cosigned message. However, it's important to understand
 that the witness is asserting observation of correct append-only operation of
 the log based on the first three lines of the checkpoint; no semantic statement
 is made about any extension lines, and consensus between witnesses on the
--- a/tlog-cosignature.md
+++ b/tlog-cosignature.md
@@ -65,8 +65,8 @@ The signature MUST be a 72-byte `timestamped_signature` structure.
 "timestamp" is the time at which the cosignature was generated, as seconds since
 the UNIX epoch (January 1, 1970 00:00 UTC).
 
-"signature" is an [RFC 8032][] signature from the witness public key over the
-message defined in the next section.
+"signature" is an Ed25519 ([RFC 8032][]) signature from the witness public key
+over the message defined in the next section.
 
 ## Signed message
 
--- a/tlog-witness.md
+++ b/tlog-witness.md
@@ -1,0 +1,226 @@
+# The Witness Protocol
+
+This document describes a synchronous HTTPS-based protocol to obtain
+[cosignatures][] from transparency log witnesses.
+
+[cosignatures]: https://c2sp.org/tlog-cosignature
+[bastion]: https://c2sp.org/https-bastion
+[checkpoint]: https://c2sp.org/tlog-checkpoint
+[note]: https://c2sp.org/signed-note
+
+## Conventions used in this document
+
+The base64 encoding used throughout is the standard Base 64 encoding specified
+in [RFC 4648][], Section 4.
+
+`U+` followed by four hexadecimal characters denotes a Unicode codepoint, to be
+encoded in UTF-8. `0x` followed by two hexadecimal characters denotes a byte
+value in the 0-255 range.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in [BCP 14][] [RFC 2119][] [RFC
+8174][] when, and only when, they appear in all capitals, as shown here.
+
+[RFC 4648]: https://www.rfc-editor.org/rfc/rfc4648.html
+[BCP 14]: https://www.rfc-editor.org/info/bcp14
+[RFC 2119]: https://www.rfc-editor.org/rfc/rfc2119.html
+[RFC 8174]: https://www.rfc-editor.org/rfc/rfc8174.html
+[RFC 6962]: https://www.rfc-editor.org/rfc/rfc6962.html
+
+## Introduction
+
+This protocol allows clients, usually logs, to obtain cosignatures from
+witnesses, making it possible to produce self-contained inclusion proofs that
+can be verified offline. When producing a new checkpoint, the log reaches out to
+witnesses to request cosignatures over it, providing a consistency proof.
+Witnesses verify that the checkpoint is consistent with the previous state of
+the tree, and return a timestamped cosignature.
+
+A witness is an entity exposing an HTTPS service identified by a name and a
+public key. Each witness is configured with a list of supported log public keys.
+For each log, uniquely identified by its origin line, the witness keeps only
+track of the latest checkpoint it observed and verified.
+
+Clients are not expected to communicate directly with the witnesses, logs and
+(sometimes) monitors are, but there is no authentication of requests beyond the
+validation of the signature on the checkpoint.
+
+Witnesses need to be operated by trusted, reliable entities with minimal churn.
+Log deployments pick a witness trust policy that gets deployed into clients and
+might be difficult or impossible to change. Developing a healthy ecosystem of
+witnesses might be the biggest hurdle to widespread tlog success. It’s
+impractical to build a separate witness ecosystem for each application, so it’s
+important that witnesses be public, interoperable, and scalable. As for
+quantity, there should be more than a few, but once reached a critical mass each
+additional witness has diminishing marginal value. Additional log deployments,
+on the other hand, don’t have diminishing returns and are a desirable goal.
+Because of this, witnesses are designed to scale well with a large number of
+rarely active logs, and to support diverse log designs, including low-latency
+and "serverless" logs.
+
+## HTTPS Interface
+
+A witness is defined by a name, a public key, and by two URL prefixes: the
+*submission prefix* for write APIs and the *monitoring prefix* for read APIs.
+
+A witness MAY use the same value for both the *submission prefix* and the
+*monitoring prefix*.
+
+If exposing the machine holding the witness key material directly to the
+Internet is undesirable, operators MAY use a [bastion][].
+
+### add-checkpoint
+
+The `add-checkpoint` call is used to submit a new checkpoint to the witness,
+along with a consistency proof from a previous checkpoint, and returns the
+cosignature.
+
+    POST <submission prefix>/add-checkpoint
+
+The request MUST be an HTTP POST. Clients SHOULD use and witnesses SHOULD
+support keep-alive HTTP connections to reduce latency and load due to connection
+establishment.
+
+The request body MUST be a sequence of
+  - a previous size line,
+  - zero or more consistency proof lines,
+  - and an empty line,
+  - followed by a [checkpoint][].
+
+Each line MUST terminate in a newline character (U+000A).
+
+The previous size line MUST consist of the string `old`, a single space (0x20),
+and the tree size of the previous checkpoint encoded as an ASCII decimal with no
+leading zeroes (unless the size is zero, in which case the encoding MUST be `0`).
+
+Each consistency proof line MUST encode a single hash in base64. The client MUST
+NOT send more than 63 consistency proof lines.
+
+The submitted checkpoint MAY include multiple signatures. 
+
+    old 20852014
+    PlRNCrwHpqhGrupue0L7gxbjbMiKA9temvuZZDDpkaw=
+    jrJZDmY8Y7SyJE0MWLpLozkIVMSMZcD5kvuKxPC3swk=
+    5+pKlUdi2LeF/BcMHBn+Ku6yhPGNCswZZD1X/6QgPd8=
+    /6WVhPs2CwSsb5rYBH5cjHV/wSmA79abXAwhXw3Kj/0=
+
+    example.com/behind-the-sofa
+    20852163
+    CsUYapGGPo4dkMgIAUqom/Xajj7h2fB2MPA3j2jxq2I=
+
+    — example.com/behind-the-sofa Az3grlgtzPICa5OS8npVmf1Myq/5IZniMp+ZJurmRDeOoRDe4URYN7u5/Zhcyv2q1gGzGku9nTo+zyWE+xeMcTOAYQ8=
+    — example.com/behind-the-sofa opLqBQsREYCgu6xQkYQwJr9fo45a62DN9EdmHXnZdXNqlcVGlCum2Wks+49/V6267UEjw6QUXTS5Rovnzv++qbSzm9Q=
+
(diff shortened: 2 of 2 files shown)


tlog-witness.md
```diff
@@ -0,0 +1,226 @@
+# The Witness Protocol
+
+This document describes a synchronous HTTPS-based protocol to obtain
```
mhutchinson, line 3: This doc refers to HTTPS, which I think would be a recommended deployment strategy. But I think that naked HTTP would still be secure?

tlog-witness.md
```diff
@@ -0,0 +1,226 @@
...
+    — example.com/behind-the-sofa Az3grlgtzPICa5OS8npVmf1Myq/5IZniMp+ZJurmRDeOoRDe4URYN7u5/Zhcyv2q1gGzGku9nTo+zyWE+xeMcTOAYQ8=
+    — example.com/behind-the-sofa opLqBQsREYCgu6xQkYQwJr9fo45a62DN9EdmHXnZdXNqlcVGlCum2Wks+49/V6267UEjw6QUXTS5Rovnzv++qbSzm9Q=
+
+The witness MUST verify the checkpoint signature against the public key(s) it
+trusts for the checkpoint origin, and it MUST ignore signature from unknown
+keys. If the checkpoint origin is unknown, the witness MUST respond with a "404
+Not Found" HTTP status code. If none of the signatures verify against a trusted
+public key, the witness MUST respond with a "403 Forbidden" HTTP status code.
+
+The previous size MUST be equal to or lower than the checkpoint size. Otherwise,
```
AlCutter, line 121: previous -> old?

tlog-witness.md
```diff
@@ -0,0 +1,226 @@
...
+If exposing the machine holding the witness key material directly to the
+Internet is undesirable, operators MAY use a [bastion][].
+
+### add-checkpoint
+
+The `add-checkpoint` call is used to submit a new checkpoint to the witness,
+along with a consistency proof from a previous checkpoint, and returns the
+cosignature.
+
+    POST <submission prefix>/add-checkpoint
```
mhutchinson, line 79: I recommend adding a log ID into the URL. We have done this for our witness: https://github.com/transparency-dev/witness/blob/main/api/http.go#L24

The big advantage of this is that verification of the checkpoint is a lot easier, because the witness can maintain a map of ID -> vkey. Without this, the witness has a potentially huge set of vkeys for all of the logs that it needs to attempt to open every note with. That's a potentially expensive operation that will be hard to protect against DoS apart from old-skool strategies like IP limiting.

If we could also make this ID be the same as https://github.com/transparency-dev/formats/blob/main/log/identifier.go then that would be super awesome.


Of course, the other option is to read the first line of the note without verifying it. But that's an anti-pattern :-)

tlog-witness.md
```diff
@@ -0,0 +1,226 @@
...
+
+If a client doesn't have information on the latest cosigned checkpoint, it MAY
+initially make a request with a previous size of zero to obtain it.
+
+The consistency proof lines MUST encode a Merkle Consistency Proof from the
+previous size to the checkpoint size according to [RFC 6962][], Section 2.1.2. The
+proof MUST be empty only if the previous size is zero or if it matches the size
+in the checkpoint. If the Merkle Consistency Proof doesn't verify, the witness
+MUST respond with a "422 Unprocessable Entity" HTTP status code.
+
```
AlCutter, line 139: I think there should be an explicit line in here about same-size-different-hash checkpoints:

If the old and new checkpoint sizes are identical, the witness MUST check that the root hashes are also identical, and, if they are found not to match, a 409 Conflict response MUST be returned, the witness MUST NOT countersign the new checkpoint.

tlog-witness.md
```diff
@@ -0,0 +1,226 @@
...
+request even if the consistency proof doesn't verify (but MUST NOT cosign the
+checkpoint) as it might be proof of log misbehavior.
+
+If all the checks above pass, the witness MUST update its record of the latest
+cosigned checkpoint and respond with a "200 Success" HTTP status code. The
+response body MUST be a sequence of one or more [note][] signature lines, each
+starting with the `—` character (U+2014) and ending with a newline character
+(U+000A). The signatures MUST be [cosignatures][] from the witness key(s) on the
+checkpoint.
+
```
AlCutter, line 150: "E.g.:"

tlog-witness.md
```diff
@@ -0,0 +1,226 @@
...
+The request MUST be an HTTP POST. Clients SHOULD use and witnesses SHOULD
+support keep-alive HTTP connections to reduce latency and load due to connection
+establishment.
+
+The request body MUST be a sequence of
+  - a previous size line,
+  - zero or more consistency proof lines,
+  - and an empty line,
+  - followed by a [checkpoint][].
+
```
mhutchinson, line 90: Will this consistency proof format be extracted into the proposed `tlog-offline-proof` spec? Would be great to have consistent (haha) consistency and inclusion proof formats.

tlog-witness.md
```diff
@@ -0,0 +1,226 @@
...
+checkpoint must be performed atomically, otherwise the following race can occur:
+
+1. Request A with size N is checked for consistency.
+2. Request B with size N+K is checked for consistency.
+3. The stored size is updated to N+K for request B.
+4. A cosignature for N+K is returned to request B.
+5. The stored size is updated to N for request A, *rolling back K leaves*.
+6. A cosignature for N is returned to request A.
+
+### roster
```
AlCutter, line 171: Just linking to summary of previous discussion we had around rosters: https://github.com/transparency-dev/witness/pull/41#issuecomment-1614965419

tlog-witness.md
```diff
@@ -0,0 +1,226 @@
...
+
+The previous size MUST be equal to or lower than the checkpoint size. Otherwise,
+the witness MUST respond with a "400 Bad Request" HTTP status code.
+
+The witness MUST check that the previous size matches the size of the latest
+checkpoint it cosigned for the checkpoint's origin (or zero if it never cosigned
+a checkpoint for that origin). If it doesn't match, the witness MUST respond
+with a "409 Conflict" HTTP status code. The response body MUST consist of the
+tree size of the latest cosigned checkpoint in decimal, followed by a newline
+(U+000A). The response MUST have a `Content-Type` of `text/x.tlog.size`.
```
mhutchinson, line 129: The approach we took here is for the witness to simply return the old checkpoint back to the caller. Any reason to prefer this approach with custom headers?

Reading between the lines it seems like you're intentionally trying to avoid having the witness serve checkpoints up so that it doesn't become load-bearing on the read path?

tlog-witness.md
mhutchinson: +1 and also consider upgrading the requirement that the witness MUST log such bad requests at least once per log. Capping this to once per absolute proof of misbehaviour maintains evidence while keeping the amount of space required to be `2 * #Logs * CheckpointSize`

tlog-witness.md
```diff
@@ -0,0 +1,226 @@
+# The Witness Protocol
```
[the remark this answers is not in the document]
mhutchinson: +1, or `tlog-sync-witness` if fewer characters is the goal.

tlog-witness.md
rgdd: Or "Example:".  Please also make it consistent with other listed examples if you change this here. You could also consider floating examples to the bottom of each subsection -- not sure if it would be better or worse.

tlog-witness.md
```diff
@@ -0,0 +1,226 @@
...
+The witness MUST persist the new checkpoint before responding. Note that
+checking the previous size against the latest checkpoint and persisting the new
+checkpoint must be performed atomically, otherwise the following race can occur:
+
+1. Request A with size N is checked for consistency.
+2. Request B with size N+K is checked for consistency.
+3. The stored size is updated to N+K for request B.
+4. A cosignature for N+K is returned to request B.
+5. The stored size is updated to N for request A, *rolling back K leaves*.
+6. A cosignature for N is returned to request A.
```
rgdd, line 169: This explanation is great, but borderline appendix material.  I.e., "Must be atomic to avoid race, see appendix" would maybe be neater.

rgdd (commented):
Suggestion: split this PR up in two, a new one with the editorial cosignature changes that I think all of us are fine with, and then keep this PR for the witness protocol.

A question to @FiloSottile: am I correct to assume you want a draft of this merged before RWC? I'm asking so we can tune our expectations of what can be done until then, and as a reminder that we are not in a rush to solve everything at once.

At risk of sounding like a broken record, I don't think neither behind-the-sofa nor sigsum are good examples in a spec. I'd rather see consistent use of example.org/footree etc. Won't comment on this very minor detail again, so take it or leave it!

Also, a very high-level comment is that I would be fine to merge mostly as is. From my perspective this needs a bit more editorial, and we need to ensure the roster gets finalized. The former doesn't have to be perfect yet, and the latter should be out-of-scope for this PR.

rgdd:
(Hmm, the review comments did not land on the threads I tried to comment on. Doh.)

rgdd:
> (Hmm, the review comments did not land on the threads I tried to comment on. Doh.)

OK never mind, refresh of the page floated things into the right places.

tlog-witness.md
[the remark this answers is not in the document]
rgdd: (Should also add: question is a bit off topic -- just curious where you're at currently. Might be easier and more appropriate to sync outside of the PR!)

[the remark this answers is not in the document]
mhutchinson: As Al said, we already have a very similar API to this inside the AW: https://github.com/transparency-dev/witness/blob/main/api/http.go. This is not exposed by design, and is instead wrapped in AW to make it an asynchronous client-mode witness, instead of the synchronous server-mode witness that this spec describes.

We don't foresee a future where the AW will be a directly addressable server (though bastion is an option down the line), and thus _this spec_ won't describe the AW as anyone interacts with it. Hence our preference that this clearly says `sync` in the name. It would be an own-goal for transparency if this spec and the AW witness network comes out, and the AW doesn't implement the `tlog-witness` spec, and there is no network of witnesses implementing said spec.

tlog-witness.md
mhutchinson: Yep, and and on-path attacker could do the same with HTTPS too.

tlog-witness.md
mhutchinson: Yeah I'm convinced. I wasn't really happy with the proposal to require logging anything and the entailments of that, but the idea of dropping damning evidence into /dev/null also feels sad. But this should be an implementation choice, not a spec choice.

tlog-witness.md
```diff
@@ -0,0 +1,226 @@
...
+origin.
+
+    GET <monitoring prefix>/roster
+
+The request MUST be an HTTP GET.
+
+Note that this endpoint MAY be served from a different prefix from that of the
+`add-checkpoint` endpoint, to reflect the fact that this endpoint serves a
+different set of clients with potentially different request patterns and rate
+limits, and to facilitate delegating it to a static hosting service.
```
mhutchinson, line 183: It also sounds like it should be a different spec too :-)

tlog-witness.md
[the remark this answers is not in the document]
FiloSottile: Sync vs async is a broader conversation that I think we should have in person (one could even say, synchronously :P). As you know, I am very skeptical that async witnesses can serve a diverse tlog ecosystem. (As a recent example, it would be much more complicated for APT repository signing to have to pause, expose tiles on HTTP somehow, poll, and then resume signing the repository. Compare with https://github.com/FiloSottile/litetlog/tree/main/cmd/apt-transport-tlog which wouldn't have to change basically to integrate synchronous witnesses.)

For the name, I think it only makes sense to have "sync" in it if async witnesses need a spec. My understanding is that AW would reach out to logs (for which we need a tiles spec) and then publish checkpoints (for which we already have checkpoint and cosignature specs). I guess we could document the endpoint at which the checkpoint is served, but that feels light for a spec, and more appropriate for an API documentation.

tlog-witness.md
haydentherapper: +1 to returning the last known checkpoint. This is helpful to avoid more roundtrips, especially for logs with significant amounts of traffic where checkpoints may be frequently updated. If a log requests a cosignature for a checkpoint that's older than what it's already verified, the cosigned checkpoint can be returned to the log and the log can verify consistency from its older checkpoint to the latest cosigned checkpoint.

tlog-witness.md
```diff
@@ -0,0 +1,226 @@
...
+response body MUST be a sequence of one or more [note][] signature lines, each
+starting with the `—` character (U+2014) and ending with a newline character
+(U+000A). The signatures MUST be [cosignatures][] from the witness key(s) on the
+checkpoint.
+
+    — witness.example/w1 CMp+6LWBU0anHGH5aNDTJkH/gj79sG+T6+iP2ThYN5krrDJbR1HDnucjL39QsZTSvVjyQLrdk3DXDqI5G2HgLatVs0pWh6Up69HVOw==
+    — witness.example/w1 I7rEps0pvK2UqkS2gSpVUDhrhVtQV9lgF6pRrWAvjJHjyWpW7VcE3SiOlVlbQNt64vWhO+DlkL0+UfzuOBMh9ChdMkP1vi/lCAsmlw==
+    — witness.example/w2 AWui8Sk55XjYLOijihBjhqEH6nS1ndDymE0a+6idX7pLcnoB+dhnz0854aLZgrrKbYKA7nC3HNJhm/kWl7oJlqU3rXXvpysAdyP3wQ==
+
+The client MUST ignore any cosignatures from unknown keys. To parse the
```
haydentherapper, line 155: This may be out of scope for the spec, but how do we expect clients to discover witness keys?

tlog-witness.md
FiloSottile: The custom headers are just a way to distinguish the witness protocol error from, say, a CDN error. I think we should have them whatever the body is.

Indeed, we want to make sure this API doesn't get abused to read checkpoints from witnesses.

How can a log request a cosignature for an older checkpoint, when the log produces the checkpoints?

Anyway, the witness certainly can't return a consistency proof, so the client here needs to have access to the tree to check any kind of consistency, so I don't think returning the whole checkpoint adds anything over returning just the size.

haydentherapper: > How can a log request a cosignature for an older checkpoint, when the log produces the checkpoints?

If a log is producing checkpoints per-entry rather than batching.

* Two requests to the log for entry inclusion
* The log generates a checkpoint for the first entry (size N)
* The log generates a checkpoint (size N+1) for the second entry
* The log sends two requests to the witness for N and N+1
* Checkpoint N+1 succeeds first, and the cosigned checkpoint is returned to the log
* Checkpoint N fails to be cosigned since N < N+1

> Anyway, the witness certainly can't return a consistency proof, so the client here needs to have access to the tree to check any kind of consistency, so I don't think returning the whole checkpoint adds anything over returning just the size.

Is the expectation that the log persists the cosigned checkpoint? Otherwise, if you return only the size, the log can compute the relevant checkpoint but needs to request another cosignature. If so, can we document that?

FiloSottile: > If a log is producing checkpoints per-entry rather than batching.

Interesting, didn't think of that!

I'd expect such a log to keep around the latest cosigned checkpoint somewhere that's accessible to the first request (since that request needs to access state from "the future" anyway to build the proofs against the latest checkpoint).

This log design gets unmanageable fast with multiple witnesses though. Even if we were to return a full cosigned checkpoint on a 409, the checkpoints returned by different witnesses could get out of sync, and then the cosignatures can't be merged.

tlog-cosignature.md
```diff
@@ -63,7 +63,7 @@ The signature MUST be a 72-byte `timestamped_signature` structure.
     }
 
 "timestamp" is the time at which the cosignature was generated, as seconds since
```
OR13, line 65: according to who? you might consider a more secure form of timestamp here.

tlog-cosignature.md
```diff
@@ -63,7 +63,7 @@ The signature MUST be a 72-byte `timestamped_signature` structure.
...
-the UNIX epoch.
+the UNIX epoch (January 1, 1970 00:00 UTC).
 
 "signature" is an [RFC 8032][] signature from the witness public key over the
```
OR13, line 68: EdDSA with Ed25519 or Ed448 ? since you are building a JSON format, you should consider JOSE and algorithmic agility, especially with PQ signatures coming soon.

See the code signing section of: 
https://bughunters.google.com/blog/5108747984306176/google-s-threat-model-for-post-quantum-cryptography

tlog-witness.md
OR13: its nice to be able to filter requests based on the URL structure, in case a single server hosts multiple distinct logs.

tlog-cosignature.md
FiloSottile: According to the witness. You can actually use the witnesses to obtain a form of secure timestamp.

tlog-cosignature.md
FiloSottile: Oh right we need to specify Ed25519, thank you.

I don't think we use JSON anywhere?

But also, JOSE-style algorithmic agility is a non-goal. We are targeting a different set of design parameters, trying to make simple rather than agile. I think we diverge from the IETF style there, and it's worth keeping in mind when reviewing these specs. See https://github.com/C2SP/C2SP/issues/63 for example.

tlog-witness.md
FiloSottile: It will depend on who sets policy. By example, the Go sumdb client policy is set by the Go team, who sets the defaults of the go tool in the open source project. I expect in most ecosystems the log operators will also set client policy (in the open, and in such a way that the relevant community trusts the resulting transparency system).

tlog-witness.md
FiloSottile: I tried but it came out awkward, because there are no other appendixes, and because it ended up being just the numbered list.

tlog-witness.md
FiloSottile: Removed the roster for now.

tlog-witness.md
FiloSottile: I think to summarize returning only the size discourages using the write path to obtain checkpoints, and the scenarios where it would be a problem are already problematic for other reasons.

> While this might be more in scope for a spec on offline signatures, do you think the spec should have any suggestions about atomic operations for the log, not just the witness?

Yeah I think that will be down to implementation preference, but I think that the conclusion that there needs to either be leader election or single-node architecture is unavoidable.

AlCutter (approved):

rgdd (approved):

PR #73, opened 2024-05-24
tlog-tiles: new spec for tiled tlogs

FiloSottile:


FiloSottile:
Opened this for early discussion. Will also make a new-spec issue to get spec name assignment.

tlog-tiles.md
```diff
@@ -0,0 +1,157 @@
...
+`<N>` and `.p/<W>` have the same meaning as in Merkle Tree tile paths above.
+
+Data tiles SHOULD be compressed at the HTTP layer. Logs MAY use
+`Content-Encoding: gzip` with no negotiation, or any compression algorithm
+requested by the client with `Accept-Encoding`. Clients SHOULD include `gzip`
+and `identity` in their `Accept-Encoding` headers.
+
+This endpoint is immutable, so its caching headers SHOULD be long-lived.
+
+Data tiles are sequences of big-endian uint16 length-prefixed entries. Each
```
jiggoha, line 133: Since this is the first use of the phrase "data tiles", could we specify what it means? Maybe just explicitly say something like "A data tile contains the full entry bundle at the path defined above."

tlog-tiles.md
```diff
@@ -0,0 +1,157 @@
...
+entry in a data tile hashes to the corresponding entry in the corresponding
+“level 0” tile according to RFC 6962, Section 2.1.
```
jiggoha, line 135: I'm having a hard time finding "level 0" or "tile" mentioned in RFC 6962. Which part of Section 2.1 do you have in mind?

```The hash of a list with one entry (also known as a leaf hash) is: ... ```

Or perhaps
```The input to the Merkle Tree Hash is a list of data entries; these entries will be hashed to form the leaves of the Merkle Hash Tree.```
?

Comment (mhutchinson) [reply tlog-tiles.md:135]: This sentence needs reworking because I can completely see your confusion. What it is trying to say is that the hashing method is the same as RFC 6962, i.e. the value that is hashed has a domain separation prefix applied to it (https://github.com/transparency-dev/merkle/blob/main/rfc6962/rfc6962.go#L24)

File: tlog-tiles.md [hunk 3]
```diff
@@ -0,0 +1,157 @@
...
+
+This endpoint is mutable, so its headers SHOULD prevent caching beyond a few
+seconds.
+
+The checkpoint MUST carry at least one Ed25519 signature by the log, and MAY
+carry additional signatures of other types.
+
+### Merkle Tree
+
+Instead of serving arbitrary consistency and inclusion proofs, logs serve the
```
smeiklej, line 54: I'm not sure what makes consistency and inclusion proofs "arbitrary" - what could be more specific to a given entry than the exact path from it to the root (or rather the neighbors on this path)?  Presumably what is meant here is that these proofs are not reusable in the way that tiles are, so maybe this is a good place to state the value of using tiled logs.

FiloSottile: I meant the client can request proof for arbitrary entries/sizes. Updated.

tlog-tiles.md
FiloSottile: We renamed data tiles to entry bundles and forgot to update the names here. Fixed.

tlog-tiles.md
FiloSottile: Indeed, split up the sentences to clarify.

FiloSottile (commented):
I also added a paragraph about logs that serve entries in a different format, such as the sumdb (which can also serve entry bundles in parallel) and Sunlight (which would lose a lot of compression by not putting the Precert next to the Merkle leaf, so can't).

tlog-tiles.md
```diff
@@ -0,0 +1,163 @@
...
+level 1 tile of width 17, and one partial level 2 tile of width 1. Note that a
+tree of size 256 will be represented by a full level 0 tile and a partial level
+1 tile of width 1.
+
+Logs MUST serve partial tiles corresponding to tree sizes for which a checkpoint
+was produced, but MAY delete any partial tile once the corresponding full tile
+becomes available. Clients MUST NOT fetch arbitrary partial tiles without
+verifying a checkpoint with a size that requires their existence, and MAY fetch
+the full tile in parallel as a fallback in case the partial tile is not
+available anymore.
```
AGWA, line 114: It would be good to specify an HTTP status code for deleted / non-existent tiles so that clients can distinguish between a successful request, a deleted tile, and an error that might be fixed by retrying.

For example: "Logs MUST return an HTTP status code between 400 and 499 inclusive if a tile does not exist."

FiloSottile: I want to make sure we are compatible with most file storage systems such as S3, but I guess those will always do 200 for success and 404 or 410 for not found?

FiloSottile: Ok, looking at https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList it seems that some 4xx might be retryable but not 404.

AGWA: I wanted to specify 404 or 410 but Let's Encrypt Twig (which appears to be S3) is returning 403 for non-existent files so I made it 4xx to be safe.

The fact that some 4XX codes are retryable is annoying. 429 (Too Many Requests) is generally retryable as well (though S3 uses 503 for that case!).

tlog-tiles.md
```diff
@@ -0,0 +1,163 @@
...
+
+[RFC 8446]: https://www.rfc-editor.org/rfc/rfc8446.html
+[BCP 14]: https://www.rfc-editor.org/info/bcp14
+[RFC 2119]: https://www.rfc-editor.org/rfc/rfc2119.html
+[RFC 8174]: https://www.rfc-editor.org/rfc/rfc8174.html
+
+## Parameters
+
+A tiled transparency log is defined by a URL *prefix*, a [checkpoint][] origin,
+and one or more Ed25519 [signed note][] public keys.
```
haydentherapper, line 21: I've raised the issue around cryptographic agility before, so this comment might sound familiar, though I've come with concrete examples for the Sigstore ecosystem for why I'd like to see more than just Ed25519.

As a first question, Is it accurate to say that the spec as-is is designed around publicly available logs that will be witnessed by a witness network? Given that the cosignature spec mandates Ed25519, is this the primary motivation for requiring Ed25519 here?

As a compromise, could we say that logs that are to be witnessed by a public witness network MUST use Ed25519, but otherwise logs MAY use any signature algorithm?

If the spec mandates cryptographic agility, it will mean that the log APIs (e.g. configuration) will be designed so that you can provide your own algorithm. If not, then I suspect that Ed25519 will be hardcoded throughout the codebase and make it harder to use alternative algorithms without making significant code changes and maintaining forks. If we go with the suggestion to mandate Ed25519 for public logs but allow for any algorithm for non-public logs, the APIs should be designed around pluggable algorithms and make it much simpler to use other algorithms.

For the Sigstore ecosystem:

As I've mentioned in other issues, the goal is meeting customers where they're at. Last week I met with someone from a package registry who noted that their infrastructure only supports RSA. I'd like to try to nudge them to support more algorithms, but if this is a hard restriction, I'd rather support RSA and have this ecosystem adopt Sigstore. Ed25519 is not nearly as prolific as ECDSA so if I need to convince an ecosystem to make changes to supported algorithms, I believe it'll be easier to convince them to adopt ECDSA (In the previous issues, it was noted there have been vulnerabilities in RSA as a reason to not adopt RSA, but ECDSA has not had the same issues).

Sigstore clients have implemented ECDSA support primarily. Getting all clients onboard with Ed25519 will be another hurdle for adoption of modern logs and further delay our tile adoption. Not infeasible of course, just adds another blocker.

tlog-tiles.md
FiloSottile: It sounds like the details of what is retryable is unfortunately backend-specific, so we can't help clients much from the spec. If that's not right, please open an issue and we can add more text to the draft after merging.

tlog-tiles.md
FiloSottile: Added a carveout for non-public logs. The goal is indeed guaranteeing that witnesses don't need to implement anything else than Ed25519.

tlog-tiles.md
```diff
@@ -46,8 +46,9 @@ with `Content-Type: text/plain; charset=utf-8`.
 This endpoint is mutable, so its headers SHOULD prevent caching beyond a few
 seconds.
 
-The checkpoint MUST carry at least one Ed25519 signature by the log, and MAY
-carry additional signatures of other types.
+If the log is public or anyway interacting with the public witness network, the
```
AlCutter, line 49: nit: "If the log is public, or is interacting in any way with the public witness network, ..."

AlCutter (approved):

PR #80, opened 2024-06-20
age-plugin: allow multiple stanzas

FiloSottile:
A plugin might be invoked with a single recipient encoding, but return multiple stanzas, e.g. to support multiple formats, to build group aliases, or to act as a proxy.

We tried and failed to figure out why this was added, but it was probably before we properly abstracted file headers stanzas from recipients.


Foxboron (approved):

dconnolly (approved):

str4d:
Post-hoc ACK.

PR #85, opened 2024-06-28
sunlight: rename to static-ct-api

### LINKED ISSUE #82: sunlight: rename to static-ct-api
The use of the Sunlight name for both [the CT log implementation](https://sunlight.dev) and for the API specified in C2SP is causing confusion. There is consensus to rename the spec a boring "Static Certificate Transparency API" or `static-ct-api` for short.

https://groups.google.com/a/chromium.org/g/ct-policy/c/v9JzlbphYBs/m/LwMuqT4YAwAJ

We don't have a formal process for this, but I think applying the new spec name assignment process would be fine.

FiloSottile:
Closes #82

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -27,8 +27,8 @@ signed-note/            @FiloSottile
 streaming-encryption.md @FiloSottile @str4d
 streaming-encryption/   @FiloSottile @str4d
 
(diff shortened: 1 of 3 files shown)


str4d (approved):
ACK da3efa836a86487122ea60713375c5ee2e370701

PR #92, opened 2024-08-14
Replace application/data with application/octet-stream

AlCutter:
This PR changes references to `Content-Type: application/data` to `Content-Type: application/octet-stream` in the `static-ct-api` and `tlog-tiles` specs.

`application/octet-stream` is an [IANA registered type](https://www.iana.org/assignments/media-types/application/octet-stream), but `application/data` is not.

### DIFF
--- a/static-ct-api.md
+++ b/static-ct-api.md
@@ -196,7 +196,7 @@ Tiles are served as at
 
     <monitoring prefix>/tile/<L>/<N>[.p/<W>]
 
-with `Content-Type: application/data`.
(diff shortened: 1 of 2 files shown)


mhutchinson (approved):

phbnf:
Ah you beat me to it I was about to speak with Martin and raise and issue after lunch if need be. My plan was to submit this to align with specs, and to change this later if the specs change.

FiloSottile:
Good catch, my bad!

FiloSottile (approved):

PR #96, opened 2024-08-22
age: Add specification for the p256tag recipient type & stanza.

remko:
As discussed in #31


age.md
```diff
@@ -273,6 +275,79 @@ other types. This is to uphold an expectation of authentication that is
...
+
+The identity implementation MUST ignore any stanza with a tag for which
+it does not have any corresponding secret.
+
+The identity implementation computes the shared secret as follows:
+
+    shared secret = P256(identity, ephemeral share)
+
+If the shared secret is the point at infinity, the identity implementation
+MUST abort
```
Foxboron, line 345:
```suggestion
MUST abort.
```

age.md
```diff
@@ -273,6 +275,79 @@ other types. This is to uphold an expectation of authentication that is
...
+The p256tag recipient is computed as 
+
+    recipient = P256(identity, basepoint)
+
+P256 is the scalar multiplication over P-256, with SEC 1 point decoding and 
+encoding.
+
+The recipient is encoded as Bech32 with HRP `age1p256tag`.
+
+    age1p256tag1qf0l9gks6x65ha077wq3w3u8fy02tpg3cd9w5j0jlgpfgqkcut2lwpmkcwq
```
str4d, line 292: The plugin protocol requires that this recipient be handled by the `age-plugin-p256tag` plugin and not by the native app, which I don't think is the intention. If we want to define a new recipient encoding that all age implementations are required to understand, it likely needs to be separate from the plugin namespace.

age.md
```diff
@@ -166,8 +166,10 @@ fresh nonce.
 
 ## Native recipient types
 
-This document specifies two core age recipient types: an asymmetric encryption
-type based on X25519, and a passphrase encryption type based on scrypt.
+This document specifies three core age recipient types: an asymmetric encryption
+type based on X25519, a passphrase encryption type based on scrypt, and a
+tagged assymetric encryption type based on NIST P-256 for use with keys stored
```
davidloiret, line 171: typo, "assymetric" => "asymmetric"

Foxboron:
I don't see Filippo tagged here for review, so friendly ping @FiloSottile

age.md
```diff
@@ -273,6 +275,79 @@ other types. This is to uphold an expectation of authentication that is
...
+
+Age does not generate p256tag identities, but supports encrypting to
+p256tag recipients.
+
+The p256tag recipient is computed as 
+
+    recipient = P256(identity, basepoint)
+
+P256 is the scalar multiplication over P-256, with SEC 1 point decoding and 
+encoding.
```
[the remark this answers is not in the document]
Disasm: Also it might be a good idea to specify if point compression should be used or not.

PR #104, opened 2024-09-11
BLAKE3: v1.0.0

oconnor663:
Most of this text is adapted from https://github.com/BLAKE3-team/BLAKE3-IETF. I will squash when merging.

### DIFF
--- a/BLAKE3.md
+++ b/BLAKE3.md
@@ -67,7 +67,7 @@ algorithms. These primitives include:
   can replace HKDF. The context string in this mode should be
   hardcoded, globally unique, and application-specific.
 
-By default, all modes returns a 32-byte output, but they can produce an
+By default, all modes return a 32-byte output, but they can produce an
 output of arbitrary size.
 Therefore, the hash mode can be used as an extendable-output-function
 (XOF), and the keyed hash mode can be used as a deterministic random bit
--- a/BLAKE3.md
+++ b/BLAKE3.md
@@ -444,10 +444,10 @@ Section 5 of the [BLAKE3 paper][paper] This section providers a brief
 overview of these, as a starting point to implementers, covering the
 most salient points.
 
-Reference implementations of BLAKE3 in the C and Rust languages are
-available in the [BLAKE3 repository][repo]. These implementations
-include parallel implementations leveraging multi-threading and
-different SIMD processing technologies.
+Optimized implementations of BLAKE3 in the C and Rust languages are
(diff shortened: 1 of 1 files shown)


veorq (approved):

BLAKE3.md
```diff
@@ -0,0 +1,921 @@
...
+  instances.
+
+* **Key derivation (`derive_key`)**:  The key derivation mode takes two
+  input values, each of arbitrary size: a context string, and key
+  material.  BLAKE3 in this mode can be used whenever a key derivation
+  function (KDF) is needed.  For example, BLAKE3 in key derivation mode
+  can replace HKDF. The context string in this mode should be
+  hardcoded, globally unique, and application-specific.
+
+By default, all modes returns a 32-byte output, but they can produce an
```
sneves, line 70:
```suggestion
By default, all modes return a 32-byte output, but they can produce an
```

oconnor663:
@FiloSottile I think I need your help to tag this as `BLAKE3/v1.0.0` now that it's landed on `main`?

FiloSottile:
> @FiloSottile I think I need your help to tag this as `BLAKE3/v1.0.0` now that it's landed on `main`?

I can do that! Just to confirm, once you tag v1.0.0 there is an expectation of not breaking backwards compatibility. I assume that's not a problem for the scheme itself since it's already widely deployed, but wanted to make sure you are ready to commit to the text as written.

oconnor663:
Apart from future edits for typos/clarity/brevity, I'm confident in the text as written. Like you said, most of it can't change without breaking test vectors, which we'll never do. (If some cryptographic bombshell was published and we had to make fixes, we'd probably give it a new name.) I think the only squishy/debatable detail is the requirement that the `derive_key` context string has to be "hardcoded, globally unique, and application-specific". That [does get pushback](https://github.com/BLAKE3-team/BLAKE3/pull/424) from time to time, but it's been documented in all of our implementations for years, and so far I think it's stood the test of time.

@veorq, @sneves, @zooko, thoughts?

zooko:
Yeah, I haven’t noticed any feedback from deployments over the last few
years that suggest we ought to change something, except for that one that
Jack mentioned, where his answers to the pushback seem satisfying to me.
Regards, Zooko

oconnor663:
@FiloSottile let's tag it.

FiloSottile:
Done! https://c2sp.org/BLAKE3@v1.0.0

oconnor663:
Sick, thanks Filippo!

PR #111, opened 2024-11-12
Cap max tile level at 6

phbnf:
RFC6962 specifies that a `TreeHeadSignature` includes an `uint64` `tree_size`.  That means that there cannot be more than `2^64` leaves, i.e `2^56` entries bundles, i.e a maximum of 7 tile levels.

### DIFF
--- a/static-ct-api.md
+++ b/static-ct-api.md
@@ -199,7 +199,7 @@ Tiles are served as at
 with `Content-Type: application/octet-stream`.
 
 `<L>` is the “level” of the tile, and MUST be a decimal ASCII integer between 0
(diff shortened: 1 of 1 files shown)


FiloSottile:
The leaf index in the SCT extension is a uint40 (enough for > 34k certs/s for a year), so I think the bounds for the levels are 0 to 5, although level 5 can only ever be a single partial tile of width 1, when the tree has exactly 2^40 leaves, right?

FiloSottile:
LGTM, this is arguably a bugfix since beyond that level the SCT extension would overflow.

FiloSottile (approved):

PR #113, opened 2025-01-21
det-keygen: new spec for deterministic ECDSA key generation

FiloSottile:


leontiad:
@FiloSottile  Would be nice to have a spec also tailored for CTR_DRGB according to [NIST.SP.800-90Ar1](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-90Ar1.pdf), Section 10.2

FiloSottile:
> @FiloSottile Would be nice to have a spec also tailored for CTR_DRGB according to [NIST.SP.800-90Ar1](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-90Ar1.pdf), Section 10.2

You can use CTR_DRGB to generate the seed, but I think it's best to have only a single algorithm for the actual key generation. HMAC is already necessary for RFC 6979 (aka NIST Deterministic ECDSA) and draft-irtf-cfrg-det-sigs-with-noise.

PR #118, opened 2025-03-20
.github/maintainers: add simple Web UI for team membership

### LINKED ISSUE #109: Apparent lack of transparency of steward and maintainer changes
/cc @C2SP/stewards

This repository used to have a CODEOWNERS file that contained the usernames of the C2SP stewards as well as the maintainers of each specification. The file was deleted by [this commit](https://github.com/C2SP/C2SP/commit/908e8b949f47698c1b9651e571c7355ef67b4e06), and replaced by "policy-bot". In [a later commit](https://github.com/C2SP/C2SP/commit/fad3f68f8b872fb5864cd2dc32d4c00c914ca08a#diff-438478e7713ffdb96be895cd2cba03144f30c04f1442ae246acdeb2f1c13efd5) README.md was updated to include links to team pages under the C2SP GitHub organisation. The team pages are not publicly accessible.

The benefit of storing stewards and maintainers in the same Git repo as the specifications themselves is that changes can be easily and cryptographically tracked over time. A non-public list of stewards and specification maintainers, where changes can't be easily tracked, is the opposite of transparent governance.

I suspect the changes were made to simplify administration, and that the governance consequences were not considered. If so, please confirm and roll back the deletion of CODEOWNERS.

I would also like to know what changes of stewards and specification maintainers have occurred since the commit above.

Thank you for starting and stewarding the C2SP project.

FiloSottile:
Fixes #109

### DIFF
--- a/.github/maintainers/maintainers.go
+++ b/.github/maintainers/maintainers.go
@@ -1,0 +1,164 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+	"slices"
+	"strings"
+	"text/template"
+
+	"github.com/bradleyfalzon/ghinstallation/v2"
+	"github.com/google/go-github/v69/github"
+)
+
+type TeamMember struct {
+	Login     string `json:"login"`
+	AvatarURL string `json:"avatar_url"`
+	HTMLURL   string `json:"html_url"`
+}
+
+const c2spGitHubAppID = 1185968
+const c2spGitHubInstallationID = 63003777
+
+func main() {
+	tr, err := ghinstallation.New(
+		http.DefaultTransport,
+		c2spGitHubAppID, c2spGitHubInstallationID,
+		[]byte(os.Getenv("GITHUB_APP_PRIVATE_KEY")),
+	)
+	if err != nil {
+		log.Fatalf("Failed to create GitHub App transport: %v", err)
+	}
+
+	githubClient := github.NewClient(&http.Client{Transport: tr})
+
+	http.HandleFunc("GET /team/{slug}", func(w http.ResponseWriter, r *http.Request) {
+		teamSlug := r.PathValue("slug")
+		if teamSlug == "" {
+			http.Error(w, "Missing team slug", http.StatusBadRequest)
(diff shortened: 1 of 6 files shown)


Shahadat7417 (approved):

.github/maintainers/maintainers.go
```diff
@@ -0,0 +1,155 @@
...
+type TeamMember struct {
+	Login     string `json:"login"`
+	AvatarURL string `json:"avatar_url"`
+	HTMLURL   string `json:"html_url"`
+}
+
+func main() {
+	tr, err := ghinstallation.New(
+		http.DefaultTransport,
+		1185968, 63003777,
```
str4d, line 23: I prefer to avoid undocumented magic numbers. Either add a comment explaining what these are for, or define constants that name them.

.github/maintainers/maintainers.go
```diff
@@ -0,0 +1,155 @@
...
+	githubClient := github.NewClient(&http.Client{Transport: tr})
+
+	http.HandleFunc("GET /team/{slug}", func(w http.ResponseWriter, r *http.Request) {
+		teamSlug := r.PathValue("slug")
+		if teamSlug == "" {
+			http.Error(w, "Missing team slug", http.StatusBadRequest)
+			return
+		}
+
+		members, _, err := githubClient.Teams.ListTeamMembersBySlug(
```
str4d, line 39: I don't know what order GitHub is using for this, but I spot-checked and it seems that the order returned here is the same order used for https://github.com/orgs/C2SP/teams/maintainers?query=membership%3Achild-team

I'm not a huge fan of that order as it has no visible reason (I'm guessing it's maybe ordered by internal GitHub ID), and would prefer alphabetical (either by name or username).

str4d (approved):
ACK d7c57f6975e4784682c59da62b944abdac97b316 with comments. This will do for now.

.github/maintainers/maintainers.go
FiloSottile: Sorted alphabetically by username (since that's what we show at the moment).

PR #119, opened 2025-04-08
signed-note: add vkey encoding

FiloSottile:


signed-note.md
```diff
@@ -139,3 +140,35 @@ The key ID is generated as recommended above, with the public key encoded as 32
...
+    <key name>+hex(32-bit key ID)+base64(signature type || public key)
+
+Verifier keys (or *vkey*s for short) are encoded as text strings composed of
+three parts separated by a plus (U+002B) character: the key name, the hex
+encoding of the key ID, and the base64 encoding of the concatenation of the
+signature type and the public key material.
+
+The signature type MUST match the key ID, and a parsed vkey SHOULD NOT be used
+to verify any other signature type. (E.g. a `0x01` vkey SHOULD NOT be used to
+verify `0x04` cosignatures, despite both being backed by Ed25519 public keys.)
```
niels-moller, line 155: I think it would be clearer to write that the key id and signature type must match the corresponding fields in signature lines this key is intended to verify, since that's what's generally applicable and I think it corresponds to intended usage. I note that the way key ids are constructed above is phrased as just a recommendation. E.g., we might define signature algorithms 0x11 and 0x12, where both use completely randomized keyids, then there's no way to examine the vkey and check that signature type matches the key id.

Then for 0x01 and 0x04, it *is* possible to recompute the proper keyid from the other fields in the vkey. It makes some sense for an implementation to check that, but not sure if we should recommend or even require that?

Taking a step back, it's unclear to me if and where we should have responsibility for validating key ids. Some preliminary thoughts below.

Say you give me a vkey, saying that's the key you're going to use for signing some notes (say we use signature algorithm 0x1, for simplicity), and that vkey has a non-spec keyid 0x1234. And then I receive a signed notes from you, and they have the same keyname, same 0x1234 key id and a valid Ed25519 signature. Should that signature be accepted or not? Say we have no preprocessing of the vkey, we just have a verify function that takes two strings as input, the signed note and the vkey, and returns a bool saying if the note is valid.

I think there's a lesson from https://hdevalence.ca/blog/2020-10-04-its-25519am/ that we need to be crystal clear on what checks to do to determine if a signature is valid or not. And to me, it gets messy to require strictness in the vkey part of the spec, which should apply to any algorithm and can't make any assumptions on how the keyid is constructed.

If we want to be strict, then I think the section on Ed25519, algorithm 0x01, above, should say that for a signature line with algorithm 0x01, the key id bytes in the signature line *must* correspond to the keyname and the pubkey used to verify the Ed25519 signature: If the Ed25519 signature is valid for for some configured key (with some arbitrary way of configuration and matching it to signature lines), verifier must still recompute keyid from that pubkey and the keyname on the signature line, and if it differs from the received keyid, the signature line must be rejected. I.e., it's something that is specified as part of verifying a signed note, not specific to the vkey representation at all.

And given than the keyid isn't cryptographically strong anyway, I would be tempted to not require verifiers to check it: If I have a vkey for algorithm 0x01, and key name, key id and algorithm match a signature line in a signed note, then I do the Ed25519 verify, and if that passes, I'm happy, and I don't care if the key holder generated keyid according to spec or not, it's just an id assigned by the key holder.

Issue #116, opened 2025-03-06
PHC string format: interest in upstreaming?


tarcieri:
I'm currently the only maintainer of the Password Hashing Competition string format specification at:

https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md

It was suggested to me that it could be upstreamed here, which seems like a good idea.

I think ideally it could be extended with algorithm-specific documentation for how to use it with other algorithms beyond Argon2 (e.g. PBKDF2, scrypt) where PHC encodings for those algorithms are being used in the wild.

FiloSottile:
I would love this. We're actually stalling better APIs for password hashes in Go because of the lack of standardized crypt prefixes. See https://github.com/golang/go/issues/16971#issuecomment-619261508 and https://github.com/golang/go/issues/16971#issuecomment-1241660424.

How would we call it, and are you volunteering as maintainer?

tarcieri:
I can continue maintaining it. There hasn't been much maintenance burden except trying to document how it's been used in the wild by Argon2.

As for a name, perhaps something like `phc-string-format` or `phc-sf`?

FiloSottile:
> As for a name, perhaps something like `phc-string-format` or `phc-sf`?

Hmm, if the scope is going to include crypt prefixes for scrypt and PBKDF2 and so on, PHC feels a bit out of place?

tarcieri:
The spec as written covers just the PHC string format.

w.r.t. algorithms besides Argon2, they generally have an existing Modular Crypt Format encoding, which is effectively `$prefix$anythinggoesafterhere`, where the second part often uses bespoke variants of Base64 which are unique to a particular algorithm. Trying to produce a comprehensive document which covers all of the MCF encodings sounds like quite an undertaking, although perhaps this spec could be expanded into something like that.

However, some tools choose to standardize on the PHC encoding, and so there are also separate, incompatible PHC encodings for the same algorithms which use the PHC format's "B64" encoding and other syntax conventions, though they use distinct prefixes from the MCF encodings and thus are unambiguous.

In cases where there's in-the-wild use of these PHC encodings, it would be nice to document them so we can ensure implementations interoperate.

FiloSottile:
I find `phc-string-format` a bit opaque, and kinda wish there was "password" and/or "hash" spelled out, but I don't have good ideas for how to do that.

[...]

PR #134, opened 2025-06-05
PHC Strings: new spec

tarcieri:
Imports the Password Hashing Competition's string format specification, a rigorously specified `crypt(3)`-alike password hash format, originally located at:

<https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md>

Also includes the specification for using the format with Argon2.

Encodings are also available for PBKDF2 and scrypt, but are not (yet) specified in this document.

As discussed in #116.

[Rendered](https://github.com/tarcieri/C2SP/blob/phc-string-format/phc-strings.md)

### DIFF
--- a/phc-strings.md
+++ b/phc-strings.md
@@ -1,17 +1,16 @@
-# PHC string format
+# PHC Strings
 
-The [Password Hashing Competition] (PHC) string format is a storage format for password hashes
-which includes all the relevant parameters needed to identify the password hashing algorithm used
-to produce a given hash and verify it against a provided password.
+[Password Hashing Competition] (PHC) Strings are a storage format for password hashes which includes
+all the relevant parameters needed to identify the password hashing algorithm used to produce a
+given hash and verify it against a provided password.
 
-PHC strings look like traditional [`crypt(3)`] password strings (a.k.a. "Modular Crypt Format"
-or MCF) and begin with `$<id>$...`. The PHC string format can be thought of as a more
-rigorously-defined subset of this traditional format, amenable to purely stack-based implementations
-and low-resource embedded systems.
+PHC Strings look like traditional [`crypt(3)`] password strings (a.k.a. "Modular Crypt Format"
+or MCF) and begin with `$<id>$...`. They're effectively a more rigorously-defined subset of this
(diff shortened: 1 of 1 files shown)


petersuttondev:
Hi, I've just implemented a parser for this and there are a couple of things that I wasn't sure on from the spec. I'm not sure where the best place to report them is, but I'll start here.

Here's my implementation: https://gist.github.com/petersuttondev/9b64c74ed15526cf775dea24d6173e04
(Most of the parsing is done by the regular expression. I use a library to convert the Python RE into the JS version).

1. Are duplicate parameter names allowed (I assumed not)?
2. Are parameter values allowed to be empty `name=` (I couldn't see a reason to prevent this)?
3. Do versions have to be in a canonical form, e.g., is both `1` and `01` allowed and if so are the equal?
4. Does version have to be positive or just non-negative (i.e., allow version `0`)?
5. Can salt or salt and hash be present but empty? I think it can be unambitiously parsed but seems useless.

Thanks, Pete.

tarcieri:
@petersuttondev that's some good feedback, thanks! I can try to add clarifications.

To go through the list:

1. no, duplicate parameter names are not allowed
2. I also don't see a reason to disallow it
3. I think we can say that version numbers need to follow the decimal encoding, which disallows unnecessary leading zeros
4. This is an important followup to the previous as the decimal encoding allows negative values. I would suggest negative is disallowed, zero-or-greater is allowed, i.e. `0` is a valid version.
5. I think in lieu of that they should be omitted entirely

tarcieri:
Alright, rebased, added an intro, and also the requested clarifications. I think this one is finally ready!

cc @C2SP/stewards

petersuttondev:
I know almost nothing about crypto (besides 1 uni course 15 years ago) and I'm not really sure the intended applications for this spec. But I was just thinking about empty salts and I think empty salt could be useful. If nothing else than for education about why salts are used.

`$md5$$286755fad04869ca523320acce0dc6a4`
`$md5$salt$4c5a18a87bf784a11495216008585c67`

FiloSottile:
You can now remove the scaffolding files from this PR, and merge it independently.

PR #136, opened 2025-06-14
Clarified wording of optional partial tiles for  static-ct-api.md

winstonDeGreef:
The MUST and subsequent MAY construction for conveying the requirement for hosting partial tiles was confusing. It took me a bit to understand that this MUST isn't absolute.

I feel qualifying the requirement directly conveys the message more clearly than creating an exception with MAY.


winstonDeGreef:
If this is accepted the sentence in tlog-tiles.md should probably also me changed.

PR #138, opened 2025-06-20
tlog-tiles: Describe how to prune the left edge of a log

davidben:
Turns out Merkle Trees support these already.

(Although I work nearby, *don't* treat this as a signal of Chrome's CT policy goals. I'm not the person to make that call and wrote this up as part of some related but different work.)

CC @FiloSottile

### DIFF
--- a/tlog-tiles.md
+++ b/tlog-tiles.md
@@ -173,7 +173,7 @@ tree size of the log's current checkpoint. In an unpruned log, the minimum index
 is zero.
 
 In response to a request for a full tile `<prefix>/tile/<L>/<N>`, or an entry
-bundle `<prefix>/tile/entries/<N>`, the log MAY return a 410 or 404 response if the
+bundle `<prefix>/tile/entries/<N>`, the log MAY return an error response if the
 resource's end index, defined above, is below or equal to the minimum index.
 
 Logs are RECOMMENDED to implement pruning by denying HTTP requests, rather than


tlog-tiles.md
```diff
@@ -90,6 +90,10 @@ Merkle Tree Hashes, with *i* from 0 to 255:
 
     MTH(D[(n * 256 + i) * 256**l : (n * 256 + i + 1) * 256**l])
 
+The tile's *start index* is defined to be `n * 256**(l+1)`. It's *end index* is
```
FiloSottile:
```suggestion
The tile's *start index* is defined to be `n * 256**(l+1)`. Its *end index* is
```

tlog-tiles.md
```diff
@@ -154,6 +159,60 @@ compatibility concerns. If they do, they SHOULD also serve entry bundles, and
...
+MAY be *pruned*. Pruning makes some prefix of the log unavailable, without
+changing the tree structure.
+
+Logs maintain a *minimum index* value. The minimum index is the index of the
+first log entry that the log publishes. It MUST be less than or equal to the
+tree size of the log's current checkpoint. In an unpruned log, the minimum index
+is zero.
+
+In response to a request for a full tile `<prefix>/tile/<L>/<N>`, or an entry
+bundle `<prefix>/tile/entries/<N>`, the log MAY return a 404 response if the
```
FiloSottile:
```suggestion
bundle `<prefix>/tile/entries/<N>`, the log MAY return a 410 or 404 response if the
```

Requiring 410 would be nicer but would make hosting on e.g. S3 impossible. It'd be nice for logs that have space for that logic in the read path to do this, though.

tlog-tiles.md
```diff
@@ -154,6 +159,60 @@ compatibility concerns. If they do, they SHOULD also serve entry bundles, and
...
+
+This pruning criteria allows a log client to obtain:
+
+* Any available entry
+* The hash value for any available checkpoint
+* An inclusion proof for any available entry to any containing checkpoint
+* A consistency proof between any two available checkpoints
+
+Pruning is similar to the practice of [temporal sharding][] of logs, except it
+does not change the structure of the tree or the identity or the log. This means
```
FiloSottile:
```suggestion
does not change the structure of the tree or the identity of the log. This means
```

tlog-tiles.md
```diff
@@ -154,6 +159,60 @@ compatibility concerns. If they do, they SHOULD also serve entry bundles, and
...
+all existing proofs remain valid, and existing log clients remain compatible
+with the pruned log.
+
+This document defines *how* to prune a log, but not policies around *when* a log
+may be pruned. Log ecosystems that permit pruning SHOULD define retention
```
FiloSottile:
```suggestion
This document defines *how* to prune a log, but not policies around *when* or *if*
a log may be pruned. Log ecosystems that permit pruning SHOULD define retention
```

FiloSottile (approved):
LGTM, but I would like a +1 from another @C2SP/tlog-tiles maintainer.

tlog-tiles.md
```diff
@@ -154,6 +159,60 @@ compatibility concerns. If they do, they SHOULD also serve entry bundles, and
...
+misconfiguration if the minimum index was set to too high a value.
+
+An entry is said to be *available* if its index is greater than or equal to the
+minimum index. A checkpoint is said to be *available* if its tree size is
+greater than the minimum index.
+
+This pruning criteria allows a log client to obtain:
+
+* Any available entry
+* The hash value for any available checkpoint
```
mhutchinson: What does the hash value for an available checkpoint mean?

tlog-tiles.md
```diff
@@ -154,6 +159,60 @@ compatibility concerns. If they do, they SHOULD also serve entry bundles, and
...
+first log entry that the log publishes. It MUST be less than or equal to the
+tree size of the log's current checkpoint. In an unpruned log, the minimum index
+is zero.
+
+In response to a request for a full tile `<prefix>/tile/<L>/<N>`, or an entry
+bundle `<prefix>/tile/entries/<N>`, the log MAY return a 410 or 404 response if the
+resource's end index, defined above, is below or equal to the minimum index.
+Logs are RECOMMENDED to implement this by denying HTTP requests, rather than
+deleting the resources from storage. This allows the log to easily recover from
+misconfiguration if the minimum index was set to too high a value.
```
mhutchinson: While this is only a recommendation, it's against the grain of how we have designed serving tiled logs. To implement this recommendation, you would need to have custom code on the read path which checks for this condition and changes the response code & result. We have designed Tessera so that the read path is directly served from static storage without any custom code on that path for 3/4 of our implementations. Implementing this recommendation would thus require Tessera deployments to route all read requests through a binary. This changes scale requirements and load-balancing / routing configuration.

tlog-tiles.md
```diff
@@ -154,6 +159,60 @@ compatibility concerns. If they do, they SHOULD also serve entry bundles, and
...
+
+Pruning is similar to the practice of [temporal sharding][] of logs, except it
+does not change the structure of the tree or the identity of the log. This means
+all existing proofs remain valid, and existing log clients remain compatible
+with the pruned log.
+
+This document defines *how* to prune a log, but not policies around *when* or *if*
+a log may be pruned. Log ecosystems that permit pruning SHOULD define retention
+policies for how long entries must be available. For example, an ecosystem might
+require that entries remain accessible for 6 months after they expire.
```
mhutchinson: Thoughts on making this paragraph indicate that in the absence of any explicit policy on pruning, logs MUST not prune?

I don't think is restrictive, and it makes ecosystems building on this spec safer by default. i.e. I think an unspecified or poorly specified pruning policy is likely to lead to verifiers not being able to get data, and logs claiming that they're behaving as per the tlog-tiles spec. While the fallout for this is being resolved, believers in the ecosystem can still be relying on data that was pruned from the log and will never be verified.

tlog-tiles.md
```diff
@@ -154,6 +159,60 @@ compatibility concerns. If they do, they SHOULD also serve entry bundles, and
...
+check if the retention policy has been honored. To mitigate this, log clients
+SHOULD be periodically updated with a lower bound on which entries to accept.
+This lower bound SHOULD be at least the log's minimum index at the time and MAY
+be higher, such as the index of the first available, currently unexpired log
+entry.
+
+TODO: Some HTTP endpoint for fetching the minimum index? The semantics would be
+something like: serving a minimum index equivalent to returning 404 from the
+tiles that would be deleted by the pruning criteria, including when evaluating a
+log client's availability policies.
```
mhutchinson, line 211: This part is my largest concern with this proposal: how does a log announce its current minimum size in a globally consistent way? Currently this opens a door for a log that has gone rogue but wants to extend the time window for performing its attack. Such a log can answer "not available" to verifiers/monitors/auditors requesting entries, but keep serving malicious entries to clients.

tlog-tiles.md
```diff
@@ -154,6 +159,60 @@ compatibility concerns. If they do, they SHOULD also serve entry bundles, and
...
+An entry is said to be *available* if its index is greater than or equal to the
+minimum index. A checkpoint is said to be *available* if its tree size is
+greater than the minimum index.
+
+This pruning criteria allows a log client to obtain:
+
+* Any available entry
+* The hash value for any available checkpoint
+* An inclusion proof for any available entry to any containing checkpoint
+* A consistency proof between any two available checkpoints
```
mhutchinson: This has a big impact on witnessing. Witnesses have 2 modes:
 - For a previously unwitnessed log: TOFU a checkpoint 
 - For a previously witnessed log: update the known checkpoint using a consistency proof and a new checkpoint

Such a witness is now forever stuck if their known checkpoint becomes unavailable.

According to the checkpoint spec "Two checkpoints are inconsistent if a consistency proof can't be constructed from one to the other". Arguably a consistency proof _can_ be constructed between an unavailable and available checkpoint by an actor that knows the pruned parts of the tree, but to someone without this hidden information, the checkpoints should be considered inconsistent.

FiloSottile:
> A lesson learned for myself from this is that we all need to be super clear about _versions_ of spec when implementing ecosystems and code. I'm probably guilty of saying, e.g. "this log implementation/deployment supports tlog-tiles", but that's a dangerous thing to say without a version number!

FWIW this is optional, so even if v1.1.0 were to add pruning, you could still implement it with a policy of no pruning ever. (But also yes! I want to see a lot more use of C2SP spec version numbers!)

tlog-tiles.md
FiloSottile: Many object storage solutions support IAM policies that allow restricting visibility to some paths, and so do static web servers such as nginx and Caddy. Depending on the backend, you can also change the permissions of hidden files. Finally, you can move files to a different directory/path/bucket. There is a spectrum of options for hiding assets without deleting them and without adding a full reverse proxy.

tlog-tiles.md
FiloSottile: Yeah, defaulting to MUST NOT in the absence of an explicit policy sounds good to me (and the general intention already).

tlog-tiles.md
FiloSottile: I don't think this is different from witness policies. Monitors need to know the witness policy of clients/believers, or they can't do their job, and a malicious log could present them with a split view. Monitors will need to know the pruning policy of clients/believers (including the precise minimum index), or they can't do their job.

tlog-tiles.md
FiloSottile: Those witnesses will indeed become stuck and will not produce cosignatures on newer checkpoints, so clients/believers will not trust them. That sounds very WAI to me? It's just one of many ways a log can break its own availability.

tlog-tiles.md
AlCutter: +1

I suspect this would effectively require a change to `checkpoint` for the log to commit to the available range, I'm not sure whether and how witnesses should interact with this; I could imagine witnesses may refuse to cosign checkpoints whose minimum index is not `>=` the previously seen minimum index (if any), similar to how that's already done on checkpoint's size, as a way of preventing oscillation of the minimum index.

FiloSottile: I’m not following, what are you defending against? Logs don’t commit to the client’s witness policy in the checkpoint.

mhutchinson: Changing the checkpoint format is possible, though it may be hard to roll out through the ecosystem. It also sounds like there is scope for the pruning threshold to go up and down ("This allows the log to easily recover from
misconfiguration if the minimum index was set to too high a value.", above).

tlog-tiles.md
mhutchinson: Accepting that this breaks a witness-log relationship forever is the most logically correct outcome. However, I suspect we'll end up with witnesses regularly being asked to implement a feature to delete/forget a log, so that they can TOFU again. I really wouldn't like this.

FiloSottile: That sounds like a flat-out violation of tlog-cosignature, assuming the witness doesn’t change identity, because it would be signing two checkpoints it did not check the consistency of.

As a practical matter, I find the scenario of a years out of date witness far fetched, certainly not a regular occurrence.

mhutchinson: Cool, not only are we both agreed that forgetting history would be bad, we're agreed that it would be spectacularly awful. I can dig it.

As to the other point, I accept that logs that are pruning history from years ago are unlikely to trip any of these wires I'm describing. However, nothing in this proposal precludes logs from pruning stuff in the order of minutes old. I don't think that it should preclude that (because it's a log ecosystem policy matter), but if it's possible, then other ecosystem participants (e.g. witnesses) need to support it (or explicitly not support it, and push back against pressures to make exceptions).

FiloSottile: IMHO they already not support it, by the letter of tlog-cosignature and tlog-witness, but I’m not opposed to reiterating it here.

tlog-tiles.md
davidben: Me too. Done!

tlog-tiles.md
davidben: Probably me just using the word "checkpoint" wrong. Is the rephrasing better?

tlog-tiles.md
davidben: ISTM the answer to "how does a log announce its current minimum size in a globally consistent way" is the same as "how does a log serve its contents in a globally consistent way". I was envisioning pruning as "just" refusing to serve some resources, combined with the ecosystem saying "yeah, actually, I'm okay if you just don't serve those resources".

I'm a little unclear on what committing to the minimum index in the checkpoint does. If a log set it to 1000 and then goes "nevermind, I'm going to set it to 500", sure that's *better* for the ecosystem than before because it's now making more stuff available.

The other reason it doesn't make sense in a checkpoint is that different copies of the log may have different minimum indices. It's really not a property of the log, but of the particular serving instance. (E.g. imagine a mirror that retains things for longer than the log.)

> Currently this opens a door for a log that has gone rogue but wants to extend the time window for performing its attack. Such a log can answer "not available" to verifiers/monitors/auditors requesting entries, but keep serving malicious entries to clients.

I'm not following. Logs generally do not serve things to clients directly in the first place, at least not in the use cases I'm familiar with. Pruned or not, the entry is still in the tree and all proofs for it will check out. When you build a system on top of a pruning log and entry 42 is pruned, you have to design it assuming that clients *will* accept an inclusion proof for entry 42, even though you can't see it. In the same way, when you build a system on top of a regular log, you have account for the possibility of the log failing to serve entry 42.

The natural approach would be to say that you have to believe that entry 42 being expired is a prerequisite for it being pruned. If you don't have *some* notion of expiry, this feature is probably not for you. If monitors see that 42 was pruned, they need to somehow believe that 42 is expired. Of course, 42 is no longer available, but this isn't that much of a contradiction:

Presumably your pruning policy was set so that the ecosystem would have long seen entry 42 already anyway. So there was ample time to evaluate the logging policy.

More importantly, the real load-bearing check is the relying parties. Relying parties should be pushed a regularly-updating low watermark of entries to accept, probably set to not just the minimum index but the first unexpired entry at the time, which should be some 6 months ahead of the minimum index and thus checkable. So if the ecosystem notices "oh yeah, entry 42 got pruned" and there is ever any question about whether that was proper, just immediately push a minimum index of 42 to all the relying parties. (Though, if it was proper, the relying parties probably got an even higher push some 6 months ago anyway.)

Of course, this only protects updated relying parties, but that's actually the game we were playing anyway. Suppose pruning was forbidden but some log just refused to serve 42. There is no cryptographic defense against that. It's just the ecosystem knowing 42 exists but was unavailable. The response would be to either distrust the log, or do a more targeted punch-out of entry 42 in the relying parties... which is exactly what this is.

Except this is better because you actually can just punch-out the expiring horizon ahead of time and then it will be very obvious when something goes wrong because your policy is 6 months behind that horizon.

tlog-tiles.md
davidben: I think realistically you would set your pruning policy such that this would never happen. But also, if it did happen, the log is presumably motivated to recover its wedged witnesses and (one hopes!) still has the data available and can still produce the consistency proof needed. Witnesses are push-based, not pull-based, right? If a log completely shut down its serving interface, but still pushed to witnesses, zero witnesses would notice.

> However, nothing in this proposal precludes logs from pruning stuff in the order of minutes old. I don't think that it should preclude that (because it's a log ecosystem policy matter), but if it's possible, then other ecosystem participants (e.g. witnesses) need to support it (or explicitly not support it, and push back against pressures to make exceptions).

The push vs pull thing aside, if a log ecosystem chose an absurd policy that tolerated O(minutes) pruning, it's presumably only considering witnesses that can stay up-to-date within O(minutes). The more generic witnesses probably don't support that and are not part of that log ecosystem.

I think that's all fine? It's super useful to design for these generic witness networks, but if the generic witness networks look like X and some application _truly_ needs Y, which is incompatible with X, that's fine. The protocol can admit both X and Y without necessarily having them cosign each other.

(Aside: I actually don't think an exception process is _necessarily_ a bad idea for extreme break-glass situations, in applications that need to design for arbitrarily out-of-date relying parties. But that's unrelated to this, so will leave that for another conversation. 😝)

tlog-tiles.md
davidben: Put another way, consider CT temporal sharding today. Mechanically, a CA could well stick a 20-year certificate in a log shard and then that certificate will remain valid long after the log is retired.

The thing stopping that is that that clients stop accepting the old log after it's expected to only contain expired entries. But that's done by pushing an update to the client. Pushing the low watermark to clients is the same thing, except that it doesn't forcibly break continuity with clients that haven't gotten the update. (Today, CT temporal sharding only works because clients are expected to automatically shut off all CT enforcement and fail open on a time bomb.)

tlog-tiles.md
davidben: I added a couple sentences in https://github.com/C2SP/C2SP/pull/138/commits/94b116a204bc65d169a9cbd591ccaabac794cd3d to hopefully address this?

tlog-tiles.md
mhutchinson: Checkpoint was better than "tree head".

I think what you're saying is the client can find the _root hash_ for an available checkpoint. That I follow, but a client can also find the _root hash_ for an _unavailable_ checkpoint too. It's in the checkpoint: https://github.com/C2SP/C2SP/blob/main/tlog-checkpoint.md

What would break if this bullet was removed? To me it indicates a distinction between available and unavailable checkpoints, and I don't think there is one in this regard.

davidben: The log doesn't serve past checkpoints, so only the latest one can be fetched that way.

All I'm trying to say is that, given some tree size N, you have all the intermediate nodes you need to compute `MTH(D[0:N])`. That is possible for `N > minimum_index`, but not necessarily smaller values.

davidben: Also note this isn't some prescriptive list. It's just some list of notable things you can still compute. As for why that's notable, https://davidben.github.io/merkle-tree-certs/draft-davidben-tls-merkle-tree-certs.html contains a few places where you may later find you need to learn the hash values of past views of the log, along with consistency proofs to your point of reference of course. (Though it's more concerned with a [more generalized notion of subtree](https://davidben.github.io/merkle-tree-certs/draft-davidben-tls-merkle-tree-certs.html#name-subtrees).)

davidben: (Switched "tree head" back to "checkpoint".)

tlog-tiles.md
FiloSottile: Maybe, since we expect relying parties to pick and enforce the minimum index and we expect monitors to learn it from relying parties (as they shouldn't trust the log about it), we should not have an endpoint for the log to state what its minimum index is at all?

davidben: Maybe! On the one hand, it's just a number and the log can _already_ communicate the same thing by just sending a bunch of 404s. On the other hand, it does seem to feel better in some vague, ill-defined way to not give so easy a way to get it from them.

The only concrete use case I had was mirrors, who probably want to start life at the log's minimum index (or whatever minimum index they were able to find). But yeah they could get it from the relying parties [edit: fixed typo]. Worst case, they can binary search on the first entry bundle that's available! Which... TBH further hints that maybe serving it or not serving it doesn't mean much.

tlog-tiles.md
```diff
@@ -154,6 +159,72 @@ compatibility concerns. If they do, they SHOULD also serve entry bundles, and
...
+  rate-limited or with limited access.
+
+An entry is said to be *available* if its index is greater than or equal to the
+minimum index. A checkpoint is said to be *available* if its tree size is greater
+than the minimum index.
+
+This pruning criteria allows a log client to obtain:
+
+* Any available entry
+* The root hash for any available checkpoint
```
mhutchinson:
```suggestion
* Sufficient tiles to calculate the root hash for any available checkpoint
```

Or maybe s/calculate/recalculate/. This makes it clearer to me the distinction between available and unavailable checkpoints.

tlog-tiles.md
mhutchinson: This thread is outdated and replaced with https://github.com/C2SP/C2SP/pull/138/files#r2166368559, which summarizes the distinction you clarified in this thread.

tlog-tiles.md
mhutchinson: We're agreed that if a log prunes in such a way that it makes witnessing impossible, then witnessing is impossible. I like the simplicity.

tlog-tiles.md
mhutchinson: Moving them seems like the most likely way to generate the status codes we need. I suspect (and Gemini agrees) that IAM denial will lead to 403 forbidden errors. We could add 403 to the status codes clients should expect, but this feels like exposing how the sausage might be made a little too much?

tlog-tiles.md
mhutchinson: > Put another way, consider CT temporal sharding today. Mechanically, a CA could well stick a 20-year certificate in a log shard and then that certificate will remain valid long after the log is retired.

Mechanically, they could if the operated the log. But it would in violation of [Chrome Log Policy](https://googlechrome.github.io/CertificateTransparency/log_policy.html), which says that `CT Logs must reject logging submissions for certificates whose notAfter timestamp falls outside the certificate expiry range`.

Nit aside, I'm still chewing on this. My feeling is that this can work, especially in established ecosystems such as CT, that are backed by a sensible policy, and things move somewhat slowly. At some level I'm bought into the idea that this is very similar to temporal logs. On the other hand, it's fundamentally different; each temporal shard is its own independent log, and it's only the policy that allows one to consider them as parts of a whole.

Currently, each individual append-only log has a clear remit: as per the [Claimant Model](https://github.com/google/trillian/blob/master/docs/claimantmodel/Logs.md): `All Statements committed to by this Checkpoint are immutable and discoverable`. If any statement (i.e. leaf) committed to by a log is not available to a verifier, they should invoke the arbitration process (or in Filippo's words, starts The Conversation). That's a really concise definition. This now changes, and it's not clear what the claim of a prune-or-append-only-log is. Maybe this isn't an exercise for this spec change, and being clear about what the logs are claiming is now up to any ecosystem that permits a prunable log to define. But hopefully you can see my discomfort at going from a clear definition of a log, to a very unclear definition.

I'll post these thoughts as-is, even though I haven't spent the time I'd like to distill them further. If the above is messy I can revisit with more facts and fewer feelings.

davidben: > Mechanically, they could if the operated the log. But it would in violation of [Chrome Log Policy](https://googlechrome.github.io/CertificateTransparency/log_policy.html), which says that CT Logs must reject logging submissions for certificates whose notAfter timestamp falls outside the certificate expiry range.

Ah yeah. (I've been spending the last few months in [Merkle Tree Certs](https://davidben.github.io/merkle-tree-certs/draft-davidben-tls-merkle-tree-certs.html) land, so I have CA-operated logs on the brain.) We could say something similar in this universe by saying CT Logs must reject submissions whose notAfter is more than N days in the future. That's probably a good idea for any Log Policy with a pruning policy. Then your policy can be phrased as "entries will be retained for at least N days + 6 months" instead of the contents of individual entries.

> I'll post these thoughts as-is, even though I haven't spent the time I'd like to distill them further. If the above is messy I can revisit with more facts and fewer feelings.

Messy is fine! And yeah the discomfort makes sense. I've similarly struggled in taking this from a "ah yeah, this should work" idea to concrete definitions, and I think they come from the same place. Hopefully we can find the right way to express this.

The Claimant Model document is neat. Would it work to say that it now promises something like:

> All Statement<sup>DOMAIN</sup> committed to by this Checkpoint are immutable and discoverable, up to RetentionPolicy<sup>DOMAIN/LOG/whatever</sup>

Now, where this gets messy is...

> If any statement (i.e. leaf) committed to by a log is not available to a verifier, they should invoke the arbitration process (or in Filippo's words, starts The Conversation).

This is tricky because the RetentionPolicy is almost certainly going to be depend a bit on the contents of the statements. The root of what makes this or temporal sharding or anything else plausible is that statements eventually stop mattering to people. But if the statement is not available, it's hard to know whether to invoke arbitration.

That said, once you *do* invoke arbitration, it is possible the policy is verifiable. Anyone can produce and authenticate the missing statements. We could probably go define an ecosystem of timestamp assertions as evidence to feed into this arbitrary process. (If some witness has previously signed "yup, all entries in checkpoint 42 expire by January 2025", you could produce that claim.)

Or the result of arbitration could be "okay, we're not sure what happen, but we don't really care because we've since told all the believers to disregard everything".

And so we can go back and say all those things that would be considered during arbitration would be a reason whether to invoke the arbitration process. Definitely not as crisp as it was before, but it's also already the case that any log might fail to serve any entry at any time. Trusted mirrors feel like a more robust story for that than everyone constantly downloading every resource from every log every day, just to make sure they're holding up their end of the bargain.

Also worth noting that "discoverable" is already not that crisp a notion. If I rate-limit or apply DoS protections on my log and it happens that *you specifically* can never discover some statement, but the statement are still produced during arbitration, have I held up my end of the bargain? At what point does rate limiting transition to unavailability?

Pruning is really just tolerated unavailability. We could even decide, nah, the pruned entries cannot be *completely* unavailable, but you can rate-limit to an extreme degree if you want.

> prune-or-append-only-log

I think it's worth to being clear that the *log* itself is still append-only. Even if entries 0-100 have been pruned, they are *still* committed to by every checkpoint. You cannot suddenly go back and pretend those entries are something else.

Of course, if the entries are _unavailable_, the commitment is only so useful, but it means one can still produce evidence when there's a question.

tlog-tiles.md
davidben: Hmm, doesn't that apply to everything in this list? The tiled interface doesn't serve inclusion proofs. You have to calculate some of the nodes yourself. I figured that was just implicit through the entire tiled interface.

mhutchinson: You're right that this is implicit in most of the others too. What I'm trying to steer this away from is the only conclusion my brain jumps to which is "of course you can get the root hash for an available checkpoint, it's in the checkpoint!". That conclusion can't be jumped to for any of the other entries in this list; you clearly have to go make some requests from the log to find the info.

Another way of looking at what I think this is saying, is that a log client could obtain the [compact range](https://github.com/transparency-dev/merkle/blob/main/docs/compact_ranges.md) for any available checkpoint?

davidben: Hehe, yeah, that's why I tried avoiding the word checkpoint. :-) Maybe part of the problem is the word "available"? Because if it says "available checkpoint", then it really reads like the checkpoint is already available in the English sense.

> Another way of looking at what I think this is saying, is that a log client could obtain the [compact range](https://github.com/transparency-dev/merkle/blob/main/docs/compact_ranges.md) for any available checkpoint?

Ah yeah, I think basically you ~~delete~~ [edit: stop serving] everything but the root hashes of every subtree in the compact range for `[0, minimum_index)`. But those are also a lot of concepts that don't yet exist in tlog-tiles.

mhutchinson: > Ah yeah, I think basically you stop serving everything but the root hashes of every subtree in the compact range for `[0, minimum_index)`. But those are also a lot of concepts that don't yet exist in tlog-tiles.

Yeah, that's true for the process of pruning. But my observation was the flip side of the coin:
 - for an available checkpoint, all nodes in the compact range expressing this tree size will be available
 - for an unavailable checkpoint, some nodes from the compact range may (will?) not be served by this log

Is this something like what this bullet point is trying to express?

mhutchinson: The other option is that you're saying something much simpler, which is that the log always presents an available checkpoint (and this checkpoint, by definition, contains a root hash)?

tlog-tiles.md
```diff
@@ -154,6 +159,72 @@ compatibility concerns. If they do, they SHOULD also serve entry bundles, and
...
+index value MUST be set to zero. Retention policies SHOULD be set so that there
+is ample time for ecosystem participants to stay ahead of the minimum index.
+
+If an entry is pruned and historical data is not available, e.g. another copy of
+the entries or a trusted summary of expiration dates, it may not be possible to
+check if the retention policy has been honored. To mitigate this, log clients
+SHOULD be periodically updated with a lower bound on which entries to accept.
+This lower bound SHOULD be at least the log's minimum index at the time and MAY
+be higher, such as the index of the first available, currently unexpired log
+entry.
```
AGWA: This paragraph makes me uneasy for the following reasons:

1. As a monitor, my foremost concern is ensuring I can see every entry which is accepted by relying parties. If RPs don't enforce a lower bound on entries to accept, then I would have to either scan through other copies of the log looking for unexpired entries (potentially very expensive) or rely on someone else's summary of expiration dates (sounds complicated, and could be wrong). I would much rather learn about the lower bound from RPs and start monitoring from that position. This is simple and directly relates to the property I care about.
2. We've had several CT logs violate their temporal range. Despite this being a policy violation, they were not retired, though they were removed on the intended schedule, with unexpired certificates still in the log. I expect the equivalent scenario to happen with Merkle Tree Certs - CAs including certificates expiring far in the future, disrupting the normal pruning of the log. It may be less disruptive to continue pruning as usual, even if it means an unexpired certificate gets pruned. That should be acceptable as long as RPs reject the entry.

So I think that the concept of verifying the retention policy should be dropped and everything centered around the lower bound that's accepted by RPs:

> If a log is subject to pruning, relying parties MUST be periodically updated with a lower bound on which entries to accept. This lower bound MUST be at least the log's minimum index and MAY be higher, such as the index of the first available, currently unexpired log entry. Monitors MAY disregard entries before the lower bound accepted by relying parties.

davidben: Ah thanks, I like that! Will update. Some minor thoughts/nitpicks:

* Does tlog know what a relying party is? It's a bit more general than CT, so I suspect we have to talk about "log clients" in general, which would encompass a pretty wide range of entities.

* I *think* we want this to be a SHOULD rather than a MUST (and then s/accepted by relying parties/accepted by up-to-date relying parties/) because we can't guarantee every log client is up-to-date. A MUST would read as if a client should shut itself off when it's too old which... honestly is true of CT today (fails open when stale), but it's a little unfortunate too.

* If it's a SHOULD, it's probably worth keeping some discussion about being able to check it with historical data, so that you still can reason about older relying parties. But I agree the RP enforcement should be the focus and is the real load-bearing part.

In some sense, older RPs is the main reason to prune instead of temporally shard. Pruning means that (MyFancyLog, min=42) and (MyFancyLog, min=1000) are still the "same" log, and RPs still on min=42 will interoperate with min=1000. I'm particular interested in this in the context of [Merkle Tree Certificates](https://davidben.github.io/merkle-tree-certs/draft-davidben-tls-merkle-tree-certs.html) because failing open when stale is a bit less comfortable when it's also the CA part of the check.

davidben: > I'm particular interested in this in the context of [Merkle Tree Certificates](https://davidben.github.io/merkle-tree-certs/draft-davidben-tls-merkle-tree-certs.html)

Although, TBH, I also personally think that a CT ecosystem that relied less on the fail open timebomb, if we could still keep all the operational benefits of that timebomb, would be better, particularly for the long tail of clients that might want to enforce CT. I'm hoping MTCs can step us closer to that.

tlog-tiles.md
davidben: How's the new text look? If it's still confusing, I'll just delete the offending line and leave "you can compute past root hashes if you need them" as an exercise to the reader. I feel like this is useful to call out, but if it mostly confuses folks, meh. :-)

tlog-tiles.md
davidben: How's the new text look?

AGWA: I went with "relying party" because "log client" could encompass monitors and submitters, but it's true that it's not defined anywhere in the spec. (RFC 6962 would say "TLS client" but that's even more CT-specific.) The spec would definitely benefit from some definitions of the different types of clients.

I see your point about the MUST suggesting that an out-of-date client needs to shut itself down.  But I'm concerned that the SHOULD in [the new text](https://github.com/C2SP/C2SP/pull/138/files#diff-fee56f2c4a01e6bf0c033bf2c4c63e888573304a58c63cd3a6f507286c66ef8cR221-R228) combined with "This mitigates the possibility of a log being pruned outside its retention policy, and historical data is not available to determine this" could lead implementers to not even bother with an RP updating mechanism if they think that historical data will always be available or that they'll always be able to distrust a log which violates its retention policy.

How about this (sticking with "log client" for now but I still think a more precise term is needed):

> If a log is subject to pruning, log clients MUST be periodically updated with a lower bound on which entries to accept. This lower bound MUST be at least the log's minimum index and MAY be higher, such as the index of the first available, currently unexpired log entry. Monitors MAY disregard entries below the lower bound accepted by up-to-date log clients.
>
> If a log client is unable to update itself, it MAY continue accepting entries per the last known lower bound. Monitors that wish to detect entries accepted by out-of-date clients, in particular entries that were pruned in violation of the log's retention policy, MAY consult a historical copy of the log if available.

davidben: Yeah, "log client" covers more folks than is ideal. I agree that's not a great term. Will noodle.

Part of the awkwardness is that tlog-tiles is a pretty low-level primitive. Really this should be specified a layer up, in the actual application. I think the point of the tlog-tiles text here is to some guidance. (That further points to SHOULD over MUST.)

> [...] lead implementers to not even bother with an RP updating mechanism if they think that historical data will always be available or that they'll always be able to distrust a log which violates its retention policy.

I'm confused about the second bit. How would you be able to distrust such a log without also being able to push new lower bounds?

Speaking generally and with an individual hat, I think distrusting a log that violates the retention policy _and refuses to repair it_ is... not an implausible choice for a relying party ecosystem. Pruning is mechanically the same as failing to serve some entries. It doesn't permanently altered the tree or anything, just some tiles return 404. If a log were offline, I imagine relying parties wouldn't _immediately_ distrust the log, but if the log can't repair itself, I expect RPs _would_ probably distrust it. ISTM a bad minimum index is similar.

If anything, since repairing is _just_ a matter of rewinding the minimum index, it seems you can set stronger expectations on the log. Then again, you can also mitigate over-pruning with a mirror (see https://github.com/C2SP/C2SP/pull/140), so perhaps no one would care.

I think distrust is actually a good analogy here. Any time you have a list of trusted external entities, you have to account for distrust and thus have _some_ answer for how those distrusts get to you. At the same time, clients differ. I don't think low-level protocols should have MUST-level mandate on update paths is a good idea. Imagine a Python application using certifi or a Go one using the roots in the Go standard library. You really should be rebuilding those binaries, and hopefully are, but Python and Go can't guarantee that happens and thus cannot actually comply with any MUST-level requirement. And yet the ecosystem does also generally assume that, once a thing is distrusted, we don't have to pay attention to it, even though there will always be unupdated clients.

TBH, I don't think the historical data thing is as scary as it sounds. If you are a monitor that only just came into existence, you don't have much to go on and should just trust up-to-date RPs' lower bounds. But if you already exist, you're presumably not going to be so behind that the minimum index passes you. (If you were, we have worse problems!) That means you actually saw all those historical certificates and could very efficiently just maintain a little table with, say, the expiration cutoffs for every month. As long as you have table entries more frequent than the retention period, that's plenty.

Or maybe that table is stored in something [vindex](https://github.com/transparency-dev/incubator/tree/main/vindex)-like with a witness network that attests to it, I don't know.

AGWA: > I'm confused about the second bit. How would you be able to distrust such a log without also being able to push new lower bounds?

I was thinking of an implementation that has an updatable log list but there's nowhere in the list for a lower bound and no code that checks entries against the lower bound because the implementer thought this was an optional part of the spec.

> Speaking generally and with an individual hat, I think distrusting a log that violates the retention policy and refuses to repair it is... not an implausible choice for a relying party ecosystem. Pruning is mechanically the same as failing to serve some entries. It doesn't permanently altered the tree or anything, just some tiles return 404. If a log were offline, I imagine relying parties wouldn't immediately distrust the log, but if the log can't repair itself, I expect RPs would probably distrust it.

The scenario that worries me is a log wrongly incorporating a certificate expiring 5 years in the future in a world where certs are supposed to be 45 days max. There's no way to repair that mistake, and if it happens at index X, then the log's minimum index becomes stuck at X for 5 years while the unpruned part of the log gets very large, neutralizing the benefits of pruning. RPs could distrust the log for violating a hypothetical requirement not to accept certificates expiring too far in the future, but I think they would hesitate to do that, just as they've declined to distrust CT logs which violated their temporal ranges. RPs would likely be better served by allowing the log to intentionally violate its retention policy by pruning entry X before it expires, and rolling forward the lower bound so up-to-date clients don't accept entry X.

(This scenario is X509-specific but could happen in any ecosystem where logged entries have variable validity periods.)

> TBH, I don't think the historical data thing is as scary as it sounds. If you are a monitor that only just came into existence, you don't have much to go on and should just trust up-to-date RPs' lower bounds.

I absolutely agree.  I only added the last sentence in my proposed text because you mentioned the need to reason about older relying parties.  I would prefer to omit discussion of out-of-date RPs and historical data since I think it muddies the waters about RP lower bound enforcement being the load-bearing check.

I believe we're largely in agreement. I definitely don't think the spec should be mandating update paths. I just think the latest language is still unclear about the importance of RPs enforcing a lower bound.

davidben: > I was thinking of an implementation that has an updatable log list but there's nowhere in the list for a lower bound and no code that checks entries against the lower bound because the implementer thought this was an optional part of the spec.

Oh! Yes, we should definitely make it clear not to do that.

(For [Merkle Tree Certs](https://davidben.github.io/merkle-tree-certs/draft-davidben-tls-merkle-tree-certs.html#name-trust-anchors), we've currently written that a list of revoked ranges is just part of the definition of a trust anchor. That tidily expresses this pruning horizon, individual bad or undisclosed entries, an entire undisclosed subtree, general revocation, and SCTNotAfter-style distrust all in one.)

> The scenario that worries me is a log wrongly incorporating a certificate expiring 5 years in the future in a world where certs are supposed to be 45 days max. There's no way to repair that mistake, and if it happens at index X, then the log's minimum index becomes stuck at X for 5 years while the unpruned part of the log gets very large, neutralizing the benefits of pruning. [...]

Ah, good call! I had not connected those dots.

Thinking about how to phrase this, what if instead of trying to classify log clients and constrain them, we said that log *applications* should account pair the pruning policy with a mechanism to reject log entries that are below some lower bound. And then we can just give an example of what CT can do, making full use of the relying party, monitor, etc., assumption. But then the true load-bearing part would be left to the layer above?

AGWA: Yeah, that sounds like a good approach.

davidben: Took another stab at it. What do you think?

On reflection, I weakened the text on monitors disregarding entries slightly. My thinking is: a monitor might well care about recently-expired certificates below the lower bound too. Certainly it's crucial to pick up all unexpired certificates, and if that's all you pick up, that's probably fine. But, especially as certificate lifetimes go down, if my site had a bad cert that expired last week, and I only _just_ noticed, that's probably still notable to the ecosystem.

Also pulled this into its own subsection because this was getting a bit long. :-)

AGWA: New text looks good!

tlog-tiles.md
```diff
@@ -154,6 +159,88 @@ compatibility concerns. If they do, they SHOULD also serve entry bundles, and
...
+unavailable, without changing the tree structure.
+
+Logs maintain a *minimum index* value. The minimum index is the index of the
+first log entry that the log publishes. It MUST be less than or equal to the
+tree size of the log's current checkpoint. In an unpruned log, the minimum index
+is zero.
+
+In response to a request for a full tile `<prefix>/tile/<L>/<N>`, or an entry
+bundle `<prefix>/tile/entries/<N>`, the log MAY return a 410 or 404 response if the
+resource's end index, defined above, is below or equal to the minimum index.
```
bwesterb, line 177: "or equal" is inconsistent with "The minimum index is the index of the first log entry that the log publishes" above.

tlog-tiles.md
bwesterb: > The only concrete use case I had was mirrors

Not just mirrors I guess. The log might've been running for a while (and have pruned entries) before being accepted or submitted to a program. Or perhaps someone wants to set up a new program. We can gossip starting indices, but that seems a bit inconvenient. We can also just call it a starting-index hint, and not make it binding. The RP has to have the minimum index its mirror assumed shipped out of band to make sure it doesn't get tricked by a pruned entry.

bwesterb (approved):

tlog-tiles.md
davidben: Oh, did I get the boundary condition wrong? I _think_ it's right? start/end are a half-open interval. We're hiding `[0, min_index)` and keeping `[min_index, infinity)`. If I have a tile `[start, end)` and `end = min_index`, it doesn't intersect with the kept range, so it's okay[*] to prune it.

[*] More specifically, if it doesn't intersect, you can prune the *leaves* of the tile but you *may* still need to the root of the tile. However, the root of the tile is one of the leaves of the parent tile, so we can still delete this tile and evaluate the condition at parent to figure out if the parent tile is needed.

bwesterb: Oh, no, you're right, nvm. 🤦

tlog-tiles.md
mhutchinson: The new text says:

"[This pruning criteria allows a log client to obtain] The root hash for any checkpoint whose tree size is greater than the minimum index"

This reads more clearly because we don't have to deal with the term "available", but my confusion still reigns because the following is also true:

"[This pruning criteria allows a log client to obtain] The root hash for any checkpoint"

Again, because the checkpoint directly contains the hash.

I don't object to this entry because it's wrong, it just appears to be trying to say something that I don't see. Perhaps it's some form of conceptual colour blindness on my part :-)

mhutchinson: Ah, wait!

Perhaps it is the word checkpoint here that's the problem. If you have a checkpoint, you have the root hash. But if you don't have the checkpoint, you can still compute a root hash for any tree of size N (N > min index). Is that the observation?

In which case, changing this line to the following might fix my mind trap:

"[This pruning criteria allows a log client to obtain] The root hash for any tree whose size is greater than the minimum index"

davidben: Yeah, that is indeed the observation.

I mean, I did switch to something of that nature in https://github.com/C2SP/C2SP/pull/138/commits/c8d6a5442deda69207dfcdc7d560ba4c6552f5df. Then I switched it back in https://github.com/C2SP/C2SP/pull/138/commits/34bd3f0034ab1cb9024be76456d08a65b53fba19 because you said you preferred checkpoint. :-P Though this was before fixing the confusing "available" terminology, so perhaps we were not yet on the same page yet.

mhutchinson: Yeah, I see the misunderstanding now. I interpreted "tree head" in that commit as meaning a published (signed) tree head, which also contains a root hash, and thus no different in my eyes, except using older terminology.

davidben: How's https://github.com/C2SP/C2SP/pull/138/commits/3a69677d3d5b9b7ac336cceb5ee4b323433c20c7

mhutchinson: IMO the other two lines didn't need changing because one always ought to be doing inclusion and consistency proofs with respect to checkpoints. That said, it isn't wrong and I can understand it. LGTM.

davidben: Ah yeah I changed those because, if checkpoint refers to the actual structure and not a general snapshot, you might want to generate proofs at points that never saw a checkpoint.

For example, suppose a log stamped checkpoints at 100, 200, and 300, but the mirror skipped a beat and went from 100 to 300. You might still care about consistency between 200 and 300 and you might want to hit a mirror instead of the origin.

tlog-tiles.md
davidben: What if we just said "an error response"? (Updated the PR.)

davidben:
How's this looking?

mhutchinson:
@davidben thanks for taking the time to write up this pruning suggestion, and thanks for pinging. Apologies for not getting back to this sooner. Al and I agreed a draft response and then PTO and business interrupts pushed this out of context.

How do people feel about leaving this Pull Request in a semi-approved state for the time being, ready to be fully integrated at the point where a dependency on this feature is needed? For example, if/when Photosynthesis depends on this spec. My understanding is that Photosynthesis is going through the IETF track, and it's not clear that you would be permitted to take a dep on the specs in this repo.

The risks of accepting this feature too soon are that:
 - complexity in the spec, which cannot be removed without a v2
 - complexity in the definition of “discoverability”; having a clear definition for log properties has been key for new transparency ecosystem adoption. We need to work out what discoverability means in this new world where logs can be pruned. A well run ecosystem should still have all claims being discoverable via other parties (e.g. mirrors), but we lose the ability to succinctly reason and communicate about this
 - other ecosystems start trying to rely on this feature as unknowing pioneers, without the expertise and experience that CT has of running at scale for a long time. Having an experienced ecosystem being the pioneer for rolling out this feature will help flush out the policy concerns (e.g. sufficient mirroring, agreement and rollout of the pruning index) that successful application of pruning will require
 - it encourages the idea of “lifetime logs”, which is an anti-pattern; we’ve historically encouraged logs to have a pre-determined shelf life and encourage their rotation. This means that all joints for doing this stay well lubricated. With lifetime logs, it’s possible that if a log fails due to a cosmic bit flip, the process of updating all participants with the new log key will be painful

Many of these issues are not objections to this proposal per se, but are about the impact on transparency adoption. Should Photosynthesis get to a point where you are ready to roll out an implementation of this with a policy-supported pruning horizon strategy, we recommend that this PR is merged and CT pioneers the exemplar for how this should be used.

Issue #133, opened 2025-05-30
New spec: tlog-mirror and tlog-ext-cosignature


davidben:
This is a proposal for a pair of new specs. The two drafts are here:

https://github.com/davidben/C2SP/blob/tlog-mirror/tlog-mirror.md
~~https://github.com/davidben/C2SP/blob/tlog-mirror/tlog-ext-cosignature.md~~ [see discussion below for ext-cosignature]

Proposed short names: tlog-mirror ~~and tlog-ext-cosignature~~
Proposed maintainers: @davidben, @AlCutter, @lukevalenta

/cc @C2SP/stewards

I'm not as plugged into the tlog ecosystem, and made a lot of guesses as to what would fit in best. Let me know if I didn't get it quite right! Some long-form thoughts below:

**Broader discussion**

The idea here is to have define a "mirror" role that's like a witness, but also promises to serve the contents of the checkpoints it has seen. This could be useful when you're not sure if the origin log can be trusted to faithfully everything (a log could just start 404-ing tiles at any point), or if it doesn't have the bandwidth.

This could just be rsync, but once we do this, it makes sense to sign the checkpoints, so clients can incorporate mirrors into their various quorum policies.  In theory, we could reuse tlog-cosignature, but the signature's semantics are specific to witnesses. Also, while key reuse is Bad(TM), it makes me a little uncomfortable to not capture the particular cosignature role in the payload *somewhere*. That suggests a new signature type.

We could make a bespoke mirror cosignature, but we'll have to now burn a signed note sigalg for each role, and you could imagine other roles that assert random properties about the log contents. So instead the idea is we make an "extended cosigner", identified by a cosigner origin. The semantics are "I promise this checkpoint is on my append-only branch" AND "I promise <whatever special things I'm defined to promise>", using the origin as a handle for both the signer and implicitly the signer's role. A given extended cosigner could even promise to validate such-and-such checkpoint extension, so this gives a path forward for those! [Edit: After discussion below, this is now just v2 cosignature and there are no more "extended cosigners".]

(Aside: if we had a time machine, the log signatures themselves could have been these too. We could have just said the log origin defines a cosigner whose promise is "this is on my append-only branch because I am authoritative for the log contents in the first place".)

With that, the main "clever" idea in tlog-mirror is this two-stage update process. My thinking was:
 
* It seems bad to commit unverified tiles. Then we have to worry about poisoning storage because a mirror got a tile from an untrusted source, like another mirror, or a USB drive on the street.

[...]

PR #140, opened 2025-06-24
tlog-mirror: first draft of the spec

davidben:
As part of this, this updates tlog-cosignature to cover cosigners with more expanded guarantees.

This incorporates some feedback since https://github.com/C2SP/C2SP/issues/133, both on the issue and out-of-band:

- tlog-ext-cosignature is replaced with cosignature/v2

- The background pull-based update process is replaced with push-based API

One piece of feedback I have not yet incorporated is that maybe cosignature/v2 is unnecessary and instead we should instead just expand the semantics of the existing cosignature/v1 signature (somewhat of a sleight of hand, but should be compatible?), and live with the cosigner name not being committed by having a MUST-level requirement against key reuse. Happy to do either, and I figure we can use the PR to discuss this.

Another thing to call out is I'm currently citing draft-davidben-tls-merkle-tree-certs-05 for some "subtree" machinery. It's a stable reference, but I'm not sure if that's better respecified for now.

### DIFF
--- a/tlog-cosignature.md
+++ b/tlog-cosignature.md
@@ -1,8 +1,10 @@
 # Transparency Log Cosignatures
 
-A cosignature is a statement by a transparency log [witness][] that it verified
-the consistency of a [checkpoint][]. Log clients can verify a quorum of
-cosignatures to prevent split-view attacks before trusting an inclusion proof.
+An cosignature is a statement by some party in a transparency log ecosystem,
+such as a [witness][], that it verified the consistency of a [checkpoint][],
+along with other properties specified by that cosigner. Log clients can verify a
+quorum of cosignatures to provide split-view attacks and also obtain assurance
+of other properties of some log entry.
 
 ```
 example.com/behind-the-sofa
@@ -33,28 +35,66 @@ document are to be interpreted as described in [BCP 14][] [RFC 2119][] [RFC
 [RFC 8174]: https://www.rfc-editor.org/rfc/rfc8174.html
 [RFC 8032]: https://www.rfc-editor.org/rfc/rfc8032.html
 
-## Format
+## Introduction
 
-Concretely, a cosignature is a [note signature][] applied to a [checkpoint][].
+A **cosigner** is a generalized [witness][]. For each supported log, it follows
+some append-only branch of the log, either by being the log operator or by
+checking consistency proofs. Along that branch, cosigners will generate
+**cosignatures** of checkpoints. A cosignature asserts that some checkpoint is
+part of the append-only branch and, optionally, provides additionally
+cosigner-specific assertions about the checkpoint.
+
+Cosigners have an **name** and a public key. The name is a unique
+identifier for the cosigner. The name MUST be non-empty, and it SHOULD be
+a schema-less URL containing neither Unicode spaces nor plus (U+002B), such
+as `example.com/mirror42`. This is only a recommendation to avoid collisions,
+and clients MUST NOT assume that the name is following this format or that
+the URL corresponds to a reachable endpoint.
+
+For example, a [witness][] is a cosigner that provides no guarantees beyond the
+append-only assertion. A [mirror][] additionally asserts that the checkpoint's
+contents are available from its monitoring interface. Other documents MAY define
+cosigner roles that provide other assertions, e.g. checking some [checkpoint][]
+extension, or some property of the entries.
+
+When an cosigner signs checkpoints, it is held responsible *both* for upholding
+the append-only property *and* for meeting its defined guarantees for all
+entries in any checkpoints that it signed.
+
+A single cosigner, with a single cosigner name and public key, MAY generate
+cosignatures for checkpoints from multiple logs. The signed message, defined
+below, includes both the cosigner name and log origin.
+
+An cosigner's name identifies the cosigner and thus the assertions provided. If
+a single operator performs multiple cosigner roles in an ecosystem, each role
+MUST use a distinct cosigner name and SHOULD use a distinct key.
+
+## Note Signatures
+
+Concretely, a cosignature is a [note signature][] applied to a
+[checkpoint][]. The note signature's key name MUST be the cosigner's name.
 
 Per the signed note format, a note signature line is
 
-    — <key name> base64(32-bit key ID || signature)
+    — <name> base64(32-bit key ID || signature)
 
-The key name SHOULD be a schema-less URL that identifies the witness. Like the
-checkpoint origin line, this is for disambiguation, and MAY match a publicly
-reachable endpoint or not.
-
-The key ID MUST be
-
-    SHA-256(<name> || "\n" || 0x04 || 32-byte Ed25519 witness public key)[:4]
-
-Clients are configured with tuples of (witness name, public key, supported
-cosignature version) and based on that they can compute the expected name and
+The key ID is determined from the version of cosignatures being used. Clients
+are configured with tuples of (cosigner name, public key, supported cosignature
+version) and based on that they can compute the expected name and
 key ID, and ignore any signature lines that don't match the name and key ID.
 
-Future cosignature formats MAY reuse the same witness public key with a
+Future cosignature formats MAY reuse the same cosigner public key with a
 different key ID algorithm byte (and a different signed message header line).
+
+## V2 Cosignatures
+
+v2 cosignatures are the current cosignature version.
+
+### Format
+
+The key ID for a v2 cosignature MUST be
+
+    SHA-256(<name> || "\n" || 0x06 || 32-byte Ed25519 cosigner public key)[:4]
 
 The signature MUST be a 72-byte `timestamped_signature` structure.
 
@@ -69,7 +109,68 @@ the UNIX epoch (January 1, 1970 00:00 UTC).
 "signature" is an Ed25519 ([RFC 8032][]) signature from the witness public key
 over the message defined in the next section.
 
-## Signed message
+### Signed message
+
+The signed message MUST be three newline (U+000A) terminated lines (one header
+line, one timestamp line, and one name line) followed by the whole note body of
+the cosigned checkpoint (including the final newline, but not including any
+signature lines).
+
+The header line MUST be the fixed string `cosignature/v2`, and provides domain
+separation.
+
+The name line MUST consistent of the cosigner name.
+
+The timestamp line MUST consist of the string `time`, a single space (0x20), and
+the number of seconds since the UNIX epoch encoded as an ASCII decimal with no
+leading zeroes. This value MUST match the `timestamped_signature.timestamp`.
+
+    cosignature/v2
+    witness.example.com/w1
+    time 1679315147
+    example.com/behind-the-sofa
+    20852163
+    CsUYapGGPo4dkMgIAUqom/Xajj7h2fB2MPA3j2jxq2I=
+
+Semantically, a v2 cosignature is a statement that, as of the specified time,
+the specified checkpoint is one of the largest size which:
+
+* has a tree hash which is consistent with all other checkpoints signed by the named cosigner
+* satisfies all other properties asserted by the named cosigner
+
+Extension lines MAY be included in the checkpoint by the log, and if present
+MUST be included in the cosigned message. However, no semantic statement is made
+about any extension line, unless the cosigner is defined to make them.
+
+## V1 Cosignatures
+
+v1 cosignatures are an older, witness-only cosignature version. This version may
+only be used by witnesses. It does not carry any additional cosigner guarantees
+and additionally does not bind the cosigner name into the signed message.
+
+A witness MAY generate v1 cosignatures using the same key used for v2
+cosignatures, however, two distinct witnesses MUST NOT use the same key.
+
+### Format
+
+The key ID for a v1 cosignature MUST be
+
+    SHA-256(<witness name> || "\n" || 0x04 || 32-byte Ed25519 witness public key)[:4]
+
+The signature MUST be a 72-byte `timestamped_signature` structure.
+
+    struct timestamped_signature {
+        u64 timestamp;
+        u8 signature[64];
+    }
+
+"timestamp" is the time at which the cosignature was generated, as seconds since
+the UNIX epoch (January 1, 1970 00:00 UTC).
+
+"signature" is an Ed25519 ([RFC 8032][]) signature from the witness public key
+over the message defined in the next section.
+
+### Signed message
 
 The signed message MUST be two newline (U+000A) terminated lines (one header
 line and one timestamp line) followed by the whole note body of the cosigned
@@ -100,5 +201,6 @@ is made about any extension lines, and consensus between witnesses on the
 extension lines SHALL NOT be assumed.
 
 [note signature]: https://c2sp.org/signed-note
+[mirror]: https://c2sp.org/tlog-mirror
 [checkpoint]: https://c2sp.org/tlog-checkpoint
 [witness]: https://c2sp.org/tlog-witness
--- a/tlog-mirror.md
+++ b/tlog-mirror.md
@@ -1,0 +1,297 @@
+# Transparency Log Mirrors
+
+This document describes how to mirror a transparency log, and how to obtain
+signatures asserting that a mirror has done so.
+
+[cosigner]: https://c2sp.org/tlog-cosignature
+[cosignature]: https://c2sp.org/tlog-cosignature
+[checkpoint]: https://c2sp.org/tlog-checkpoint
+[tiled transparency log]: https://c2sp.org/tlog-tiles
+[witness]: https://c2sp.org/tlog-witness
+[percent-encoded]: https://www.rfc-editor.org/rfc/rfc3986.html#section-2.1
+[subtree]: https://www.ietf.org/archive/id/draft-davidben-tls-merkle-tree-certs-05.html#name-subtrees
+[subtree consistency proof]: https://www.ietf.org/archive/id/draft-davidben-tls-merkle-tree-certs-05.html#name-subtree-consistency-proofs
+
+## Conventions used in this document
+
+The base64 encoding used throughout is the standard Base 64 encoding specified
+in [RFC 4648][], Section 4.
+
+`U+` followed by four hexadecimal characters denotes a Unicode codepoint, to be
+encoded in UTF-8. `0x` followed by two hexadecimal characters denotes a byte
+value in the 0-255 range.
+
+`[start, end)`, where `start <= end`, denotes the half-open interval containing
+integers `x` such that `start <= x < end`.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in [BCP 14][] [RFC 2119][] [RFC
+8174][] when, and only when, they appear in all capitals, as shown here.
+
+[RFC 4648]: https://www.rfc-editor.org/rfc/rfc4648.html
+[BCP 14]: https://www.rfc-editor.org/info/bcp14
+[RFC 2119]: https://www.rfc-editor.org/rfc/rfc2119.html
+[RFC 8174]: https://www.rfc-editor.org/rfc/rfc8174.html
+[RFC 6962]: https://www.rfc-editor.org/rfc/rfc6962.html
+
+## Introduction
+
+A mirror is a [cosigner][] that stores a copy of a log and, when signing a
+[checkpoint][], provides the additional guarantee that the mirror has
+durably logged the contents of the checkpoint and has made it accessible.
+
+A mirror is defined by a name, a public key, and by two URL prefixes:
+the *submission prefix* for write APIs and the *monitoring prefix* for read
+APIs. A mirror MAY use the same value for both the *submission prefix* and the
+*monitoring prefix*.
+
+For each supported origin log, the mirror is configured with:
+
+* The log's public key
+* The log's URL prefix
+* An optional list of monitoring prefixes for other mirrors for the log
+
+The mirror maintains a copy of each origin log and serves it publicly via the
+[tiled transparency log][] interface. It uses a URL prefix of
+`<monitoring prefix>/<encoded origin>`, where `encoded origin` is the log's
+origin, [percent-encoded][]. The checkpoint served from this prefix MUST include
+a [cosignature][] from the mirror.
+
+## Updating a Mirror
+
+The mirror update process is designed to be safely interruptible, while avoding
+large atomic operations. For each origin log, a mirror maintains the following:
+
+* A copy of the log. The latest checkpoint of this log copy is known as the
+  *mirror checkpoint*.
+
+* A *pending checkpoint*, which is at or ahead of the mirror checkpoint. If
+  ahead of the mirror checkpoint, the pending checkpoint describes entries that
+  have yet to be incorporated into the mirror.
+
+* A list of *pending entries* that have yet to be incorporated into the mirror
+  checkpoint. The mirror's *next entry* is the log index of the first entry that
+  is not in either the log or pending entry list.
+
+The update process ensures that all current and past pending checkpoints are
+consistent, and all pending entries are contained in the current pending
+checkpoint. Thus mirrors MAY commit pending entries to the log, serving them as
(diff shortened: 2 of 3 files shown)


tlog-cosignature.md
```diff
@@ -1,8 +1,10 @@
 # Transparency Log Cosignatures
 
-A cosignature is a statement by a transparency log [witness][] that it verified
-the consistency of a [checkpoint][]. Log clients can verify a quorum of
-cosignatures to prevent split-view attacks before trusting an inclusion proof.
+An cosignature is a statement by some party in a transparency log ecosystem,
```
lukevalenta:
```suggestion
A cosignature is a statement by some party in a transparency log ecosystem,
```

tlog-cosignature.md
```diff
@@ -33,29 +35,128 @@ document are to be interpreted as described in [BCP 14][] [RFC 2119][] [RFC
...
-Concretely, a cosignature is a [note signature][] applied to a [checkpoint][].
+A **cosigner** is a generalized [witness][]. For each supported log, it follows
+some append-only branch of the log, either by being the log operator or by
+checking consistency proofs. Along that branch, cosigners will generate
+**cosignatures** of checkpoints. A cosignature asserts that some checkpoint is
+part of the append-only branch and, optionally, provides additionally
+cosigner-specific assertions about the checkpoint.
 
-Per the signed note format, a note signature line is
+Cosigners have an **name** and a public key. The name is a unique
```
lukevalenta:
```suggestion
Cosigners have a **name** and a public key. The name is a unique
```

tlog-cosignature.md
```diff
@@ -33,29 +35,128 @@ document are to be interpreted as described in [BCP 14][] [RFC 2119][] [RFC
...
+and clients MUST NOT assume that the name is following this format or that
+the URL corresponds to a reachable endpoint.
+
+For example, a [witness][] is a cosigner that provides no guarantees beyond the
+append-only assertion. A [mirror][] additionally asserts that the checkpoint's
+contents are available from its monitoring interface. Other documents MAY define
+cosigner roles that provide other assertions, e.g. checking some [checkpoint][]
+extension, or some property of the entries.
+
+When an cosigner signs checkpoints, it is held responsible *both* for upholding
```
lukevalenta:
```suggestion
When a cosigner signs checkpoints, it is held responsible *both* for upholding
```

tlog-cosignature.md
```diff
@@ -33,29 +35,128 @@ document are to be interpreted as described in [BCP 14][] [RFC 2119][] [RFC
...
 
-    — <key name> base64(32-bit key ID || signature)
+A single cosigner, with a single cosigner name and public key, MAY generate
+cosignatures for checkpoints from multiple logs. The signed message, defined
+below, includes both the cosigner name and log origin.
 
-The key name SHOULD be a schema-less URL that identifies the witness. Like the
-checkpoint origin line, this is for disambiguation, and MAY match a publicly
-reachable endpoint or not.
+An cosigner's name identifies the cosigner and thus the assertions provided. If
```
lukevalenta:
```suggestion
A cosigner's name identifies the cosigner and thus the assertions provided. If
```

tlog-cosignature.md
```diff
@@ -33,29 +35,128 @@ document are to be interpreted as described in [BCP 14][] [RFC 2119][] [RFC
...
+"timestamp" is the time at which the cosignature was generated, as seconds since
+the UNIX epoch (January 1, 1970 00:00 UTC).
+
+"signature" is an Ed25519 ([RFC 8032][]) signature from the witness public key
+over the message defined in the next section.
+
+### Signed message
+
+The signed message MUST be three newline (U+000A) terminated lines (one header
+line, one timestamp line, and one name line) followed by the whole note body of
```
lukevalenta:
```suggestion
line, one name line, and one timestamp line) followed by the whole note body of
```

tlog-cosignature.md
```diff
@@ -33,29 +35,128 @@ document are to be interpreted as described in [BCP 14][] [RFC 2119][] [RFC
...
+the cosigned checkpoint (including the final newline, but not including any
+signature lines).
+
+The header line MUST be the fixed string `cosignature/v2`, and provides domain
+separation.
+
+The name line MUST consistent of the cosigner name.
```
lukevalenta:
```suggestion
The name line MUST be the cosigner name.
```

tlog-cosignature.md
```diff
@@ -33,29 +35,128 @@ document are to be interpreted as described in [BCP 14][] [RFC 2119][] [RFC
...
+
+    cosignature/v2
+    witness.example.com/w1
+    time 1679315147
+    example.com/behind-the-sofa
+    20852163
+    CsUYapGGPo4dkMgIAUqom/Xajj7h2fB2MPA3j2jxq2I=
+
+Semantically, a v2 cosignature is a statement that, as of the specified time,
+the specified checkpoint is one of the largest size which:
```
lukevalenta:
```suggestion
the specified checkpoint is of the largest size which:
```

davidben:
Oh another detail is this will want to be slightly updated for https://github.com/C2SP/C2SP/pull/138. I figure whichever PR lands second can pick that up.

tlog-mirror.md
```diff
@@ -0,0 +1,297 @@
...
+[BCP 14]: https://www.rfc-editor.org/info/bcp14
+[RFC 2119]: https://www.rfc-editor.org/rfc/rfc2119.html
+[RFC 8174]: https://www.rfc-editor.org/rfc/rfc8174.html
+[RFC 6962]: https://www.rfc-editor.org/rfc/rfc6962.html
+
+## Introduction
+
+A mirror is a [cosigner][] that stores a copy of a log and, when signing a
+[checkpoint][], provides the additional guarantee that the mirror has
+durably logged the contents of the checkpoint and has made it accessible.
```
lukevalenta:
```suggestion
durably logged and made accessible the contents of the checkpoint.
```

lukevalenta: Not sure I love my suggested phrasing, but previously this could be read that the mirror was only making the checkpoint accessible but not the log contents.

tlog-mirror.md
```diff
@@ -0,0 +1,297 @@
...
+
+The mirror cosigner MUST NOT sign the checkpoint in this process. It MAY respond
+with witness cosignatures if the mirror operator wishes to additionally provide
+a separate witness service using its pending checkpoint. If so, this witness
+service MUST be a distinct cosigner from the mirror cosigner, with a distinct
+name. The mirror's signature is computed later, as described below.
+
+### add-entries
+
+The mirror implements a `add-entries` endpoint to upload entries for a supported
```
lukevalenta:
```suggestion
The mirror implements an `add-entries` endpoint to upload entries for a supported
```

tlog-mirror.md
```diff
@@ -0,0 +1,297 @@
...
+`num_packages = (rounded_end - rounded_start) / 256` packages. Entry
+package `i`, for `0 <= i < num_packages`, MUST be computed from the interval
+`[start, end)` where:
+
+    start = upload_start, rounded_start + i * 256
+    end = upload_end, rounded_start + (i + 1) * 256
+
+The package MUST contain the following values, concatenated.
+
+* The log entries in `[start, end)`, each with a big-endian, 16-bit length
```
lukevalenta: Maybe we should tweak the tlog-tiles phrasing to use 16-bit instead of uint16 for consistency's sake: https://github.com/C2SP/C2SP/blob/main/tlog-tiles.md?plain=1#L136.

tlog-mirror.md
```diff
@@ -0,0 +1,297 @@
...
+  ahead of the mirror checkpoint, the pending checkpoint describes entries that
+  have yet to be incorporated into the mirror.
+
+* A list of *pending entries* that have yet to be incorporated into the mirror
+  checkpoint. The mirror's *next entry* is the log index of the first entry that
+  is not in either the log or pending entry list.
+
+The update process ensures that all current and past pending checkpoints are
+consistent, and all pending entries are contained in the current pending
+checkpoint. Thus mirrors MAY commit pending entries to the log, serving them as
```
lukevalenta, line 79: I'd missed that the mirror could go ahead and serve pending entry bundles before the corresponding mirror checkpoint is updated. Makes sense!

tlog-mirror.md
```diff
@@ -0,0 +1,297 @@
...
+#### Request Body
+
+The request body MUST have `Content-Type` of `application/octet-stream` and
+contain the following values, concatenated.
+
+* 8 bytes, encoding a big-endian, unsigned 64-bit integer: `upload_start`
+* 8 bytes, encoding a big-endian, unsigned 64-bit integer: `upload_end`
+* 2 bytes, encoding a big-endian, unsigned 16-bit integer: `ticket_size`
+* `ticket_size` bytes, containing an opaque `ticket` value, described below
+* A sequence of *entry packages*, described below
```
lukevalenta, line 136: Entry packages vs entry bundles are confusing (I thought they were the same thing on my first read). Maybe we can just call these chunks instead?

```suggestion
* A sequence of *chunks*, described below
```

tlog-mirror.md
```diff
@@ -0,0 +1,297 @@
...
+
+More precisely, if `upload_start` is equal to `upload_end`, there are no
+packages. Otherwise, let `rounded_start` be `upload_start` rounded down to a
+multiple of 256, and let `rounded_end` be `upload_end` rounded up to a
+multiple of 256. The request MUST contain
+`num_packages = (rounded_end - rounded_start) / 256` packages. Entry
+package `i`, for `0 <= i < num_packages`, MUST be computed from the interval
+`[start, end)` where:
+
+    start = upload_start, rounded_start + i * 256
```
lukevalenta:
```suggestion
    start = max(upload_start, rounded_start + i * 256)
```

tlog-mirror.md
```diff
@@ -0,0 +1,297 @@
...
+    end = upload_end, rounded_start + (i + 1) * 256
```
lukevalenta:
```suggestion
    end = min(upload_end, rounded_start + (i + 1) * 256)
```

tlog-mirror.md
```diff
@@ -0,0 +1,297 @@
...
+When committing authenticated entries to the pending entries list, it is
+possible that, due to a concurrent instance of `add-entries`, some entries have
+already been added to the pending entries list or the mirror. Mirrors MUST
+correctly handle this case and continue operating correctly. This may require
+synchronization of individual log resources. In doing so, the mirror MAY assume
+that the two copies of the entries are identical. They will both be proven
+consistent with the pending checkpoints.
+
+When updating the mirror checkpoint to `upload_end`, it is possible that some
+concurrent of `add-entries` has already updated the checkpoint to `upload_end`
```
lukevalenta:
```suggestion
concurrent instance of `add-entries` has already updated the mirror checkpoint to `upload_end`
```

tlog-mirror.md
```diff
@@ -0,0 +1,297 @@
...
+or past it. In this case, the mirror MUST NOT rewind the checkpoint and MUST
+instead skip the update.
+
+Entry packages are divided at multiples of 256 to align with the entry bundle
+representation in the tiled log interface. It is expected that most
+implementations will compute exactly one entry bundle from each entry package
+and commit it directly to log storage when the package is authenticated.
+
+A mirror that uses a different representation MAY buffer entry packets and defer
```
lukevalenta:
```suggestion
A mirror that uses a different representation MAY buffer entry packages and defer
```

tlog-mirror.md
```diff
@@ -0,0 +1,297 @@
...
+size 512, it might commit entry packages two at a time.
+
+A mirror MAY process an entry package without waiting for the previous entry
+package to be durably committed to storage. However, the mirror MUST NOT sign
+or update its mirror checkpoint until all entries are durably committed. If the
+mirror commits entries out of order, it MUST correctly compute the next entry to
+be the *first* missing entry, even if some subsequent entries have been
+committed. Mirror clients will then reupload the subsequent entries.
+
+Mirrors MAY additionally implement other update processes, provided it continues
```
lukevalenta:
```suggestion
A mirror MAY additionally implement other update processes, provided it continues
```

lukevalenta (commented):
From a first pass this is looking good!

tlog-mirror.md
davidben: Yeah, I really wasn't sure how to write this. What I _want_ to write is you just go ahead and make your entry bundles and write them to the log store and move on with life.

But it would also be plausible for the log to be stored differently, in which case maybe the representations are separate. And so this was trying to be the more general phrasing, but it was a little weird because it read like a staging area.

tlog-mirror.md
davidben: Ah hmm. I could also match tlog-tiles to uint16, but "uint16" isn't defined anywhere. Then again, everyone's going to understand what that means. 🤷

tlog-mirror.md
davidben: Hmm, maybe. I did like the parallel because they're basically entry bundles, but authenticated (and the first one might skip some entries you already have), but it is confusing. "Chunk" is a very generic though.

Dunno. No strong feelings either way.

tlog-mirror.md
lukevalenta: Yeah, probably just a carryover from the first version of these specs being written for a Golang library.

tlog-mirror.md
```diff
@@ -0,0 +1,297 @@
...
+The request body uploads the log entries whose indices are in
+`[upload_start, upload_end)`. Entries are grouped into bounded-size entry
+packages. Each package has a [subtree consistency proof][] that allows the
+mirror to verify and commit the entries in the package without buffering the
+entire request body.
+
+Each entry package is determined by a half-open interval `[start, end)` of log
+indices. The request MUST contain entry packages whose intervals' disjoint
+union, in order, is the interval `[upload_start, upload_end)`. The overall
+interval MUST be divided into packages at multiples of 256.
```
lukevalenta:
```suggestion
interval MUST be divided into packages at multiples of 256, to align with the corresponding entry bundles.
```

tlog-mirror.md
lukevalenta: Fair! I'm fine to keep "entry packages" for now and just added some suggested text about trying to align entry packages with entry bundles.

tlog-mirror.md
lukevalenta: The text seems clear enough now!

lukevalenta (approved):

rgdd:
Do I get the gist of the changes you're suggesting for cosignature/v2 if I summarize it as follows:

* The timestamped append-only claim/promise is the same as in cosignature/v1
* But the witness *name* is included in the signed message, which means the name cannot be changed without also controlling the public key.

You're suggesting to convey *additional* claims/promises via the name, like "I mirrored all entries up until this checkpoint" when you see the name "exampe.org/mirror". These additional claims/promises are ecosystem specific and intentionally not defined by tlog-cosignature.

tlog-mirror.md
```diff
@@ -0,0 +1,297 @@
...
+The mirror implements a [witness][]'s `add-checkpoint` endpoint to update its
+pending checkpoint for a log:
+
+    POST <submission prefix>/<encoded origin>/add-checkpoint
+
+`encoded origin` is the log's origin, [percent-encoded][].
+
+The request is handled identically to that of a witness, updating the pending
+checkpoint (but not the mirror checkpoint), with the exception that it does not
+need to generate and respond with any cosignatures. The mirror MAY handle the
```
rgdd, line 107: Is there a good reason for not always returning a cosignature here without the the mirror promise, and so just implementing add-checkpoint as described in tlog-witness?

rgdd:
Minor comment / question - why is the percent encoded origin line needed? Can't this just be part of the submission/monitor prefix if the operator wants to include origin line in percent encoding?

tlog-mirror.md
davidben: Ah yeah, the section after that says:

> Entry packages are divided at multiples of 256 to align with the entry bundle
> representation in the tiled log interface. It is expected that most
> implementations will compute exactly one entry bundle from each entry package
> and commit it directly to log storage when the package is authenticated.

Things got long so I started banishing some of the "this is how you're meant to implement it" talk to another section.

tlog-mirror.md
davidben: Just that the mirror would need to then allocate two different cosigner names, and so I figure it's better to make it optional whether the mirror operator wants to be held responsible for two different services at once.

davidben:
> Do I get the gist of the changes you're suggesting for cosignature/v2 if I summarize it as follows:
> 
> * The timestamped append-only claim/promise is the same as in cosignature/v1
> * But the witness _name_ is included in the signed message, which means the name cannot be changed without also controlling the public key.
> 
> You're suggesting to convey _additional_ claims/promises via the name, like "I mirrored all entries up until this checkpoint" when you see the name "exampe.org/mirror". These additional claims/promises are ecosystem specific and intentionally not defined by tlog-cosignature.

Yup, exactly! Having one signature construction for witnesses and mirrors means that the quorum-checking software doesn't need to have different slots for the two, or be updated every time we have a new role. The knowledge about the role is instead lifted up to whoever picked cosigners and set the quorum policy. ("I need any 2 of these, any 3 of these, etc.")

Having it extensible for new roles is handy for a few reasons. Spitballing a bit:
* Perhaps you all want to define a witness that *does* check some checkpoint extension line! I dunno, maybe you put a [vindex hash](https://github.com/transparency-dev/incubator/tree/main/vindex) in the checkpoint extension
* Maybe the cosigner checks that log entries are blue... or it's a [CA and promises to have validated everything](https://davidben.github.io/merkle-tree-certs/draft-davidben-tls-merkle-tree-certs.html#name-certification-authority-cos)
* Maybe some ecosystem wants a variation on tlog-mirror where the promise isn't "this is available" but "this will be available after a mirror-specific MMD because I'm behind four layers of caches".

(Aside: if I had a time machine, I would go back and propose that log signatures and witnesses just be different instances of the same cosigner construction.)

The overall observation is that if you supported the cosigner, you presumably already know what that cosigner is about. If you didn't support it, you don't really care if it's a witness or mirror. You were going to ignore the signature anyway. So we can make everything implicit from the cosigner's identity.

The name commitment isn't _strictly_ necessary. If we take it out, we instead have to require that you absolutely MUST NOT share keys between cosigners. You really SHOULD NOT do that anyway, but people end up reusing keys anyway. E.g. if the mirror operator decides to provide a witnessing service per your other comment, I bet folks will do that. Thus the question at the top: would you all prefer we go ahead and do `cosignature/v2`, because committing names is good hygiene, or would you all prefer we just "transparently" upgrade the `cosignature/v1` promise and just prohibit key reuse.

> Minor comment / question - why is the percent encoded origin line needed? Can't this just be part of the submission/monitor prefix if the operator wants to include origin line in percent encoding?

Just so that one mirror can mirror multiple logs, just like one witness can witness multiple logs. Although, as I write this I realize that the origin on the submission prefix is kinda weird because the `add-checkpoint` payload includes the origin in the request body already. So I probably should take it out of the `add-entries` URL and put it into the request body...

tlog-mirror.md
davidben: Looks like lots of C2SP specs all say uintN. I've switched to that just because it's shorter and honestly everyone will know what that means. Meh. :-)

tlog-mirror.md
lukevalenta: Having both 'entry packages' and 'entry bundles' in the same sentence is what clarified things for me, so if we can find a way to work that in earlier in the doc I think it could be useful. No strong feelings though.

davidben: Ah makes sense. SGTM.

tlog-mirror.md
```diff
@@ -0,0 +1,297 @@
...
+
+    POST <submission prefix>/add-entries
+
+#### Request Body
+
+The request body MUST have `Content-Type` of `application/octet-stream` and
+contain the following values, concatenated.
+
+* 2 bytes, encoding a big-endian uint16: `log_origin_size`
+* `log_origin_size` bytes, containing the log origin
```
lukevalenta: We use the term `log_origin` below, so maybe just define it explicitly here in case someone is doing a text search through the document.
```suggestion
* `log_origin_size` bytes, containing the log origin: `log_origin`
```

davidben: Whoops, yes.

tlog-cosignature.md
```diff
@@ -1,8 +1,10 @@
 # Transparency Log Cosignatures
 
-A cosignature is a statement by a transparency log [witness][] that it verified
-the consistency of a [checkpoint][]. Log clients can verify a quorum of
-cosignatures to prevent split-view attacks before trusting an inclusion proof.
+A cosignature is a statement by some party in a transparency log ecosystem,
+such as a [witness][], that it verified the consistency of a [checkpoint][],
+along with other properties specified by that cosigner. Log clients can verify a
+quorum of cosignatures to provide split-view attacks and also obtain assurance
```
eliasrudberg: Here is looks like the phrase "prevent split-view attacks" has been changed to "provide split-view attacks". Is that a mistake, it should still be "prevent" and not "provide" there?

davidben: Ahahaha, that's quite a typo. Fixed. I probably retyped that clause at some point while I was moving text around and accidentally mangled it. Thanks!

rgdd:
@davidben, @lukevalenta: please take a look at <https://github.com/C2SP/C2SP/pull/146>. Basically just a maintainer's touch on cosignature/v2 in an attempt to help move it forward. But TL;DR: I like the direction it's headed. Thanks for working on this!

tlog-mirror.md
rgdd: Ack, and I suppose this makes even more sense if it would require management of two separate key pairs.

tlog-mirror.md
```diff
@@ -0,0 +1,297 @@
...
+* `num_hashes` [subtree consistency proof][] hash values
+
+The subtree consistency proof is computed from the [subtree][] defined by
+`[rounded_start + i * 256, end)`, and the log checkpoint with tree size
+`upload_end`.
+
+TODO: This is currently citing an individual IETF draft for subtrees, though it
+is versioned and immutable. Should we, for now, copy and paste that text
+somewhere here? (Subtrees are also slightly more general than needed here. Every
+subtree we consider is directly contained in the target tree size.)
```
bwesterb, line 182: The simpler definition is as follows. Let `[start, end)` be a subtree of D_n. We further assume the subtree is full or a suffix. Then define `TRUNC_PATH(start, end, D)` as

```
TRUNC_PATH(0, n, D_n) = {}
TRUNC_PATH(start, end, D_n) = TRUNC_PATH(start, end, D[0:k]) : MTH(D[k:n]) for end ≤ k; and
TRUNC_PATH(start, end, D_n) = TRUNC_PATH(start - k, end - k, D[k:n]) : MTH(D[0:k]) for start ≥ k
```

where as usual `k` is the largest power of two below `n`.

**Claim.** The case `start < k < end` does not occur.
*Proof.* If the subtree is full, then start must be a multiple of end-start which is a power of two. If start=0, then obviously k≥end. If start>0, then the next power of two is 2*start≥end, so again k=start or k≥end.
If the subtree is not full, then it must be a suffix. If it's the entire tree, then we're in the case `TRUNC_PATH(0, n, D_n)`. If it's not the entire tree, then by virtue of being a subtree start is a non-zero multiple of a power of two that's larger or equal to end-start. It can't be equal, as we'd have a subtree. Thus the next power of two is strictly larger than end. Hence k≤start.

With this in hand, we straightforwardly see `SUBTREE_SUBPROOF(start, end, D_n, true) = TRUNC_PATH(start, end, D_n)`.

We can also just use the existing inclusion proofs and truncate them. Indeed: for any  start ≤ m < end, that  the regular inclusion proof `PATH(m, D_n) = PATH(m-start, D[start:end]) | TRUNC_PATH(start, end, D_n)`.

davidben: Ah thanks! Yeah, I guess the question is how many different kinds of animals do we want to define, and which existing definitions are acceptable to cite. @FiloSottile thoughts?

FiloSottile: At this stage I think it's fine to cite an I-D. A pre-v1.0.0 C2SP spec is akin to an I-D itself, and we can wait until we see how things settle to decide how to draw the arrows or what to duplicate in the final versions.

davidben: Sounds good! Yeah, let's leave it be then.

PR #146, opened 2025-06-30
tlog-cosignature: Allow additional statements

rgdd:
This PR puts a maintainer's touch on https://github.com/C2SP/C2SP/pull/140 by targeting continued use of cosignature/v1 rather than going up to cosignature/v2. This does not mean we did not like the suggested cosignature/v2 changes (witness name is tied to the cosignature and can be used as a signing context), but rather that we want to be sure there aren't other things we also want in a cosignature/v2. One way to view this is: while it's a painpoint to manage multiple keys for different cosigning contexts, it's also a painpoint to manage multiple different cosignature versions. So we want to avoid creating a cosignature/v2 prematurely and start by making a backwards compatible update to the text in cosignature/v1.

Blocking merge:

* Input from @davidben, @lukevalenta: are you happy with this update / do you agree it is sufficient?
* Indications that all three maintainers think this is the right way to move forward, otherwise $discuss.

Nice to have before merge:

* A v1.0.0 tag, so that this can become v1.1.0.

Finally -- many thanks for the great C2SP contributions davidben/lukevalenta!

(If anyone else has input that's of course also welcome. I already gathered input from a few Sigsum contributors that basically gave thumbs up when I described the changes that were brewing here.)

### DIFF
--- a/tlog-cosignature.md
+++ b/tlog-cosignature.md
@@ -1,9 +1,9 @@
 # Transparency Log Cosignatures
 
-A cosignature is a statement by a transparency log [witness][] that it verified
+A cosignature is a statement by a *cosigner* that it verified
 the consistency of a [checkpoint][]. Log clients can verify a quorum of
 cosignatures to prevent split-view attacks before trusting an inclusion proof.
-A witness may make additional statements relating to a checkpoint.  Log clients
+A cosigner may make additional statements relating to a checkpoint.  Log clients
 that know about this can then be assured of additional cosigning properties.
 
 Below is an example of a checkpoint that contains a cosignature.
@@ -45,19 +45,19 @@ Per the signed note format, a note signature line is
 
     — <key name> base64(32-bit key ID || signature)
 
-The key name SHOULD be a schema-less URL that identifies the witness. Like the
+The key name SHOULD be a schema-less URL that identifies the cosigner. Like the
 checkpoint origin line, this is for disambiguation, and MAY match a publicly
 reachable endpoint or not.
 
 The key ID MUST be
 
-    SHA-256(<name> || "\n" || 0x04 || 32-byte Ed25519 witness public key)[:4]
+    SHA-256(<name> || "\n" || 0x04 || 32-byte Ed25519 cosigner public key)[:4]
 
-Clients are configured with tuples of (witness name, public key, supported
+Clients are configured with tuples of (cosigner name, public key, supported
 cosignature version) and based on that they can compute the expected name and
 key ID, and ignore any signature lines that don't match the name and key ID.
 
-Future cosignature formats MAY reuse the same witness public key with a
+Future cosignature formats MAY reuse the same cosigner public key with a
 different key ID algorithm byte (and a different signed message header line).
 
 The signature MUST be a 72-byte `timestamped_signature` structure.
@@ -70,7 +70,7 @@ The signature MUST be a 72-byte `timestamped_signature` structure.
 "timestamp" is the time at which the cosignature was generated, as seconds since
 the UNIX epoch (January 1, 1970 00:00 UTC).
 
-"signature" is an Ed25519 ([RFC 8032][]) signature from the witness public key
+"signature" is an Ed25519 ([RFC 8032][]) signature from the cosigner public key
 over the message defined in the next section.
 
 ## Signed message
@@ -93,17 +93,17 @@ leading zeroes. This value MUST match the `timestamped_signature.timestamp`.
     CsUYapGGPo4dkMgIAUqom/Xajj7h2fB2MPA3j2jxq2I=
 
 Semantically, a v1 cosignature is a statement that, as of the specified time,
-the consistent tree head with the largest size the witness has observed for the
+the consistent tree head with the largest size the cosigner has observed for the
 log identified by the origin line has the specified root hash.
 
 Extension lines MAY be included in the checkpoint by the log, and if present
 MUST be included in the cosigned message. However, it's important to understand
-that the witness is asserting observation of correct append-only operation of
+that the cosigner is asserting observation of correct append-only operation of
 the log based on the first three lines of the checkpoint; consensus between
-witnesses on the extension lines SHALL NOT be assumed, and no semantic statement
-is made about any extension lines unless the witness operator says otherwise.
+cosigners on the extension lines SHALL NOT be assumed, and no semantic statement
+is made about any extension lines unless the cosigner's operator says otherwise.
 
-A witness operator MAY make additional statements about a checkpoint.  These
+A cosigner MAY make additional statements about a checkpoint.  These
 additional statements need to be communicated out of band to those defining
 trust policies based on tuples of (public key, supported cosignature version).
 A given tuple MUST imply a single set of statements.  These statements MUST
--- a/tlog-cosignature.md
+++ b/tlog-cosignature.md
@@ -3,6 +3,10 @@
 A cosignature is a statement by a transparency log [witness][] that it verified
 the consistency of a [checkpoint][]. Log clients can verify a quorum of
 cosignatures to prevent split-view attacks before trusting an inclusion proof.
+A witness may make additional statements relating to a checkpoint.  Log clients
+that know about this can then be assured of additional cosigning properties.
+
+Below is an example of a checkpoint that contains a cosignature.
 
 ```
 example.com/behind-the-sofa
@@ -95,10 +99,20 @@ log identified by the origin line has the specified root hash.
 Extension lines MAY be included in the checkpoint by the log, and if present
 MUST be included in the cosigned message. However, it's important to understand
 that the witness is asserting observation of correct append-only operation of
-the log based on the first three lines of the checkpoint; no semantic statement
-is made about any extension lines, and consensus between witnesses on the
-extension lines SHALL NOT be assumed.
+the log based on the first three lines of the checkpoint; consensus between
+witnesses on the extension lines SHALL NOT be assumed, and no semantic statement
+is made about any extension lines unless the witness operator says otherwise.
+
+A witness operator MAY make additional statements about a checkpoint.  These
+additional statements need to be communicated out of band to those defining
+trust policies based on tuples of (public key, supported cosignature version).
+A given tuple MUST imply a single set of statements.  These statements MUST
+include the cosignature/v1 semantics, and MAY include other statements that
+are non-conflicting.  Examples of non-conflicting statements include "I also
+mirrored the log up until the checkpoint size" and "I also reproduced the sparse
+Merkle tree root in the extension lines".  See [tlog-mirror][] for an example.
 
 [note signature]: https://c2sp.org/signed-note
 [checkpoint]: https://c2sp.org/tlog-checkpoint
 [witness]: https://c2sp.org/tlog-witness
+[tlog-mirror]: https://github.com/C2SP/C2SP/pull/140/


tlog-cosignature.md
```diff
@@ -71,34 +80,55 @@ over the message defined in the next section.
...
-the log based on the first three lines of the checkpoint; no semantic statement
-is made about any extension lines, and consensus between witnesses on the
-extension lines SHALL NOT be assumed.
+the log based on the first three lines of the checkpoint; consensus between
+witnesses on the extension lines SHALL NOT be assumed, and no semantic statement
+is made about any extension lines unless the witness operator says otherwise.
+
+A witness operator MAY make additional statements about a checkpoint.  These
+additional statements need to be communicated out of band to those defining
+trust policies based on using a unique witness name or a unique witness key.
```
eliasrudberg: I think there may be problems related to the "using a unique witness name" here. How can anyone be certain that a name is unique?

Since anyone can start a new witness without needing to know about all other existing witnesses, it can happen that the same name is used by separate witnesses. Those witnesses can have different opinions about what the name means in terms of which "additional statements" the name indicates.

So I think that the name can be useful only when taken together with the key.

Therefore I wonder if instead of "based on using a unique witness name or a unique witness key", it may be better to write "based on using a unique combination of witness name and witness key", or something like that. What do you think?

rgdd: > So I think that the name can be useful only when taken together with the key.

That's the intent, I'll see what I can do to clarify tomorrow. Thanks for the comment!

rgdd: Updated to clearly specify that it is the tuple (witness name, public key, cosignature/v2 algorithm) that needs to be distinct for disambiguation, and that it's okay to use the same key with multiple different names (unlike in v1). 

Let me know what you think and if you consider your comment addressed!

tlog-cosignature.md
```diff
@@ -71,34 +80,57 @@ over the message defined in the next section.
...
+checkpoint tree size" and "I also reproduced the sparse Merkle tree hash in the
+extension lines for this checkpoint size".  See [tlog-mirror][] for an example.
+
+## Changes compared to cosignature/v1
+
+* The witness name is cryptographically tied to the resulting cosignature.
+* It is explicitly allowed to make additional statements about a checkpoint.
+  These additional statements need to be communicated out of band, identified by
+  the tuple (witness name, public key, cosignature/v2) in trust policies.  In
+  cosignature/v1, the public key cannot be reused with different witness names.
```
mhutchinson: Can you add some justification for _why_ these changes are useful. The main purpose behind this seems to be to allow the same key to be used in different contexts (by setting different witness names). Why is embedding the name better than using different keys for different types of claim?

tlog-cosignature.md
eliasrudberg: @rgdd thanks, I think it's much better now.

I think it's good to spell it out like you have done in one place, like this:

> tuples of (witness name, public key, supported cosignature version)

Then it's clear that what goes into a trust policy is information about the supported (required?) cosignature version.

In two other places the text says 

> tuples of (witness name, public key, cosignature/v2)

and that for me becomes confusing because it sounds like an actual consignature is part of the tuple, which could lead to some misunderstanding. So I think it would be better to write the whole "tuples of (witness name, public key, supported cosignature version)" in the other places also, or find some better way to express it.

What I mean is that when you write "(witness name, public key, cosignature/v2)" then a reader might take that literally and think that the tuple consists of:
- a witness name
- a public key
- a cosignature/v2

when in fact the tuple really consists of:
- a witness name
- a public key
- information about the required cosignature version

I think that I do understand what you mean, I just think there is a risk of misunderstanding there.

Another thought is that maybe instead of "supported cosignature version" it would be better to say "required cosignature version". My thinking is that someone writing a trust policy will write what is *required* for verification. But I'm not sure, I guess it depends on how you look at things.

tlog-cosignature.md
rgdd: The upside of allowing a signing context (a.k.a witness name - maybe it should be called something else) is that witness operators who also want to make additional statements can manage fewer keys. I'd say it is only marginally cleaner on paper that the witness name is tied to the cosignature - so that's not a good why in and of itself.

You're right that it would be possible to stick to cosignature/v1 and extend with language about additional statements and distinct keys.

mhutchinson: I very much support `cosignature/v1` being flexible enough to work as a platform to also make other claims. I didn't see anything in the spec that ruled that out, but we could add some words to say that key owners MAY also define _futher_ claims (e.g. mirroring, or having verified up to this point in the log).

This was always the intent, in a similar way that I think log signatures can also make other claims beyond "append-only, globally consistent", such as "there are no two conflicting statements in this log".

rgdd: The PR for this would basically be the same, expect that there's no name incorporated into the cosignature so the key would have to be unique to disambiguate, e.g., "now i've checked append-only" and "now i've also mirrored the log" 1m later. (This assumes the claims happen at different times -- doesn't necessarily have to be the case.)

A perhaps easier way forward would be to extend cosignature/v1 similar to this PR but without the name, and then at some point in the future if/when there's a v2 clean up so that the name is incorporated into the cosignature. And then possibly make other changes as well. But basically do this on a longer timeline.

Very open for input here (thanks for the input so far martin and elias)!

tlog-cosignature.md
rgdd: Thanks, rephrased consistently with "supported cosignature version" in all places that mention the tuple.

rgdd: RE: supported vs required, i think both works.

tlog-cosignature.md
mhutchinson: Yeah, this sounds good to me. I'm not against this change as part of a bundle of other changes, but on its own, this doesn't warrant a v2 IMO.

tlog-cosignature.md
```diff
@@ -95,10 +99,23 @@ log identified by the origin line has the specified root hash.
...
-is made about any extension lines, and consensus between witnesses on the
-extension lines SHALL NOT be assumed.
+the log based on the first three lines of the checkpoint; consensus between
+witnesses on the extension lines SHALL NOT be assumed, and no semantic statement
+is made about any extension lines unless the witness operator says otherwise.
+
+A witness operator MAY make additional statements about a checkpoint.  These
+additional statements need to be communicated out of band to those defining
+trust policies based on tuples of (public key, supported cosignature version).
+The same tuple MUST NOT be reused across different cosigning contexts.
```
mhutchinson: "different cosigning contexts" is potentially ambiguous; from our discussions I know that a cosigning context fixes a set of claims that can be relied on. But this sentence could be interpreted to mean something like "each key should only be used for cosigning a single log".

mhutchinson: Pulled the trigger before reading the next sentence...

rgdd: Updated the text to hopefully make this even more clear -- let me know!

tlog-cosignature.md
```diff
@@ -95,10 +99,22 @@ log identified by the origin line has the specified root hash.
...
-is made about any extension lines, and consensus between witnesses on the
-extension lines SHALL NOT be assumed.
+the log based on the first three lines of the checkpoint; consensus between
+witnesses on the extension lines SHALL NOT be assumed, and no semantic statement
+is made about any extension lines unless the witness operator says otherwise.
+
+A witness operator MAY make additional statements about a checkpoint.  These
+additional statements need to be communicated out of band to those defining
+trust policies based on tuples of (public key, supported cosignature version).
+The same tuple MUST NOT be reused across different cosigning contexts, where
```
mhutchinson: There are two terms that are introduced here that are not _defined_, but exemplars are provided: "cosigning context" and "additional implied statement". The latter is suboptimal because the statement shouldn't be implied, it should be explicitly stated somewhere. I think we could probably avoid the terms. I'm using claim here, but this doesn't necessarily tie it to the claimant model, I'm using it according to a standard English definition:

"""
A tuple MUST imply a single set of claims. These MUST include the `cosignature/v1` claims and MAY include other non-conflicting claims such as "I also mirrored the log up until the checkpoint size" and "I also reproduced the sparse Merkle tree root in the extension lines".  See [tlog-mirror][] for an example."
"""

rgdd: How about that? Basically your suggestion but sticking to "statement" instead of "claim". Mostly because that's the term we've been using in the spec so far, and i think claim runs the risk of being preloaded with other thoughts due to e.g. the claimant model.

(And thanks -- very good suggestions!)

rgdd: > How about that?

Refers to the dc85..commit, maybe wasn't obvious hehe. :)

mhutchinson: I can live with this, even though the terminology bothers me a little. In the base draft, the first 2/3 of the usages of "statement" are perfect: they talk about the signature line as being the statement. The last usage ("no semantic statement is made about any extension lines") is a bit funky, but if that's how we're using it now, then YOLO.

Thanks for being so cooperative with my suggestions 👍

rgdd: I'm not opposed to updating the terminology, but I prefer that we do it in a separate PR. Thanks again for the very helpful input and suggestions so far!

rgdd:
Squashed up the fixup commits and added a thanks in the commit message to @mhutchinson and @eliasrudberg for helping me tinker on the text in this PR.

tlog-cosignature.md
```diff
@@ -95,10 +99,20 @@ log identified by the origin line has the specified root hash.
...
+is made about any extension lines unless the witness operator says otherwise.
+
+A witness operator MAY make additional statements about a checkpoint.  These
+additional statements need to be communicated out of band to those defining
+trust policies based on tuples of (public key, supported cosignature version).
+A given tuple MUST imply a single set of statements.  These statements MUST
+include the cosignature/v1 semantics, and MAY include other statements that
+are non-conflicting.  Examples of non-conflicting statements include "I also
+mirrored the log up until the checkpoint size" and "I also reproduced the sparse
+Merkle tree root in the extension lines".  See [tlog-mirror][] for an example.
```
davidben, line 113: I might perhaps add something like "A witness operator that operates multiple witness instances making distinct additional statements MUST use distinct public keys for each witness instance."

davidben (approved):
Works for me!  Thanks! Yeah, Filippo mentioned he was leaning that way actually before I put together the PR. I only left it alone just to illustrate the v2 approach and figured you all can decide what you prefer. I'm happy with either.

Once this lands, I can go update the tlog-mirror PR and also change some of the terminology in Merkle Tree Certs.

In a vacuum, I think binding the witness name would have been good, but it's not critical, and it totally makes sense to not want to rev the signature construction for a change like that. (Plus, for MTCs, I expect we'd be reusing the concepts but defining our own signature construction anyway, so I especially don't have a horse in this race. Nothing against the C2SP construction. Just fallout from figuring out how to thread the C2SP/IETF needle because our signatures will intersect with the X.509 bits. Also we have this goofy [subtrees](https://davidben.github.io/merkle-tree-certs/draft-davidben-tls-merkle-tree-certs.html#name-subtrees) optimization.)

Regarding terminology, I noticed you also ended up with "witness" instead of "cosigner". That's fine by me! I interpreted "witness" to specifically mean the consistency-proof-following construction, but I was just guessing. Happy to just defer to you all on how to define these. :-)

Some misc thoughts on that, just to make sure we're on the same page:

I guess the thinking then is we'd say things like "a mirror is a witness that additionally promises to have mirrored the log", etc. Is that right?

I think another consequence is that a "witness" might not *necessarily* implement the protocol in tlog-witness, just any service that generates cosignatures meeting the guarantees. (It happens that tlog-mirror implements a modified witness protocol anyway, because this funky two-stage commit thing seemed valuable, but I think it would have been reasonable for one to not do that.)

One last thing is the relationship between cosignatures and log signatures. Now, in C2SP you all already separated them and I wouldn't propose retroactively merging them for C2SP. But for MTCs, I actually ended up merging the two.

The observation that log signature semantics are _also_ compatible with these new witness semantics. A witness signature isn't that it checked consistency proofs, but that every checkpoint it ever signs is consistent with each other. The log operator promises that too, just by virtue of knowing what it signed.

The merger then provided a nice hook (the additional statements framework) for the extra promise MTCs wanted to embed, which is "I have certified every entry in this checkpoint". It also means the relying party's verification process is very tidy and uniform ("look for sufficient cosignatures to satisfy you—you should look for enough to make you happy about authenticity and transparency but the core protocol is pretty agnostic to all this"). It also meant we didn't have to define subtree signatures twice, etc.

It also makes it very easy for log operators, as a robustness measure, do something like not actually expose the core log key, but instead operate a pair of (log, witness/mirror), with the witness/mirror key being the exposed log key. That way you defend a bit against accidental mishaps.

So, with that context, either in the MTC world which has to define a new signature construction anyway, or in a hypothetical alternate universe where tlog had merged the two... would "witness" still be a reasonable term for this even further generalized witness-or-mirror-or-log-owner signer?

rgdd:
> Works for me! Thanks! Yeah, Filippo mentioned he was leaning that way actually before I put together the PR. I only left it alone just to illustrate the v2 approach and figured you all can decide what you prefer. I'm happy with either.

Great -- thanks for the input!

> 
> Once this lands, I can go update the tlog-mirror PR and also change some of the terminology in Merkle Tree Certs.
> 
> In a vacuum, I think binding the witness name would have been good

100% agree, if/when there's a v2 I will advocate for this.

> , but it's not critical, and it totally makes sense to not want to rev the signature construction for a change like that. (Plus, for MTCs, I expect we'd be reusing the concepts but defining our own signature construction anyway, so I especially don't have a horse in this race. Nothing against the C2SP construction. Just fallout from figuring out how to thread the C2SP/IETF needle because our signatures will intersect with the X.509 bits. Also we have this goofy [subtrees](https://davidben.github.io/merkle-tree-certs/draft-davidben-tls-merkle-tree-certs.html#name-subtrees) optimization.)

I'll probably provide some comments somewhere else (Filippo provided hints on where it would be helpful), but what I can say right away is I did read the latest Merkle tree certificates draft and I liked the overall direction.

Pretty sure that the maintainers of c2sp.org's tlog specs are all interested in shaping any specs that have to be in IETF for $reasons. We've put quite a lot of effort in trying to converge the ecosystem rather than splitting it, and so far we've been very successful at that. I'm sure we'll be talking more to figure out what's the most helpful way to move things forward. 

> 
> Regarding terminology, I noticed you also ended up with "witness" instead of "cosigner". That's fine by me! I interpreted "witness" to specifically mean the consistency-proof-following construction, but I was just guessing. Happy to just defer to you all on how to define these. :-)

This is actually closer to my mental model. I.e., when I think of a witness, I think of a "cheap" append-only verifier that doesn't have to see the entire log (as per the tlog-witness spec). And if all goes well, produce a cosignature.

> 
> Some misc thoughts on that, just to make sure we're on the same page:
> 
> I guess the thinking then is we'd say things like "a mirror is a witness that additionally promises to have mirrored the log", etc. Is that right?

Yes. But the more I think about this I don't think it's helpful, see the bottom of this response.

> 
> I think another consequence is that a "witness" might not _necessarily_ implement the protocol in tlog-witness, just any service that generates cosignatures meeting the guarantees. (It happens that tlog-mirror implements a modified witness protocol anyway, because this funky two-stage commit thing seemed valuable, but I think it would have been reasonable for one to not do that.)
>

This is consistent with how I think of the specifications, i.e., tlog-witness should depend on tlog-cosignature but not vice versa.

> One last thing is the relationship between cosignatures and log signatures. Now, in C2SP you all already separated them and I wouldn't propose retroactively merging them for C2SP. But for MTCs, I actually ended up merging the two.
>

I don't see a problem with this, assuming proper namespaces so it's clear what's signed and who's signing (such as in the proposed cosignature/v2).

We did not consider this a few years ago because we optimized for backwards compatibility with existing things like Go's checksum database. I appreciate the observation though, nice one!

> The observation that log signature semantics are _also_ compatible with these new witness semantics. A witness signature isn't that it checked consistency proofs, but that every checkpoint it ever signs is consistent with each other. The log operator promises that too, just by virtue of knowing what it signed.
> 
> The merger then provided a nice hook (the additional statements framework) for the extra promise MTCs wanted to embed, which is "I have certified every entry in this checkpoint". It also means the relying party's verification process is very tidy and uniform ("look for sufficient cosignatures to satisfy you—you should look for enough to make you happy about authenticity and transparency but the core protocol is pretty agnostic to all this"). It also meant we didn't have to define subtree signatures twice, etc.
> 
> It also makes it very easy for log operators, as a robustness measure, do something like not actually expose the core log key, but instead operate a pair of (log, witness/mirror), with the witness/mirror key being the exposed log key. That way you defend a bit against accidental mishaps.
> 
> So, with that context, either in the MTC world which has to define a new signature construction anyway, or in a hypothetical alternate universe where tlog had merged the two... would "witness" still be a reasonable term for this even further generalized witness-or-mirror-or-log-owner signer?

Not sure what a unified format would be called, but it's probably not the most important part. What is important is to have clear terminology for the different parties that use the format, because trust policies will need to be specified based on understanding what the signatures from these different parties mean.

The key words I'd expect in a trust policy (and what meaning I'd associate them with) would be something like:

log - authoritatively moves the tree head forward, publishes entries; append-only for anything else to make sense

mirror - happy to cosign the most recent tree head it knows about and managed to replicate from entries; append-only is a requirement to make sense

witness - happy to cosign the most recent tree head it knows about if append only

I would find it a bit akward and clunky to say "a mirror is a witness that also replicates the log". I would rather help people build the mental model that a witness is a cheap append-only verifier whereas a mirror is a log replicator that (obviously) needs to verify append-only to function correctly. Both can use cosignature/v1.

With this in mind, I think you were onto something when doing s/witness/cosigner in a bunch of places and I missed it. I'll make one more commit with this change and let others weigh in on the change too. For me, it's a good idea because it's easier to overload "cosigner" and couple it with a distinct name, compared to overloading witness and trying to bolt it on to a different name like "mirror". And I do think it is helpful to make it easy to find distinct (and ideally easy) names.

Thanks again for all the helpful suggestions / input / comments / questions / etc!

davidben:
> Pretty sure that the maintainers of c2sp.org's tlog specs are all interested in shaping any specs that have to be in IETF for $reasons. We've put quite a lot of effort in trying to converge the ecosystem rather than splitting it, and so far we've been very successful at that. I'm sure we'll be talking more to figure out what's the most helpful way to move things forward.

Of course! Yeah, the MTC draft was just what seemed the best starting point to present to both communities. I'm sure there'll be a lot of hopping between IETF and C2SP to converge on what makes everyone happy. :-)

> With this in mind, I think you were onto something when doing s/witness/cosigner in a bunch of places and I missed it. I'll make one more commit with this change and let others weigh in on the change too. For me, it's a good idea because it's easier to overload "cosigner" and couple it with a distinct name, compared to overloading witness and trying to bolt it on to a different name like "mirror". And I do think it is helpful to make it easy to find distinct (and ideally easy) names.

Works for me. New version still LGTM.

tlog-cosignature.md
```diff
@@ -89,16 +93,26 @@ leading zeroes. This value MUST match the `timestamped_signature.timestamp`.
...
-is made about any extension lines, and consensus between witnesses on the
-extension lines SHALL NOT be assumed.
+that the cosigner is asserting observation of correct append-only operation of
+the log based on the first three lines of the checkpoint; consensus between
+cosigners on the extension lines SHALL NOT be assumed, and no semantic statement
+is made about any extension lines unless the cosigner's operator says otherwise.
+
+A cosigner MAY make additional statements about a checkpoint.  These
+additional statements need to be communicated out of band to those defining
+trust policies based on tuples of (public key, supported cosignature version).
```
lukevalenta, line 108: This should be the same tuple as configured on clients above, right?
```suggestion
trust policies based on tuples of (cosigner name, public key, supported cosignature version).
```

lukevalenta (approved):
LGTM! I'm happy with sticking with cosignature/v1 for the time being.

tlog-cosignature.md
davidben: @rgdd can correct me if I'm wrong, but I interpreted this to mean that the public key (and version) alone must be enough to fix the set of statements. I.e. this was the workaround for the cosigner name not being bound into the signature.

(Including the version is slightly interesting. Presumably as soon as we bump the version, we'll fix that, and _then_ it'll be possible to have multiple statement sets per key. But meh, we can sort out that text when it comes to it.)

lukevalenta: Makes sense! Probably worth adding a sentence to that effect here (pointing out that since the cosigner name is not bound in signatures, it should not be used to fix statements).

tlog-cosignature.md
FiloSottile: Agreed, I think it is implied by one possible reading of the rest of the text, but making it explicit is better than not.

davidben:
Uploaded https://github.com/C2SP/C2SP/pull/149 to pick up the change in https://github.com/C2SP/C2SP/pull/146/files#r2177948486

PR #147, opened 2025-06-30
chacha20-poly1305-siv: new specification

### LINKED ISSUE #129: New spec: ChaCha20-Poly1305-SIV
_ChaCha20-Poly1305-SIV_ (_CCP-SIV_) is a misuse-resistant, key-committing AEAD scheme. It is built from ChaCha20 and Poly1305 without modifying their designs, making it compatible with existing APIs from cryptographic libraries. Furthermore, the overhead is just two ChaCha20 blocks over ChaCha20-Poly1305 whilst supporting a larger nonce.

More information can be found in the draft at #130.

It would be maintained by me.

This design has not received any peer review, so I would appreciate feedback. It obviously shouldn't be published if there's a problem.

/cc @FiloSottile, @dconnolly, @str4d

FiloSottile:
Closes #129


str4d (approved):
ACK cbcfb46f6e9f3826e499be1e8ad15e721ce7f839

PR #155, opened 2025-08-15
Update pruning text to handle a tree size of zero in tlog-tiles.md

lukevalenta:
The minimum index as currently defined can be equal to the tree size only if the tree size is zero.

### DIFF
--- a/tlog-tiles.md
+++ b/tlog-tiles.md
@@ -190,11 +190,11 @@ deleting the resources from storage. This means:
 
 This pruning criteria allows a log client to obtain:
 
-* Any entry whose index is greater or equal to the minimum index
+* Any entry whose index is greater than or equal to the minimum index
 * The root hash for any tree whose size is greater than or equal to the minimum
   index
 * An inclusion proof between any entry and any containing tree, if the entry's
-  index is greater or equal to the minimum index
+  index is greater than or equal to the minimum index
 * A consistency proof between any two trees whose sizes are both greater than
   or equal to the minimum index
 


lukevalenta:
The text "[t]he minimum index is the index of the first log entry that the log publishes" is still not quite right for a tree size of zero, since there is no index-0 entry yet. Maybe it should be "minimum size" instead of "minimum index" to avoid the awkwardness of dealing with the size-0 tree?

davidben:
Hmm. I guess I figured you're allowed to, at least mechanically, prune all the way up to your tree size. Then the starting point (tree size = 0, min idx = 0) isn't a special case. I think everything still works out just fine if you do that, it's just kinda silly.

Perhaps we should s/the index of the first log entry that the log publishes/a lower bound on the indices of the log entries that the log publics/?

lukevalenta:
> Perhaps we should s/the index of the first log entry that the log publishes/a lower bound on the indices of the log entries that the log publics/?

That could work, but then maybe we'd want to call the variable "pruned size" or similar then? I think the awkwardness come from the fact that we want this value to initially be zero, but if it refers to an index then we need special handling since the index-0 entry doesn't exist in an empty tree.

If we make the variable refer to a tree size instead, I think it fixes that awkwardness. Here's an attempt:

"pruned size": a lower bound on the indices of entries that the log publishes; entries with indices below the pruned size are not guaranteed to be available.

davidben:
Ah I see. No particular preferences on my end. I guess whether it's a tree size or an index is whether you focus on the [0, X) range or the [X, tree_size) range. Intuitively, the latter feels like the more important one, but I guess we also talk about pruning and pruning is a thing that happens to the [0, X) range. 🤷

lukevalenta:
> the latter feels like the more important one

That's a good point and I agree. Thinking about it in terms of the half-open interval helps; since [0, 0) is empty, nothing is published. With that in mind, your rephrasing from https://github.com/C2SP/C2SP/pull/155#issuecomment-3192598452 makes sense to me. Thanks!

Also, I _think_ the log must still keep around any tiles needed to reconstruct the root hash of the minimum index-sized tree. (e.g., for a tree with minimum index X, the log still has to be able to keep the 'left edge' necessary to reconstruct the root hash of the tree of size X, since that's needed for consistency proofs to grow the tree).

I'll mull over it a bit more over the weekend, and push an update to the PR.

davidben:
> Also, I think the log must still keep around any tiles needed to reconstruct the root hash of the minimum index-sized tree. (e.g., for a tree with minimum index X, the log still has to be able to keep the 'left edge' necessary to reconstruct the root hash of the tree of size X, since that's needed for consistency proofs to grow the tree).

I believe that actually just falls out of the pruning criteria. If it doesn't, we should fix that. :-) Specifically, the pruning criteria doesn't delete partial tiles. (If it were to, we would have to treat the end index as the corresponding full tile's end index, which is greater than the tree size and thus greater than min_idx.)

Specifically, the pruning criteria will only delete full tile `[start, end)` if `end <= min_idx`. You might still care about the root hash of that tile... but if you have a full tile, the level above you definitely exists and contains a (full or partial) tile which contains your root hash. So now we're okay by induction. Following the induction, if you keep going up, you'll always eventually find either an unpruned full tile, or the partial tile at that level.

lukevalenta:
> I believe that actually just falls out of the pruning criteria.

Yep, agreed! One thing I was caught up on was the text "[t]his pruning criteria allows a log client to obtain.. [t]he root hash for any tree whose size is greater than the minimum index". I think this should be greater than _or equal to_ the minimum index. In the commit message https://github.com/C2SP/C2SP/pull/155/commits/ac8091c152962da480954051ae3213710ca2de6a I tried to justify why.

```
Also clarify that we can always compute the root hash of the tree of
size _equal_ to (not just strictly greater than) the minimum index.  To
see this, say the minimum index is X.

* If X=0, the root hash is just SHA256(<empty string>)
* If X=tree_size, the root hash can be computed from the 'right edge' of
  nodes (or pulled from the current checkpoint).
* If 0<X<tree_size, the inclusion proof of the index-X element in the
  tree of size X+1 necessarily includes the tree nodes to compute the
  root hash for the tree of size X.
  ```

Do you buy it?

davidben:
Ah yeah, I buy it.

davidben (approved):
LGTM, though we'll also need a tlog-tiles person to approve it.

mhutchinson (approved):

Issue #159, opened 2025-09-02
New spec: COCKTAIL-DKG (Distributed Key Generation for FROST (RFC 9591))


soatok:
[ChillDKG](https://github.com/BlockstreamResearch/bip-frost-dkg) is a Distributed Key Generation protocol for [FROST](https://www.rfc-editor.org/rfc/rfc9591.html) that doesn't require a secure broadcast channel to use safely. ChillDKG exists as a BIP for Bitcoin and is only specified over the secp256k1 curve.

With permission from Tim Ruffing and Jonas Nick, I'm proposing a C2SP fork of ChillDKG that covers the entire menu of ciphersuites from RFC 9591. 

I call my fork, COCKTAIL. The name COCKTAIL is a recursive acronym that stands for, "COCKTAIL Orchestrates Cryptographic Key Threshold Agreement for Interoperable Libraries."

* Short name: `cocktail-dkg`
* Proposed maintainers: @soatok @bytemare @tjade273

I will follow this issue up with a pull request in the near future with the first draft of the proposed specification text.

/cc @C2SP/stewards

str4d:
ACK on inclusion of this spec.

With my Zcash dev hat on, I would love for this to be written in a way that makes it easily extendable to (or include) the proposed [ZIP 312 ciphersuites](https://zips.z.cash/zip-0312#ciphersuites).

PR #164, opened 2025-10-08
new spec: COCKTAIL-DKG

soatok:
See #159

This is a first pass at implementing a specification for COCKTAIL-DKG, a generalization of ChillDKG for other elliptic curve groups.

The test vectors are still undefined. I figured we could calculate them later.

 /cc @soatok @bytemare @tjade273, @C2SP/stewards

**To view the changes quickly, [click here](https://github.com/soatok/C2SP/blob/cocktail-dkg-initial/cocktail-dkg.md).**

### DIFF
--- a/cocktail-dkg.md
+++ b/cocktail-dkg.md
@@ -2,8 +2,8 @@
 
 [c2sp.org/cocktail-dkg](https://c2sp.org/cocktail-dkg)
 
-- **Version**: v0.0.1
-- **Authors**: 
+- **Version**: v0.1.0
+- **Authors**:
   - [Daniel Bourdrez](https://github.com/bytemare)
   - [Soatok Dreamseeker](https://github.com/soatok)
   - [Tjaden Hess](https://github.com/tjade273), *[Trail of Bits](https://trailofbits.com)*
@@ -26,8 +26,8 @@ However, [the original FROST paper](https://eprint.iacr.org/2020/852) did specif
 Secret Sharing, Pedersen commitments, and Proofs of Possession. This protocol is secure only if communications are
 performed over a secure channel, with the following requirements:
 
-* Messages are authenticated and encrypted in transit, to prevent man-in-the-middle attacks
-* All participants have some consensus mechanism (e.g., a transparency log) so everyone has a consistent view of the
+- Messages are authenticated and encrypted in transit, to prevent man-in-the-middle attacks
+- All participants have some consensus mechanism (e.g., a transparency log) so everyone has a consistent view of the
   protocol messages
 
 To satisfy these requirements in a standalone protocol, Nick and Ruffing proposed [ChillDKG](https://github.com/BlockstreamResearch/bip-frost-dkg)
@@ -91,6 +91,9 @@ throughout the COCKTAIL-DKG protocol.
 - $x_i$: The final, long-lived secret share for participant $i$, where $x_i = \sum_{j=1}^{n} s_{j,i}$.
 - $Y_i$: The public verification share for participant $i$.
 - $Y$: The final group public key, where $Y = \sum_{j=1}^{n} C_{j,0}$.
+- $payload_{i,j}$: An optional application-defined payload from participant $i$ to participant $j$, which may be empty.
+- $S^{(e)}_{i,j}$: The ephemeral-to-static ECDH shared secret, where $S^{(e)}_{i,j} = e_i * P_j$.
+- $S^{(d)}_{i,j}$: The static-to-static ECDH shared secret, where $S^{(d)}_{i,j} = d_i * P_j$.
 
 ### Operations
 
@@ -109,14 +112,16 @@ defined below. Implementations **MUST** adhere to these formats to ensure intero
 
 Let $G$ be an elliptic curve group with a standardized name (e.g., "P-256", "Ed25519", "secp256k1").
 
-* **Scalar**: A scalar is an integer in the range $[0, q-1]$, where $q$ is the order of the group $G$. A scalar is
+- **Scalar**: A scalar is an integer in the range $[0, q-1]$, where $q$ is the order of the group $G$. A scalar is
   encoded as a fixed-size, big-endian byte array. The size is determined by the curve (e.g., 32 bytes for P-256,
-  secp256k1, and Ed25519/Ristretto255).
-* **Elliptic Curve Point**: An elliptic curve point is encoded in its compressed form, as a fixed-size byte array. The
+  secp256k1, and Ed25519/Ristretto255). Implementations **MUST** reject any decoded scalar value $s$ where $s \geq q$.
+- **Elliptic Curve Point**: An elliptic curve point is encoded in its compressed form, as a fixed-size byte array. The
   size and format are determined by the curve (e.g., 33 bytes for P-256 and secp256k1, 32 bytes for
-  Ed25519/Ristretto255).
-* **The Point at Infinity**: The point at infinity, which is the identity element of the group, is represented by a byte
-  array of the same length as a standard point encoding, but filled with all zero bytes.
+  Ed25519/Ristretto255). Implementations **MUST** validate that decoded points are valid curve points.
+- **The Point at Infinity**: The point at infinity, which is the identity element of the group, is represented by a byte
+  array of the same length as a standard point encoding, but filled with all zero bytes. The point at infinity **MUST**
+  be rejected when parsing VSS commitment points $C_{i,k}$, ephemeral public keys $E_i$, and static public keys $P_i$.
+  Accepting identity elements in these positions would compromise protocol security.
 
 #### Protocol Messages
 
@@ -128,13 +133,19 @@ representations of their constituent parts in the specified order.
 This message contains the participant's VSS commitment, their Proof-of-Possession, their ephemeral public key, and the
 encrypted shares for all other participants.
 
-* $C_i$: The VSS commitment, which is a list of $t$ elliptic curve points.
-  * Format: $C_{i,0} \parallel C_{i,1} \parallel \cdots \parallel C_{i,t-1}$
-* $PoP_i$: The Proof of Possession, which is a signature. The size depends on the signature scheme used by the
+- $C_i$: The VSS commitment, which is a list of **exactly** $t$ elliptic curve points. Implementations **MUST** verify
+  that the commitment contains exactly $t$ points; any other length indicates a malformed message or an attempted
+  [threshold elevation attack](https://blog.trailofbits.com/2024/02/20/breaking-the-shared-key-in-threshold-signature-schemes/).
+  - Format: $C_{i,0} \parallel C_{i,1} \parallel \cdots \parallel C_{i,t-1}$
+- $PoP_i$: The Proof of Possession, which is a signature. The size depends on the signature scheme used by the
   ciphersuite.
-* $E_i$: The ephemeral public key, an elliptic curve point. 
-  * It does not refer to isogenies. Here, E stands for "ephemeral".
-* $c_{i,j}$: An encrypted secret share. The size depends on the AEAD scheme used by the ciphersuite.
+- $E_i$: The ephemeral public key, an elliptic curve point.
+  - It does not refer to isogenies. Here, E stands for "ephemeral".
+- $c_{i,j}$: An encrypted ciphertext containing the secret share $s_{i,j}$ and an optional application-defined payload
+  $payload_{i,j}$. The plaintext format is $s_{i,j} \parallel payload_{i,j}$, where $payload_{i,j}$ may be empty.
+  The ciphertext size depends on the AEAD scheme and the payload size. The minimum ciphertext size is the scalar
+  encoding size plus the AEAD authentication tag size (e.g., 32 + 16 = 48 bytes for most ciphersuites).
+  Implementations **SHOULD** enforce a maximum ciphertext size to prevent resource exhaustion attacks.
 
 The full message is the concatenation of these elements:
 
@@ -146,12 +157,12 @@ msg_{1|i} = C_i \parallel PoP_i \parallel E_i \parallel c_{i,1} \parallel c_{i,2
 
 This message aggregates the public information from all participants.
 
-* $C_{j,0}$: The zero-coefficient commitment from participant $j$, an elliptic curve point.
-* $PoP_j$: The Proof of Possession from participant $j$.
-* $C_{agg}$: The aggregated commitment for non-zero coefficients. This is a list of $t-1$ points.
-  * Format: $C_{agg,1} \parallel \cdots \parallel C_{agg,t-1}$
-* $E_j$: The ephemeral public key from participant $j$.
-* $c_{j,i}$: The ciphertext from participant $j$ intended for participant $i$.
+- $C_{j,0}$: The zero-coefficient commitment from participant $j$, an elliptic curve point.
+- $PoP_j$: The Proof of Possession from participant $j$.
+- $C_{agg}$: The aggregated commitment for non-zero coefficients. This is a list of $t-1$ points.
+  - Format: $C_{agg,1} \parallel \cdots \parallel C_{agg,t-1}$
+- $E_j$: The ephemeral public key from participant $j$.
+- $c_{j,i}$: The ciphertext from participant $j$ intended for participant $i$.
 
 The message broadcast to participant $i$ is structured as:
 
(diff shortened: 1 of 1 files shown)


cocktail-dkg.md
```diff
@@ -0,0 +1,807 @@
...
+- $E_i$: The ephemeral public key for participant $i$, where $E_i = e_i * G$.
+- $PoP_i$: A Proof of Possession signature from participant $i$.
+- $x_i$: The final, long-lived secret share for participant $i$, where $x_i = \sum_{j=1}^{n} s_{j,i}$.
+- $Y_i$: The public verification share for participant $i$.
+- $Y$: The final group public key, where $Y = \sum_{j=1}^{n} C_{j,0}$.
+
+### Operations
+
+- $Add(P1, P2)$: Elliptic curve point addition.
+- $RandomScalar()$: Generates a cryptographically secure random scalar in the range $[0, q-1]$.
```
ineiti: Add "uniform"

cocktail-dkg.md
```diff
@@ -0,0 +1,807 @@
...
+
+This section provides detailed definitions for the notation, message formats, and cryptographic primitives used
+throughout the COCKTAIL-DKG protocol.
+
+### Notation
+
+- $n$: The total number of participants in the DKG ceremony.
+- $t$: The threshold, i.e., the minimum number of participants required to generate a signature.
+- $i, j$: Indices representing participants, where $1 <= i, j <= n$.
+- $G$: The generator point of the elliptic curve group.
```
ineiti: I would change this to $B$, for base point, so it's different from $G$, the group.

cocktail-dkg.md
```diff
@@ -0,0 +1,807 @@
...
+
+### Round 1: Commitment and Encryption
+
+1. **Generate Polynomial:** Participant $i$ generates a secret polynomial $f_i(x)$ of degree $t-1$:
+   $f_i(x) = a_{i,0} + a_{i,1}x + \cdots + a_{i,t-1}x^{t-1}$.
+   The coefficients $a_{i,k}$ are chosen as random scalars. $a_{i,0}$ is the participant's primary secret share.
+2. **Compute VSS Commitment:** Participant $i$ computes a VSS commitment $C_i$ to their polynomial by creating a public
+   commitment for each coefficient:
+   $C_i = (C_{i,0}, C_{i,1}, \cdots, C_{i,t-1})$, where $C_{i,k} = [a_{i,k}] G$.
+3. **Compute Proof of Possession (PoP):** Participant $i$ computes a digital signature $PoP_i$ over the $context$ string
```
ineiti: Shouldn't there be an additional context/domain in the signature? Signing something like

`"COCKTAIL-DKG-V1.0" | $context$ | $i$`

Would make re-usage of the signature in another context more difficult.

cocktail-dkg.md
```diff
@@ -0,0 +1,807 @@
...
+
+### Setup
+
+Each participant $i$ is assumed to have:
+- A unique identifier $i$ from $1$ to $n$.
+- The parameters $n$ (total participants) and $t$ (threshold).
+- A long-term static key pair $(d_i, P_i)$.
+- A list of the static public keys of all other participants, ${P_1, ..., P_n}$.
+- A ciphersuite defining the elliptic curve group, hash function, and AEAD scheme.
+- A $context$ string, which should be unique to the DKG session to prevent replay attacks.
```
ineiti: I propose to define the $context$ as something like this:

`H( "COCKTAIL-DKG-CONTEXT" | Seconds_since_unix_epoch | P_1 | P_2 | ... | P_n )`

This allows every participant to verify that they agree on the context before signing and producing proofs.

cocktail-dkg.md
```diff
@@ -0,0 +1,807 @@
...
+### Round 2: Share Decryption and Verification
+
+The coordinator waits to receive $msg_{1|i}$ from all $n$ participants. It then broadcasts a list of all received messages
+to every participant. Upon receiving the list of all $msg_{1|i}$ messages, each participant `i` performs the following
+steps:
+
+1. **Verify All PoPs:** For each other participant $j$ from $1$ to $n$:
+    - Participant $i$ verifies the proof of possession $PoP_j$ using participant $j$'s public commitment $C_{j,0}$ and
+      the $context$ string.
+    - If any $PoP_j$ is invalid, participant $i$ aborts, identifying participant $j$ as malicious.
```
ineiti: If the coordinator is _honest but curious_ this works. But in the case of a malicious coordinator they could send a wrong signatures which would censor certain participants. I think there is a possibility to avoid this, but it would complicate the protocol a bit.

cocktail-dkg.md
```diff
@@ -0,0 +1,807 @@
...
+- **COCKTAIL(Pallas, BLAKE2b-512)**
+  - **Note**: Compatible with [ZIP-0312](https://zips.z.cash/zip-0312#ciphersuites).
+  - **`H6` Hash**: BLAKE2b-512 (with a 56-byte output)
+  - **`H6` Prefix**: `COCKTAIL-DKG-Pallas-BLAKE2b-H6`
+  - **Key/Nonce**: The first 32 bytes of the `H6` output are the key, and the next 24 bytes are the nonce.
+  - **AEAD**: XChaCha20-Poly1305
+
+## Security Considerations
+
+- **Coordinator Role**: The coordinator is trusted for availability and to correctly broadcast messages, but is not
```
ineiti, line 549: I would also consider how the participants interact with the coordinator:

- how do the participants authenticate with the coordinator? W/o this, there is an easy way to abort the protocol by third parties
- how do the participants and the coordinator start the protocol and choose the challenge? This is an important step if you want to actually implement the protocol

cocktail-dkg.md
```diff
@@ -0,0 +1,807 @@
...
+
+### Blame-Finding and Reporting
+
+A key feature of a secure DKG protocol is the ability to identify malicious participants. When an error occurs, the
+protocol should not only terminate but also output information about who caused the failure.
+
+* **Coordinator's Role**: The coordinator is in a unique position to detect errors in $msg_{1|i}$ messages. If the
+  coordinator receives a malformed message or one with an invalid PoP from participant $i$, it should immediately abort
+  the protocol and broadcast a blame message identifying participant $i$ as faulty.
+* **Participant's Role**: Participants must validate all data they receive. If a participant $i$ receives an invalid
```
ineiti: I would well define how the protocol looks to prove that a participant is malicious. From my experience it's extremely difficult to avoid wrong blame. In this protocol the coordinator could create a random signature and send it in place of the correct signature of participant $e$. This would then lead to all participants blaming participant $e$.

cocktail-dkg.md
```diff
@@ -0,0 +1,807 @@
...
+
+The output of $H6$ is used to derive the key and nonce for the AEAD.
+
+### Ciphersuite Definitions
+
+- **COCKTAIL(Ed25519, SHA-512)**
+  - **`H6` Hash**: SHA-512
+  - **`H6` Prefix**: `COCKTAIL-DKG-Ed25519-SHA512-H6`
+  - **Key/Nonce**: The first 32 bytes of the `H6` output are the key, and the next 24 bytes are the nonce.
+  - **AEAD**: XChaCha20-Poly1305
```
tmthrgd, line 437: I can't see any advantage to using the X-variant if you're deriving both the key and nonce with `H6`. It's just a wasted HChaCha20 call, while being marginally less standardised.

cocktail-dkg.md
soatok: That's what CertEq is for.

ineiti: OK, I see now that in _Error Categories_, 2., it is mentioned that the protocol MUST abort in case of such an error. This is definitely the easiest solution, but allows all participants to abort the protocol, so it needs all participants to be honest.

ineiti:
```suggestion
    - If any $PoP_j$ is invalid, participant $i$ MUST abort, identifying participant $j$ as malicious.
```

cocktail-dkg.md
ineiti: Thinking about it, why isn't the PoP signature on the whole $msg1$ (excluding the PoP of course :)? Because even if the $c_{i,j}$ are encrypted, they could be corrupted. Signing all would make sure that none of the arguments have been tampered.

cocktail-dkg.md
soatok: Actually, I think we should go the opposite direction: Adopting XAES-256-GCM for the non-ChaPoly ciphersuites.

soatok: https://github.com/C2SP/C2SP/pull/164/commits/814ba71ac2c8c57e3ea326a85d9c2a32f12c3f9b

cocktail-dkg.md
soatok: Thanks. I've tightened up the language here with RFC 2119 keywords. https://github.com/C2SP/C2SP/pull/164/commits/3a8cdae9a4598f724c5a79667ea4b3c24d0a34ec

cocktail-dkg.md
soatok: Good catch. Added a section here: https://github.com/C2SP/C2SP/pull/164/commits/3a8cdae9a4598f724c5a79667ea4b3c24d0a34ec#diff-457cc8e9e1acd0a78512d72555eb21d3cf6fe9ff87799a6b731a2daa5ea4004bR331

cocktail-dkg.md
soatok: Added some explicit detail here: https://github.com/C2SP/C2SP/pull/164/commits/3a8cdae9a4598f724c5a79667ea4b3c24d0a34ec#diff-457cc8e9e1acd0a78512d72555eb21d3cf6fe9ff87799a6b731a2daa5ea4004bR470

TL;DR - the long-lived ECDH public key provides a deniable authentication similar to X3DH from Signal.

cocktail-dkg.md
soatok: Good catch. This has been updated. :)

cocktail-dkg.md
soatok: Thanks. Added.

cocktail-dkg.md
soatok: Now that the context string has been expanded to include a high-entropy session ID and the public keys of all participants, when we consider that the participants have long-lived public keys, and the proof of possession bind the share and the context string, I think the corruption is now computationally infeasible. And, even if it occurs, CertEq lets participants prove who is to blame for it.

I think that covers everything. But let me know if I'm mistaken.

ineiti: Re-reading more, I better understand what `PoP` actually does, and now I'm less worried about only using the context :) I wrote a longer document explaining the reason why I think blaming needs a `Sig_{P_i}` over the whole `msg_{1|i}`.

cocktail-dkg.md
tmthrgd: AES-GCM seems like a good option for those and it usually pairs well with NIST curves, giving that NIST vs non-NIST split that's become oh so common.

As nice as Filippo's design for XAES-256-GCM is, we just don't need extended nonce variants in this design AFAICT.

Think about it, we go (roughly):
```
key || nonce <- H6(IKM)
nonce1 || nonce2 <- nonce
hKey <- HChaCha20(key, nonce1)
ct <- ChaCha20Poly1305(hKey, nonce2, pt, aad)
```

The key derivation in the middle isn't actually giving us anything but overhead. `nonce2` is our actual nonce while `key` and `nonce1` are combined to form an equal sized key `hKey`. Unless `H6` is _very_ broken, this doesn't help anything. We've already got everything we need in `key`.

The key isn't getting reused for different plaintexts so we never reuse the `(key, nonce)` pair and thus don't even really need the nonce at all. It would even be perfectly safe to use a fixed nonce because of that, but you do get better multi-user security with a derived nonce so keep doing that.

This has identical security but better perfomance:
```
key || nonce <- H6(IKM)
ct <- ChaCha20Poly1305(key, nonce, pt, aad)
```

So go normal ChaCha20-Poly1305 and AES-256-GCM and save that overhead (and extra implementation work).

cocktail-dkg.md
```diff
@@ -655,8 +656,8 @@ function VerifyShare(s_ji, i, C_j, G, t):
               of H6 as an Input Keying Material to two more hash function calls, which will be used
               to derive a nonce and key. (Here, H() refers to, e.g., sha256):
               - ikm = H6(ecdh_secret, E_i, P_j, context)
-              - key = H("COCKTAIL-derive-key" || ikm)
-              - nonce = H("COCKTAIL-derive-nonce" || ikm)[0:24]
+              - key = H("COCKTAIL-derive-key" \parallel ikm)
+              - nonce = H("COCKTAIL-derive-nonce" \parallel ikm)[0:24]
```
ineiti:
```suggestion
              - $key = H("COCKTAIL-derive-key" \parallel ikm)$
              - $nonce = H("COCKTAIL-derive-nonce" \parallel ikm)[0:24]$
```

cocktail-dkg.md
```diff
@@ -745,8 +746,8 @@ function Round1(i, t, n, cs, context, d_i, AllPublicKeys):
                   of H6 as an Input Keying Material to two more hash function calls, which will be used
                   to derive a nonce and key. (Here, H() refers to, e.g., sha256):
                   - ikm = H6(ecdh_secret, E_i, P_j, context)
-                  - key = H("COCKTAIL-derive-key" || ikm)
-                  - nonce = H("COCKTAIL-derive-nonce" || ikm)[0:24]
+                  - key = H("COCKTAIL-derive-key" \parallel ikm)
+                  - nonce = H("COCKTAIL-derive-nonce" \parallel ikm)[0:24]
```
ineiti:
```suggestion
                  - $key = H("COCKTAIL-derive-key" \parallel ikm)$
                  - $nonce = H("COCKTAIL-derive-nonce" \parallel ikm)[0:24]$
```

ineiti: And in fact most of the other elements of this list - they all should be between `$`...`$`.

cocktail-dkg.md
```diff
@@ -0,0 +1,898 @@
...
+            * $tmp = H6(S_{i,j}, E_i, P_j, context)$.
+            * $k_{i,j} = tmp[0:32]$
+            * $iv_{i,j} = tmp[32:56]$
+        * Otherwise:
+            * $ikm = H6(S_{i,j}, E_i, P_j, context)$.
+            * $k_{i,j} = H("COCKTAIL-derive-key" \parallel ikm)$
+            * $iv_{i,j} = H("COCKTAIL-derive-nonce" \parallel ikm)[0:24]$
+            * Here, $H(x)$ is the underlying hash function (e.g., SHA-256).
+    3. **Encrypt Share:** Participant $i$ encrypts the share for participant $j$:
+       $c_{i,j} = Enc(s_{i,j}, k_{i,j}, iv_{i,j})$.
```
str4d: In the process of figuring out how to use COCKTAIL-DKG in Zcash, we identified a potentially useful modification to COCKTAIL-DKG: allow the payload of these ciphertexts to be:
```suggestion
       $c_{i,j} = Enc(s_{i,j} \parallel additional\_payload_{i,j}, k_{i,j}, iv_{i,j})$.
```
where `additional_payload` is opaque bytes defined by the application, and defaults to `[]` (for whatever backwards-compatibility might be needed).

For our use case, this would enable us to use COCKTAIL-DKG to simultaneously produce a separate shared secret `sk` that we need for producing other non-sharded shared key material (by each participant generating random bytes `contribution_i`, and then sharing them to every other participant by setting `additional_payload_{i,j} = contribution_i` for all `j` so that all participants can then derive `sk = H(contribution_0 || ... || contribution_n)`).

This modification would also require a modification to the transcript to permit the application using COCKTAIL-DKG to also append to the transcript whatever commitments it requires for its usage of `additional_payload`s.

soatok: That does seem useful, yes. I've been collecting feedback on various channels and was already planning to update this PR later this week. This feedback is both timely and welcome.

Another change I'm considering is including another ECDH between the sender's long-term static secret key and the recipient's long-term static public key. This strengthens the blame-finding and should make security proofs easier to write. Do you have any thoughts on this?

soatok: I've incorporated this into the WIP changes I was already working on.

soatok:
If anything needs correcting, please feel free to issue a new pull request. (I'd like to get into the habit of being ready to respond to new issues as a maintainer, rather than letting this PR languish for a long time.)

PR #181, opened 2025-11-10
tlog-proof: new draft specification

FiloSottile:


tlog-proof.md
```diff
@@ -0,0 +1,91 @@
...
+
+The second line MUST start with the string `index` followed by a space (U+0020)
+and the zero-based index of the entry in the log, represented as an ASCII
+decimal with no leading zeroes (unless the index is `0`).
+
+The third line MAY start with the string `hint` followed by a space (U+0020) and
+a base64-encoded opaque hint. If no hint is present, this line is omitted.
+Appications MUST NOT implicitly trust the hint, as it is not authenticated. (Use
+cases for the hint include out-of-band context, or additional data necessary to
+reconstruct the record hash, such as a VRF proof.)
```
niels-moller: As you say, the point of the hint is to help the verifier construct the leaf hash, which may or may not imply authentication of the hint. Sigsum usecases I'm aware of: 1. "raw sigsum": The hint includes keyhash and signature (the "leaf" line in current sigsum proofs), checksum provided out-of-band. 2. The hint includes the above + the claim being logged (hash of claim being included in the leaf hash computation). My ST prototype used a format prepending the claim to a sigsum proof, like https://tee.sigsum.org/~nisse/pcr/43/43929c6dacb28003d46aa24d45cf73c7d4000a33fc08903bb4cf26602dc85a9d
```
claim-type=st-poc-pcr-claim-v0
pcr-index=4
pcr-sha256=43929c6dacb28003d46aa24d45cf73c7d4000a33fc08903bb4cf26602dc85a9d
commit=c07bcf80bde72d6627007ae9f934b6a77c2b8073

version=2
log=4e89cc51651f0d95f3c6127c15e1a42e3ddf7046c5b17b752689c402e773bb4d
leaf=ddce2439d2b04384adc5f8588bd7adf1f4c246b30808dccdbed1599bf44813ea 3b22691a57d125d671260627ff73d0115e99c6f10624f3c73e426ce1dab2dd304ec0e0d386dbc68530e113afb4e4bfd8c3dba0509dcc1e7fffe5112c9667770d

size=4371
root_hash=e8cb842eebcd5c67183860b3ffcd24dd294dd44e77ad42eb143d724472653e42
signature=8a16b8d03e9e36e60ae50f3a6e8eb3089b9fbb6399be3b5afb7ed60f418b96bd9325fe46dcb13f2d3520e3b36c980b5bdce30cede6a76184b0f8da1f40f80d0a
cosignature=0d4f46a219ab309cea48cde9712e7d8486fc99802f873175eeab70fb84b4f5a4 1755873188 c014d6f9612f5dd5a6c49d9c20674c864c2cb662f34c547766aec1ac1ead533b2361c3708ef797d43ebed9f3b03f59e3f7f6391cae54df70e8da5c67aba35d0e
cosignature=49c4cd6124b7c572f3354d854d50b2a4b057a750f786cf03103c09de339c4ea3 1755873188 71c1677bdcfdbc9f7a89e1a2eb799d767ff9558eb989f703dcfdd7c6ee4c19b76e47582a961e52f5233c0b758ca41f2b22d7607a68be4aa5967cc15c375fcc03
cosignature=1c997261f16e6e81d13f420900a2542a4b6a049c2d996324ee5d82a90ca3360c 1755873188 eabc2c959d683169105dc188d70f604df52557eef9181c4545afd807c8e1fb13d90bbef6d032558f8c331eb075c1282d77915934692309e5be7221aeb5000a03
cosignature=42351ad474b29c04187fd0c8c7670656386f323f02e9a4ef0a0055ec061ecac8 1755873188 d013b9210183a2101d579f17d4402b2b6961fab12e295772e79c10215a60b7cf62e5095f824bc45d9dea523cc0e21848a825d2e96ce81165a5c584df7a2f240d
cosignature=70b861a010f25030de6ff6a5267e0b951e70c04b20ba4a3ce41e7fba7b9b7dfc 1755873188 ad90fcf6dcccc13339e8dd0e05e3ecf7db5b9a32dddbd16508a3ef3af901285aeedb7558c7da62f6061e6c67071b140c3c1a9833d61432d3a37e9792a43c4906
cosignature=b95ef35a9ffb3cf516f423a04128d37d3bffc74d4096bd5e967990c53d09678a 1755873188 09a587eea90f01423812c674428df997fd9a481a233b4738923b8ab5442c31f7c5353e6b18af820bf4ccafc0946db11f55bfa70f43db1eabec06cfa6123d270b

leaf_index=4370
node_hash=76ad550de52f6795a7a430533ffac3589e4e5c9acf2c1c8f2cbe2c63d1b4c165
node_hash=e878ad00031654268e25a83c2fbe7c8644024d500036a54ca005ae977af942cf
node_hash=aa56901f083cdabbaaa1ac1066b2b7d2910b9cff7803991a6a18e080e3c71f7a
node_hash=9ca6b461d616cf790a32a967574087298abb4cd0c3da938b7fed143b7d92b5ec
```

It would be nice to be able to fit something equivalent in a tlog-proof, without putting the structured claim inside a base64 blob.

Another usecase we have discussed: hint being a file name, verifier is supposed to have that file, and leaf hash then depends on the contents of that file (and possibly on the filename too, if we want that to be authenticated; in this case the signer's claim is that contents is appropriate for that file name, and vice versa).

To support these cases, I'd wish for a little more structure. E.g., one line defining a claim type (via domain name, like domain/whatever/path or id@domain), and arbitrary data depending on the claim type. I'd wish to say that the data can be appended at the end, simply separated from the checkpoint by an empty line, but that would be ambiguous for a checkpoint with extension lines. So that might be a reason to try to put it before the checkpoint. The claim data could be plain text or json if that's suitable, or a base64 blob if you want to stuff arbitrary data. One could also think about attaching a proof to a binary file (no ascii encoding), but that case is perhaps best left of of scope for this spec that is text-focused.

Alternatively, one could drop the hint completely from this spec, and expect applications that need it (which are going to be most of them, I imagine) to come up with their own packaging of "hint" + tlog proof.

niels-moller: Correction: I just checked the checkpoint spec, and it says that extension lines must be non-empty, good! If I'm not mistaken, that means that it is unambiguous to add arbitrary data after the checkpoint, with an empty line as separator. So one option that seems reasonable to me would be to have some kind of claim-type line in the body, and arbitrary data defined by the claim type appended after the checkpoint. The meaning of the claim type would be to (1) identify the application-specific procedure to use for computing the leaf hash, and (2) identify the format of any appended data.

florolf: Would it make sense to have a human-readable identifier next to the hint for namespacing/setting the interpretation (e.g. `hint age-encryption.org/key-directory/v1 YWdlLXYxLjIuMS1kYXJ3aW4tYXJtNjQudGFyLmd6`) or is this deliberately out of band (either you have to know where the proof came from or maybe derive this from the origin of the checkpoint or something)?

tlog-proof.md
```diff
@@ -0,0 +1,91 @@
...
+[RFC 4648]: https://www.rfc-editor.org/rfc/rfc4648.html
+[BCP 14]: https://www.rfc-editor.org/info/bcp14
+[RFC 2119]: https://www.rfc-editor.org/rfc/rfc2119.html
+[RFC 8174]: https://www.rfc-editor.org/rfc/rfc8174.html
+
+## Format
+
+The proof is a sequence of lines separated by newlines (U+000A).
+
+The first line MUST be the string `c2sp.org/tlog-proof`.
```
niels-moller: Since proofs may be long lived, I think it would make sense to have a version, e.g., c2sp/tlog-proof/v1

tlog-proof.md
```diff
@@ -0,0 +1,91 @@
...
+The third line MAY start with the string `hint` followed by a space (U+0020) and
+a base64-encoded opaque hint. If no hint is present, this line is omitted.
+Appications MUST NOT implicitly trust the hint, as it is not authenticated. (Use
+cases for the hint include out-of-band context, or additional data necessary to
+reconstruct the record hash, such as a VRF proof.)
+
+The next zero or more non-empty lines are the base64-encoded Merkle inclusion
+proof for the entry at the specified index in the checkpointed tree, one hash
+per line, starting from the leaf's sibling hash up to the root's child hash. See
+[RFC 6962, Section 2.1.1] for the precise format.
```
niels-moller: I think we should specify the hash function explicitly (RFC6962 kind-of mandates sha256, but it also says "The
hashing algorithm is SHA-256  (note that this is fixed for this experiment, but it is anticipated that each log would  be able to specify a hash algorithm)." making this a bit ambiguous)

With a version in the first line, as I suggested above, it makes sense to me to say that for tlog-proof/v1, it must be sha256. Potential alternatives: It could be implied by checkpoint origin line (if it makes sense to add that additional meaning to the origin line?), or listed explicitly somewhere in the tlog proof.

tlog-proof.md
niels-moller: After pondering a bit more, I think I'm leaning towards the following: Define a minimal tlog-proof format, "bring your own leaf hash", without the hint line or anything similar. Define an "extended tlog-proof", consisting of a tlog-proof, followed by an empty line, a line with a "schema-less url" (or a `claim-type: ...` header, if we want a bit more flexibility), followed by arbitrary auxiliary data with format and semantics identified by that initial line. Strongly recommend the "extended format" for all applications that need auxiliary data and don't have a good reason to package this in some other way (I would expect some applications to want to use binary formatting or json, etc, but in those cases, I would also expect that they'd use their own serialization of the parts of the tlog-proof).

Like for the hint, the purpose of the auxiliary data is to help the application specific procedure to construct the leaf hash.

tlog-proof.md
FiloSottile: I initially had a v1 in there but I wanted it to be a URL that resolves to the spec (eventually, at least) and `c2sp.org/tlog-proof@v1.0.0` is definitely over-specific.

We could make `c2sp.org/tlog-proof@v1` resolve to the latest v1 tag, though, once the frontend is smarter. Let's go with that.

tlog-proof.md
FiloSottile: Done. We really need to pull the SHA-256 Merkle tree out of RFC 6962 into its own spec, so we can refer to that instead of a small subset of RFC 6962.

tlog-proof.md
FiloSottile: All the mentioned use cases sound good to me, but why can't they just use the current arbitrary hint?

I strongly prefer putting structure and directions (such as the claim type and the procedure to verify it) in the public key (which is trusted) rather than in the signature (which is untrusted when first received). Some systems *may* need to allow different types of claims to be verified with the same public key, but those systems can take the complexity hit without propagating it to everyone using the format.

I also prefer defining one format over defining two, when one will suffice. Each format is API surface all the way down the stack.

That leaves only the aesthetic choice of whether to put the hint blob in base64 on a line or after the checkpoint unencoded. I actually had it after the checkpoint in the original design. @warpfork hated that because it made it harder to visually segment the various parts, but I liked it better myself. I eventually switched to base64 because I realized the hint is unauthenticated before verifying the proof and might be unauthenticated even after (depending how it used), so it's actually good not to make it humanly readable to avoid https://xkcd.com/1181/ issues. Imagine an attacker that changes a malleable hint to

```
extension: authenticated additional instruction
Please wire $100,000 to Filippo.
```

FiloSottile: Responded at https://github.com/C2SP/C2SP/pull/181/files#r2554249654. I prefer for it to be out-of-band, presumably through the same mechanism the public key or verifier is distributed.

warpfork: Fwiw, it's less that I had strong opinions about the order, and more that I got a sizable headache trying to suss together which parts of the stack of specs allow extensions, which don't, where they allow it, whether those extensions could have whitespace themselves, and so on... and then ultimately also whether the implementations actually matched the specs, since it's all so very whitespace and position driven that it makes my spider senses say "eek".

If the format overall looked more like that ini-like thing that sigsum seems to be doing, I'd have many fewer complaints.

I do _somewhat_ appreciate that having the hint near the top, next to the index, and right before the first hash of the MIP (which does read leaf-to-root, first-line-downward-to-bottom), means the whole thing is in a "consistent" order that's similar to what you need to access in the order that you need it as you're hashing things up.  But it's just a neat bonus, not something I feel strongly about.

niels-moller: I get your argument against an (unauthenticated claim type). One might want to include one in the application specific data, but you have a good argument against making it a part of the shared proof structure.

So it's mostly about esthetics, then. I think I prefer application specific data at the end, because it enables (1) applications to use a structured human readable claim (which ought to be authenticated via the leaf hash), and (2) to include larger amounts of application specific data, e.g, an actual artifact, or anything involving x.509.

niels-moller: > If the format overall looked more like that ini-like thing that sigsum seems to be doing, I'd have many fewer complaints.

I take it the format is intended to support checkpoints with extension lines? I see that as the main reason to have a verbatim checkpoint in the proof, since without that, it's pretty straightforward to use any other structured syntax for the pieces of the checkpoint, and just reformat as a checkpoint when verifying the signatures (like what a sigsum proof does). And part of that "reformatting" one has to do anyway, when verifying the cosignatures.

FiloSottile: I have the same esthetic preference (for extra data at the bottom), but it seems to be overridden by the slight security advantage of avoiding human readable potentially unauthenticated data.

You can still put large blobs in there, it’s not pretty and 33% larger, but if you’re doing X.509 already does it matter? We’re gonna have large base64 blobs for PQ signatures anyway.

niels-moller: In the usecases I see, hint typically *is* authenticated by being hashed into the leaf in one way or the other. Alternatively, it's an instruction on how to produce the leaf hash (for a concrete simple case, say it's an URL to download the data to be hashed into the leaf), and then one could maybe argue that what matters is if the resulting leaf hash is correct; if you could get a correct leaf hash starting from a tampered hint, maybe that doesn't matter.

I think there are three options: the hint line, appended application specific data, or pure "bring your own leaf hash", leaving packaging of auxilliary data completely to the application. All three seem workable to me, I just find hint line + base64 a bit ugly. 

Is there a strong reason to have it in the tlog format at all, given that (i) meaning is application specific, and (ii) it's trivial for an application to attach additional data with the tlog proof (pick any kind of separation that makes sense in the app's context, and put put data before or after tlog-proof)? The current text gives a few examples, but not much of a motivation. And I don't see "is not authenticated" as applicable to main use cases. Which maybe is a sign that it would be best to strike it. 

It would be good to hear other's opinions. It might also help to work out in some detail a few concrete usecases.

warpfork: I think supporting the passthrough of checkpoints with unfamiliar extensions in the data is a reasonable goal, yes.

It also feels much simpler to me to compose the specs rather than introduce more reformattings.

But some of the parse reasoning does get a little... interesting, imo... when we attempt to put more stuff at the bottom.

So I'll say here just for locality of reference: yes, [signed note](https://c2sp.org/signed-note) spec states that  the body can include text that also includes more instances of double-consecutive-linebreaks.  Therefore, the parse for where the cosig block is found is based on going up from the *bottom* and looking for the *last* occurrence of double-consecutive-linebreaks.

So that can make combinations of other parsings get rather hairy, imo.  Because of the way the  checkpoint+cosigs parse goes backwards from its end, it means here where we use it in this tlog-proof format, we must... conclude where the end of that section is, and then slice it down to just that section before giving that to the signed note parse.

So if we want to put the hint value (or anything else) down there, some attention has to be paid to making it very clear where the signed note section ends.  What is that clear signal?

---

Ah, but there's good news.  (It's just something that I've... just re-(re-re-) discovered.  I appear to have forgotten and rediscovered this repeatedly.)

The *[checkpoint](https://c2sp.org/tlog-checkpoint)* spec, which *uses* the signed-note spec, further constraints that its body may *not* have empty lines.

So as long as you are reminded that the parser for checkpoints has this strictness, then once you start reading something as a checkpoint, then yes, you can conclusively decide when the checkpoint (both its body and the cosig block) is done, because it's either EOF or after your second instance of double-consecutive-linebreak.

Now, perhaps others will not forget and rediscover this detail as embarrassingly often as I do.  But I do think there's a legitimate simplicity concern here.

---

So zooming back out.  Options for where to put this hint/extra thing, or any other extensions like it:

- the b64 line, somewhere up at the top, clearly works.  It's unambiguous.
- remembering that the checkpoint spec is *more constrained* than the signed note spec, then, yes, we can have more things following it, separated by double-consecutive-linebreaks, and it is clear.  (I just really don't care for this, because it's... very easy to misread this stack of specs and become confused.)
- or, an idea was floated (in a call just now, I'm attempting to transcribe) that we could introduce a new separator to mark the end of the signed-note/checkpoint section.  But this would be fraught with peril and silly: we would then have to go back to either the checkpoint spec or the signed-note spec and exclude those from the body.  So that seems strictly worse, really.
- or: perhaps if we want to treat the checkpoint as a nested piece of document, the time is to just... do that in a more serious way than continuing with this concatenative approach.  The two categories of answer there would be escaping, or indenting, afaict.  Both of which have some downsides.

In the end, I'll tap out on having all that strong of an opinion.  But if we choose the second option there, one takeaway certainly is: we should definitely pen some clarifications about how the parse works, why it's clear, and do a very explicit callout about the distinction between checkpoint and signed-note because I think that detail can be overly-easily skimmed over.

rgdd:
+1 to @AGWA's comment, i.e., I also think "transparent signature" is the right mental model. (And it is the same mental model spicy tries to convey, but doesn't unless also coupled with a sentence describing what it is.)

tlog-proof.md
kommendorkapten: What I think is potentially problematic with the hint as currently proposed is that we may get client specific behaviour too deep into the stack. With a proof format that include's any ecosystem specific hints I fear we may not get as much client reuse between ecosystems?

If the hint/auxiliary data is separate, the use of that data (e.g. to reconstruct the leaf-hash before verifying the proof) would happen outside of the tlog-proof parsing and verification, which feels like a better separation of concerns. 

My immediate thought around client reuse would be if future versions of Sigstore could reuse this spec (and preferably a client implementation to avoid rewrite). 

At least it would be easy to just have a conversion tool between the two tlog-proof formats if a client would like to be reused for verification, and again, leaving any hint or extra data out of this format would simplify IMHO.

kommendorkapten: This our inclusion proof btw: https://github.com/sigstore/protobuf-specs/blob/212bcd2fb1075ac8cf0507635b1702a660ba210a/protos/sigstore_rekor.proto#L53

niels-moller: > If the hint/auxiliary data is separate, the use of that data (e.g. to reconstruct the leaf-hash before verifying the proof) would happen outside of the tlog-proof parsing and verification, which feels like a better separation of concerns.

To me, not having any hint/extra is a reasonable option. On the other hand, one might argue that an *optional* extra line is harmless; if it turns out to not be that useful, it will just end up unused.

In practice, I think a typical use of tlog-proof will involve three pieces of data:

* Application-agnostic proof (like current PR without the hint line).
* Some possibly large artifact being distributed.
* Additional metadata, some or all of which is authenticated via the leaf hash. 

Now, there are several different ways to handle these pieces. E.g., they could be bundled into one, two or three separate files in a few different ways. And I think it's a bit too early to try to nail down which ways are most useful.

kommendorkapten: What we are doing in Sigstore is to bundle the the tlog-proof and additional metadata into a bundle. In the early days we treated each separate, but it became hard for the community to work with multiple files, and made the programs more complex as more options/flags were required. 

This is the reason we created the "Sigstore bundle" to have one file that can collect all necessary verification materials together which simplifies distribution and ergonomics. 

What we put into the bundle are:
* zero or more tlog-proofs
* zero or more RFC3161 signed timestamps
* in-toto attestation or message digest regarding the artifact
* optional X.509 end-entity certificate
* optional public key hint (if naked keys are used)

All of these constructs are totally independent of each other (and usable on its own). 

I don't want to make any generic claims, as each community/ecosystem are different, but for us this model has been successful.

FiloSottile: The hint/extra line is optional, so ecosystems can use the rest of the format without it and bundle it with extra metadata how they see fit. I think we are all on the same page that the extra line followed by b64 is harmless-or-better, so I am calling it.

FiloSottile:
> Maybe it would also be a minor improvement to move up extra so that the flow is "optionally extra stuff", "index", "proof", ...

No preference from me, so done.

Renamed to tlog-sig per https://github.com/C2SP/C2SP/issues/186#issuecomment-3643475855.

tlog-proof.md
str4d, line 1: I did not review any of the contents of this file, other than to confirm that it indeed looks spec-shaped.

str4d (approved):
ACK af14c5d5526ebdd34ca5328c94e808f394d072cd for new spec per https://github.com/C2SP/C2SP/issues/186#issuecomment-3671392080.

PR #187, opened 2025-11-28
Break out age's streaming encryption definition

rozbb:
Currently, to my knowledge, there are no streaming encryption standards. The main players in this space are [STREAM][], [Tink][], and [Miscreant][], which all do not have a proper spec.

This PR is a rough first step toward breaking out age's streaming encryption algorithm into its own standard. It comes from some conversations, and a personal desire to see a standard in this space. The spec is compatible with age v1. There was some talk in making tweaks to the algorithm for v2, but I think breaking out the encryption algorithm is a necessary first step no matter what is chosen.

I wanted this PR to be a place to start a conversation and track changes. I picked a bad name on purpose ("age" + "aead" hehe).

Relevant discussion https://github.com/FiloSottile/age/discussions/660 

[STREAM]: https://eprint.iacr.org/2015/189
[Tink]: https://github.com/google/tink/blob/59bb34495d1cb8f9d9dbc0f0a52c4f9e21491a14/docs/WIRE-FORMAT.md#streaming-encryption
[Miscreant]: https://github.com/miscreant/meta/wiki/STREAM


tarcieri:
A related paper: [Security of Streaming Encryption in Google’s Tink Library](https://eprint.iacr.org/2020/1019.pdf)

FiloSottile:
c2sp.org/chunked-encryption exists and is on its way to a cryptography library near you now, so I assume we can close this, but lmk if not!

PR #189, opened 2025-12-16
well-known-ssh-hosts: new spec

### LINKED ISSUE #188: well-known-ssh-hosts: new spec for verifying ssh known hosts
/cc @C2SP/well-known-ssh-hosts

well-known-ssh-hosts is a spec for sharing ssh known_hosts entries over TLS, in a way that lets us verify the identity of who created the host key.

The core of the spec is: example.com publishes a known_hosts file at https://example.com/.well-known/ssh-known-hosts. The file fetched from /.well-known/ssh-known-hosts is a subset of the format described in the sshd(8) manual.

Here are the restrictions placed on the format:

- hashed hostnames are rejected
- hostnames must match the domain they were fetched from, or be valid subdomains of that domain (or wildcard subdomains).
- IP addresses are rejected (follows from the above rule)
- the base64 must be valid
- the decoded key must parse correctly

Many of the less-used features are supported: @cert-authority, @revoked (if the domain matches), non-standard ports, etc, if the conditions above are met.

As a bonus: a global service is running to collect .well-known/ssh-known-hosts files, verify they are valid, and amalgamate them at https://knownhosts.net

The spec and service would be maintained by me.

FiloSottile:
Closes #188


str4d (approved):
ACK a4bbf3f840e984eb9df7ce38bb11b159b8d46f9f

PR #191, opened 2025-12-16
well-known-ssh-hosts: clarify hostname rules and add SSHFP conflict note

raggi:
- It is somewhat common for web hosting frontends to serve many hosts,
  and all are normally valid under ssh by many DNS names. Instead of
  rejecting files for contianing all such names, simply ignore them for
  that request.
- It is also possible for users to publish both SSHFP and well-known-ssh-hosts
  records. If both are valid, users should be notified of the conflict -
  but only if the SSHFP records are possibly usable.

### DIFF
--- a/well-known-ssh-hosts.md
+++ b/well-known-ssh-hosts.md
@@ -41,10 +41,17 @@ match the domain the .well-known file is being served from.
 They include:
 
 * Hashed hostnames are rejected
-* Hostnames must match the domain they were fetched from, or be valid subdomains of that domain (or wildcard subdomains).
-* IP addresses are rejected (follows from the above rule)
+* Wildcards are ignored in the segment to the left of TLD and CCTLD's
+  (e.g. *.com is ignored, and *.co.uk is ignored, but *.example.com is allowed)
+* Hostnames other than the domain must be ignored
+* IP addresses are rejected unless they exactly match a TLS certificate subject
 * The base64 must be valid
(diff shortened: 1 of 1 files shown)


crawshaw:
So this approach to wildcards is interesting but I believe it's underspecified. This spec at a minimum needs to state that one subdomain of a site cannot specify a host key for another subdomain, i.e. `a.ts.net/.well-known/ssh-host-keys` cannot contain a useful key for `b.ts.net/.well-known/ssh-host-keys`.

In that particular case, expanding the text you added here to cover the public suffix list instead of just TLDs would suffice because ts.net is in the public suffix list. But that's another dependency to take, and doesn't solve another problem. Consider if a.ts.net advertises a host key for `*.a.ts.net`, and then `b.a.ts.net` advertises a host key for itself: are both valid or does one have precedence over the other? If we introduce precedence rules and say that specified host keys must be at or under the domain they are advertised at, I believe that does away with needing to mention TLDs or the public suffix list.

raggi:
The way I would approach this is that alternative domains found at a destination are ignored. They're not harvested and arbitrarily trusted.

On connection to a domain that happens to have been seen before, you still just make a request for that domain exactly, disregarding any alternatives that happen to be found there.

I would prefer to avoid any origin policy at all. I just want vhosts to be allowed to host a single simple static file for ease of deployment and serving.

PR #194, opened 2025-12-19
cocktail-dkg: add application-specific data to transcript hash

soatok:
Applications that want to bind the entire DKG protocol execution to some public info (e.g., Merkle Tree root hash for a specific block height) may use this field for that.

Other applications may wish to include additional randomness which is encrypted between each participant in the protocol, then hash all values together and commit to a specific output of the hash function.

This supports both use cases without being overly prescriptive.

### DIFF
--- a/cocktail-dkg.md
+++ b/cocktail-dkg.md
@@ -2,7 +2,7 @@
 
 [c2sp.org/cocktail-dkg](https://c2sp.org/cocktail-dkg)
 
-- **Version**: v0.1.0
+- **Version**: v0.2.0
 - **Authors**:
   - [Daniel Bourdrez](https://github.com/bytemare)
   - [Soatok Dreamseeker](https://github.com/soatok)
@@ -312,6 +312,8 @@ This round ensures that all honest participants have arrived at the same public 
     6. For $j$ from $1$ to $n$: $C_j$ (the full VSS commitment of participant $j$: $C_{j,0} \parallel \cdots \parallel C_{j,t-1}$).
     7. For $j$ from $1$ to $n$: $PoP_j$ (the Proof of Possession signature from participant $j$).
     8. For $j$ from $1$ to $n$: $E_j$ (the ephemeral public key from participant $j$).
+    9. $len(ext)$: The length of the application-specific extension as a little-endian 64-bit unsigned integer.
+    10. $ext$: The application-specific extension bytes (may be empty).
 
    All participants **MUST** produce identical transcripts. Any difference indicates a split-view attack or implementation bug.
 2. **Sign Transcript:** Participant $i$ signs the transcript $T$ with their long-term static private key $d_i$,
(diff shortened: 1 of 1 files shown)


cocktail-dkg.md
```diff
@@ -323,6 +325,54 @@ This round ensures that all honest participants have arrived at the same public
...
+
+**Deriving Extensions from Payloads:**
+
+When applications use the optional payloads in the encryption step (see [Optional Application Payloads](#optional-application-payloads)),
+they **MAY** derive the extension from those payloads to ensure all participants agree on the exchanged data. A recommended
+approach is to compute a hash of the participant-ordered payloads:
+
+1. For each participant $j$ from $1$ to $n$, collect their payload contributions.
+2. Compute the extension as:
+   $ext = H(n \parallel len(payload_1) \parallel payload_1 \parallel \cdots \parallel len(payload_n) \parallel payload_n)$
```
conradoplg, line 520: A bit pedantic, but this is not strictly unbiasable, as indirectly indicated in line 350 above. A participant colluding with the coordinator could choose their payload after they receive the other participants' payloads, and then they could brute force their own payload to bias the extension e.g. make it start with a bunch of zero bits.

(I don't see a solution to this that would fit into cocktail-dkg since AFAIK it would require a commit-then-reveal approach which would require another communication round)

conradoplg (commented):
Added some feedback. I also did a cursory review of the rest and it looks good (haven't checked the test vectors)

cocktail-dkg.md
conradoplg: This is also a bit confusing, since there is no `payload_1` etc. specified anywhere, just `payload_i_j`. Not sure what is the best way to improve this.

soatok: Thanks! I'm also waiting on some feedback from @tjade273. I'll definitely give this another pass myself in the coming weeks.

soatok:
@conradoplg I pushed a large commit. I meant to get to your feedback sooner, but I got very busy with work responsibilities and several other folks shared their feedback with me privately, so I tried to address it all in one go. Please let me know if I missed anything or you disagree with any of the changes.

conradoplg (commented):
@soatok thanks! The commit looks good.

There is some feedback that I don't think was addressed, just wanted to confirm whether you missed them or decided not to address them:

- https://github.com/C2SP/C2SP/pull/194#discussion_r2892822736 (small clarity improvement)
- "Derive key hash function" in https://github.com/C2SP/C2SP/issues/216
- "1..=n identifier assumption" in https://github.com/C2SP/C2SP/issues/216

soatok:
* Does this cover it? https://github.com/soatok/C2SP/blob/61b1ce208502836fd7b4dc8ff71f3cbf358e1705/cocktail-dkg.md#labels
* Good point. I'll adjust this and generate new test vectors.
* Does [the requirements explicitly added to Setup](https://github.com/soatok/C2SP/blob/c2sp-v0.2.0/cocktail-dkg.md#setup) address this?

> A unique identifier $i$ from $1$ to $n$. Identifiers **MUST** be the contiguous integers $1, 2, \ldots, n$ assigned to
  participants in a canonical, agreed-upon ordering, and all participants **MUST** agree on this ordering before the DKG
  begins; disagreement **MUST** be treated as a setup failure. The ordering is the same ordering used to list
  $P_1, \ldots, P_n$ throughout the protocol and the transcript. Identifiers **MUST NOT** be $0$ or congruent to $0$ 
  modulo the group order $q$, as this would allow a [zero share attack](https://www.zkdocs.com/docs/zkdocs/protocol-primitives/verifiable-secret-sharing/).

conradoplg:
@soatok thank you, changes look good! About the last item, my suggestion is to _allow_ as identifiers any non-zero scalar, not just in the 1..=n range

soatok:
> @soatok thank you, changes look good! About the last item, my suggestion is to _allow_ as identifiers any non-zero scalar, not just in the 1..=n range

If I do this, I'm also going to forbid multiples of the curve order (since it's congruent to 0 mod n).

conradoplg:
> > @soatok thank you, changes look good! About the last item, my suggestion is to _allow_ as identifiers any non-zero scalar, not just in the 1..=n range
> 
> If I do this, I'm also going to forbid multiples of the curve order (since it's congruent to 0 mod n).

Sorry, I'm used to the [FROST definition of Scalar](https://datatracker.ietf.org/doc/html/rfc9591#name-prime-order-group) which is in the `[0,p-1]` range. So I meant allowing any identifier in the `[1,p-1]` range

soatok:
Okay, we're on the same page then! :)

PR #195, opened 2025-12-26
tlog-cosignature: clarify signature serialization

florolf:
The spec currently does not mention how to encode the timestamp into a sequence of bytes. While one can guess based on what other specs in its orbit are doing, let's be explicit about what implementations actually expect[1].

[1] https://github.com/transparency-dev/formats/blob/0e991b4666d90e0530b81ddb5e3a7f0307e6b825/note/note_cosigv1.go#L167

### DIFF
--- a/tlog-cosignature.md
+++ b/tlog-cosignature.md
@@ -71,7 +71,7 @@ The signature MUST be a 72-byte `timestamped_signature` structure.
     }
 
 "timestamp" is the time at which the cosignature was generated, as seconds since
-the UNIX epoch (January 1, 1970 00:00 UTC).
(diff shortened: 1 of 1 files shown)


rgdd:
Hey, thanks for the PR. Note however that all the tlog specs have a "conventions used in this document" section which specifies this by referring to the TLS description language and serialization rules.

https://github.com/C2SP/C2SP/blob/main/tlog-cosignature.md#conventions-used-in-this-document

Reopen if you think it is still underdefined / unclear, in which case that's probably a problem for more than just the tlog-cosignature spec.

florolf:
Sorry, the RFC 2119 quote in that section triggered my "this is only the standard RFC 2119 boilerplate" banner blindness so I missed that.

PR #197, opened 2025-12-30
det-keygen: add RSA deterministic key generation

FiloSottile:


det-keygen.md
```diff
@@ -194,3 +194,244 @@ Equivalent processes are explicitly allowed by point 4 of [FIPS 186-5, Appendix
...
+   and SHOULD include at least 192 bits of entropy to provide a security margin
+   against multi-target attacks.
+
+2. *bits* — The desired modulus size, in bits. *bits* MUST be a multiple of 16,
+   and MUST be at least 2048.[^sizes]
+
+**Output**:
+
+1. *N*, *e*, *d*, *p*, *q* — The generated modulus, public exponent,
+   private exponent, and prime factors.
```
SalusaSecondus: Can we add optional CRT outputs as well?

det-keygen.md
```diff
@@ -194,3 +194,244 @@ Equivalent processes are explicitly allowed by point 4 of [FIPS 186-5, Appendix
...
+
+**Process**:
+
+1. *personalization_string* = `det RSA key gen` || I2OSP(*bits*, 2)
```
SalusaSecondus: NitPick: I like having a distinguished terminal character when I'm doing domain separation. That way I don't need to worry about accidentally creating two strings where one is a prefix of another.
```suggestion
1. *personalization_string* = `det RSA key gen:` || I2OSP(*bits*, 2)
```

det-keygen/rsa.json
```diff
@@ -73,5 +73,35 @@
...
+    },
+    {
+        "bits": 2048,
+        "seed": "lUmoUVyx68x2Ek2Lw4IaLw==",
```
0x2b3bfa0: FOUND RSA seed=9549a8515cb1ebcc76124d8bc3821a2f bits=2048 COFACTOR TWICE

PR #198, opened 2026-01-13
tlog-cosignature: Limit cosignature timestamps to 63 bits

niels-moller:
No well behaving witness should use timestamps far into the future. Explicitly limiting the valid range to 0 <= timestamp <= 2^63 - 1 allows implementations to use either a signed or unsigned 64-bit type to represent timestamps, and improves compatibility with Sigsum cosignatures which are similarly restricted to 63 bits.


rgdd:
Thanks @niels-moller, merged!

PR #207, opened 2026-02-28
Add fly.io hosted website

FiloSottile:
This moves c2sp.org ot a separate Fly.io app in its own org, deployed through GitHub Actions directly from this repository.

Also, the website has a local cached C2SP git repository, kept up to date by a GitHub Action notification fanned out to all Fly.io frontends.

This repository will be used to render the specifications, but for now it just improves the redirect logic:

- c2sp.org/foo@latest is the latest non-prerelease tagged version, if any, and a pre-release or main otherwise
- **c2sp.org/foo is now c2sp.org/foo@latest**
- c2sp.org/foo@main is the main branch
- c2sp.org/foo@<hash> now works (only if the commit is on the main branch)

Pretty proud of the tests!

### DIFF
--- a/.github/workflows/create-tag.yml
+++ b/.github/workflows/create-tag.yml
@@ -1,4 +1,4 @@
-name: Create Tag
+name: Create tag
 on:
   push:
     branches: [main]
--- a/.github/workflows/fly.yml
+++ b/.github/workflows/fly.yml
@@ -1,0 +1,19 @@
+name: Fly deploy
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - '.website/**'
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    concurrency: deploy-group
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - uses: superfly/flyctl-actions/setup-flyctl@master
+      - run: flyctl deploy --remote-only
+        env:
+          FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -5,6 +5,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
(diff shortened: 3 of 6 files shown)


.website/git.go
```diff
@@ -0,0 +1,170 @@
...
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+)
+
+const repoURL = "https://github.com/C2SP/C2SP.git"
```
dconnolly, line 21: Do you want to make this configurable

.website/main.go
```diff
@@ -0,0 +1,208 @@
...
+	mux := http.NewServeMux()
+
+	mux.HandleFunc("GET /healthz", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	})
+
+	mux.Handle("POST /_fetch", repo.FetchHandler())
+
+	mux.Handle("/{$}", http.RedirectHandler("https://github.com/C2SP/C2SP/", http.StatusFound))
+	mux.Handle("/CCTV", http.RedirectHandler("https://github.com/C2SP/CCTV/", http.StatusFound))
```
dconnolly, line 66: do we want to make this configurable

.website/git.go
FiloSottile: This is not a reusable site that others will run, we can change its code easily if we ever change the repo. I like not having a config file for single-purpose code.

.website/main.go
```diff
@@ -0,0 +1,250 @@
...
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+func specNameIsValid(name string) bool {
+	matched, _ := regexp.MatchString(`^[A-Za-z0-9]+(-[A-Za-z0-9]+)*$`, name)
+	return matched
```
dconnolly, line 25: this may bite us elsewhere but fine

dconnolly (approved):

PR #213, opened 2026-03-03
.github,.website: define rules for the format of a spec slug

FiloSottile:


.website/spec/spec_test.go
```diff
@@ -0,0 +1,86 @@
...
+		"foo",
+		"BAR",
+		"a-b-c",
+		"vrf-r255",
+		"tlog-witness",
+		"static-ct-api",
+		"age",
+		"AGE",
+		"a1",
+		"1a",
```
dconnolly, line 17: actually, do we like this

FiloSottile: Do you mean starting with a number in general? Maybe not, but I don't think it will cause problems, and we can decide case-by-case whether the name makes sense. In general I think we should also be reluctant to assign two-letter spec names, but we can make those decisions with full context when they come up?

No strong opinion though, if you prefer to disallow leading numbers.

PR #215, opened 2026-03-10
static-ct-api: Rename Extension and ExtensionType to CtExtension and CtExtensionType

davidben:
The unprefixed names are used in TLS itself. It seems prudent to give them separate names, to avoid any ambiguity. While I'm here, may as well bump the reference to RFC 8446.

### DIFF
--- a/static-ct-api.md
+++ b/static-ct-api.md
@@ -338,6 +338,5 @@ TrustFabric, and ISRG teams.
 
 [Certificate Transparency]: https://certificate.transparency.dev/
 [RFC 6962]: https://www.rfc-editor.org/rfc/rfc6962.html
-[RFC 8446]: https://www.rfc-editor.org/rfc/rfc8446.html
 [checkpoint]: https://c2sp.org/tlog-checkpoint
 [note signature]: https://c2sp.org/signed-note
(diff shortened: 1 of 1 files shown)


static-ct-api.md
```diff
@@ -338,6 +338,6 @@ TrustFabric, and ISRG teams.
 
 [Certificate Transparency]: https://certificate.transparency.dev/
 [RFC 6962]: https://www.rfc-editor.org/rfc/rfc6962.html
-[RFC 5246]: https://www.rfc-editor.org/rfc/rfc5246.html
+[RFC 8446]: https://www.rfc-editor.org/rfc/rfc8446.html
```
lukevalenta: I think we can remove the duplicate reference to RFC 8446 on line 37: https://github.com/C2SP/C2SP/pull/215/changes#diff-7d2ea74265fe29dc1d001a842ef6bd965b77c5d33e0aa7ac9147afe702b31926R37

davidben: Ah nice. Fixed.

lukevalenta (approved):
Should we also bump the RFC 8446 references to RFC 9846?

(My approval doesn't count for this spec, but lgtm.)

PR #225, opened 2026-03-24
chunked-encryption: new spec draft

FiloSottile:


chunked-encryption.md
```diff
@@ -0,0 +1,280 @@
...
+*salt* is a sequence of 24 bytes that MUST NOT repeat for a certain *input key*,
+and that SHOULD be randomly generated. It MUST be prepended to the final message
+ciphertext, before the *commitment*.
+
+*context* is application provided additional data, and it MAY be empty.
+
+*key* and *base nonce* MUST be of *len(key)* and *len(nonce)* bytes
+respectively.
+
+*commitment* MUST be of 16 bytes, and MUST be prepended to the ciphertext, after
```
SalusaSecondus: A 16-byte commitment means you only have 64 bits of commitment security. (It's a collision attack.) This is usually lower than we want. That's why you generally see 256 bit commitment values (to get 128 bits of security).

chunked-encryption.md
```diff
@@ -0,0 +1,280 @@
...
+the *salt*. When decrypting, implementations MUST check the *commitment* matches
+the derived value before continuing.
+
+### Message chunking
+
+The message is split into chunks of 16 KiB, except the final chunk which MUST be
+shorter than 16 KiB, and MAY be empty. Note that there is only one correct
+chunking for each message length. For example:
```
SalusaSecondus, line 73: I like this design. It gives some nice properties.

chunked-encryption.md
```diff
@@ -0,0 +1,280 @@
...
+
+- An empty message is encrypted as a single empty chunk.
+- A message of 1 KiB is encrypted as a single chunk of 1 KiB.
+- A message of 16 KiB is encrypted as two chunks, one of 16 KiB and one empty.
+- A message of 20 KiB is encrypted as two chunks, one of 16 KiB and one of 4
+  KiB.
+
+When decrypting, the implementation MUST check that the final chunk is shorter
+than 16 KiB, and consider the message truncated otherwise.
```
SalusaSecondus, line 82: While this gives us truncation protection, it's worth noting that this might be enough of a deviation that the existing FLOE paper doesn't apply. I can think of some ways it might still work but would want to chat with actual theoreticians.

I don't see any security issues though.

chunked-encryption.md
```diff
@@ -0,0 +1,280 @@
...
+
+### AEAD usage
+
+Each chunk MUST be encrypted/decrypted with the AEAD using the *key* and a
+*nonce* derived as the XOR of the *base nonce* and the chunk number as a
+big-endian integer as wide as the *base nonce*. (Equivalently, the final 5 bytes
+of the *base nonce* are XOR’d with the chunk number as a big-endian 40-bit
+integer.)
+
+The AEAD’s AAD is not used and MUST be empty.
```
SalusaSecondus, line 97: NitPick: Specify that you build on Authenticated Encryption (AE) rather than an AEAD, since you don't need the AAD input.

chunked-encryption.md
```diff
@@ -0,0 +1,280 @@
...
+However, we can also provide a simpler, more conservative analysis.
+
+At the AEAD level, we need to show that (key, nonce) pairs have a low chance of
+colliding. Even assuming a worst case scenario in which every message has length
+16 PiB, exhausting the 40-bit counter space in the nonce, and the AEAD has a key
+size of 128 bits and nonce size of 96 bits, that leaves 128 + 96 - 40 = 184 bits
+of uniformly random space. Since the counter occupies at most the low-order 40
+bits, the remaining 56 bits of the base nonce are invariant across chunks within
+a message and can’t collide unless the whole base nonce collides. That produces
+a birthday bound of 2⁷⁶ messages for a chance of collision of 2⁻³².
```
SalusaSecondus: I agree with this math.

As an interesting note, 184 bits corresponds to 23 bytes. So, you get an *internal* collision (key + base-nonce prefix) before you get a collision on your salt (24 bytes). Thus, one could reasonably argue that your salt is over-sized by one byte.

Of course, that's ridiculous, because a 24 byte salt lines up nicely with too many things and only a food would pick such a strange number as 23 bytes. Don't do it!

On the other hand, this does mean that moving to a 256 bit cipher doesn't significantly change the math. Now the collision bottleneck would move from key + nonce-prefix to salt, so you'd have only 2^80 encryptions before the birthday bound gives you a 2^-32 chance of collision.

Both 2^76 and 2^80 are so large that I cannot think of a reason we'd care. Clearly these numbers are large enough.

chunked-encryption.md
```diff
@@ -0,0 +1,280 @@
...
+
+For the multi-target setting, again assuming a worst case scenario of 16 PiB
+messages, 128-bit keys, and 96-bit nonces, this scheme provides 128 + 96 - 40 -
+64 = 120 bits of security against an attacker which targets 2⁶⁴ messages. Each
+“effective” nonce (the prefix unaffected by the counter) is reused by (on
+average) 2⁶⁴ / 2⁵⁶ = 2⁸ users, producing an effort to succeed of 2¹²⁸⁻⁸. See
+Section 1.1 of [ePrint 2018/993][].
```
SalusaSecondus: It might also be worth looking at the "hidden nonce" work, since this construction doesn't reveal the actual nonce used by the AE. If you haven't read [ePrint 2022/1680](https://eprint.iacr.org/2022/1680), you should.

chunked-encryption.md
```diff
@@ -0,0 +1,280 @@
...
+
+At the block mode level, counter modes have a wearout limit of
+2<sup>48.5</sup> blocks, or ~5.6 PiB for 128-bit ciphers like AES, for an
+attacker advantage of 2⁻³². See Equation (7) of [ePrint 2024/051][]. A full-size
+message of 16 PiB provides the attacker with an advantage of 2⁻²⁹, which is
+still small and would require an impractical amount of resources. Applications
+may nonetheless choose to limit message size to e.g. 4 PiB by limiting the chunk
+counter to 2³⁸ instead of 2⁴⁰.
```
SalusaSecondus: This does bother me. I agree that the probabilities are still low but I really would not want to push us past the 2^48 block boundary. I also don't know the real-world resource requirements of this attack when optimized.

SalusaSecondus (commented):
Overall seems very reasonable. I dislike going past the (conservative) PRF-PRP distinguishing bound. I also don't like the short commitment value.

Other than that, this seems good. I'll keep thinking on it.

FiloSottile:
I find draft-irtf-cfrg-aead-limits basically unreadable, so I use it as a list of references to look at, which is how I arrived at the numbers in the Security Analysis. I think their bounds are so small because they target 2^-50 advantage, while the analysis in this document targets 2^-32. Any reason 2^-32 is "noticeable" and 2^-50 is not?

chunked-encryption.md
FiloSottile: Good catch, switched to 256 bits.

chunked-encryption.md
FiloSottile: In a theoretical paper, I'd agree that AE is more accurate, but developers are familiar with the concept of AEAD and the APIs are named AEAD and the IANA registry is for AEADs, so I think it's more helpful to use that name.

chunked-encryption.md
FiloSottile: Indeed, it's sort of acknowledged later

> The commitment provides a verification oracle for the input key and salt, but since their combined space (128 + 192 = 320 bits in the worst case) exceeds the derived key and base nonce space (128 + 96 - 40 = 184 bits in the worst case), it does not become the bottleneck for multi-target security.

I am 100% team "large enough is enough" and I don't care about 256 bit security levels. (That's not to say 256-bit values aren't useful to hit "large enough" collision or multi-target bounds.)

chunked-encryption.md
FiloSottile: Thanks for the pointer! ePrint 2018/993 models hidden nonces as part of the XN transform.

> An even more effective approach (at least with respect to preventing the above attack) are so called “unpredictable nonces”, and this is the approach taken by TLS 1.3 [20] and previously used within the SRTP protocol [1]. 

> The first one implements unpredictable nonces, as in TLS 1.3 and SRTP, and we refer to it as XN.

chunked-encryption.md
FiloSottile: I went back and forth a lot on this, and I tried to figure out the actual attack resource figures.

AFAIK, the best confidentiality attack (as opposed to distinguishing attack, which is not relevant to us because mostly we don't care about an attacker being able to tell something is encrypted with AES-CTR) is https://link.springer.com/chapter/10.1007/978-3-319-78375-8_24. I am not 100% sure of my reading of the paper, but I think the attack is essentially as unfeasible at 2^(n/2-14) blocks as at 2^(n/2-16).

Flipping this around, why is an attacker distinguishing advantage of 2⁻²⁹ unacceptable, but 2⁻³² is acceptable?

We could just say there's 2^38 max chunks instead of 2^40, it'd be fine, I don't think there's a lot of use cases that need more than 4 PiB messages but not 16 PiB. I just wish we had a principled reason for it.

FiloSottile: Alright, switched to a 2^38 / 4 PiB limit, those extra PiB are not worth anyone feeling uncomfortable with the security bounds.

FiloSottile:
https://github.com/FiloSottile/chunked is a Go implementation of this spec draft. It has pull-based (`io.Reader`) encryption, push-based (`io.Writer`) encryption, pull-based decryption, and random access (`io.ReaderAt`) decryption.

I am quite pleased with how simpler the implementation was compared to the age design, despite being stronger.

There is a large set of test vectors in `testdata/vectors.json`.

chunked-encryption.md
```diff
@@ -0,0 +1,288 @@
...
+If instantiated with an AEAD having a tag length of 16 bytes, this scheme has a
+fixed overhead of 56 bytes and a marginal overhead of approximately 0.1%.
+
+## Specification
+
+At a high level, for each *message* a single-use random *key*, *base nonce*, and
+*commitment* are generated with HKDF-Expand from an *input key*, *salt*, and
+*context*. The message is split into 16 KiB chunks, with the last chunk being
+always shorter. Each chunk is then encrypted with an AEAD, the *key*, and the
+*base nonce* XOR’d with a chunk counter.
```
intricate, line 27: So is it the case that this scheme wouldn't support "segment rewrites" (which I believe [FLOE](https://github.com/C2SP/C2SP/blob/01194db11950a0ae2b2b2401afa0a4063a822458/FLOE.md) technically supports due to its usage of random nonces per segment)? 

At least, when using an AEAD that isn't nonce-misuse resistant?

If I'm misunderstanding, please forgive my ignorance :grin:

FiloSottile: Correct, this is sequential encryption, random-access read.

alex:
Idle thought: given the strong recommendation for "use AES-128-GCM w/ SHA256" it might be nice to give that instantiation of the format a name, and then libraries can expose that, rather than exposing the parametrized version.

FiloSottile:
Agreed, want to do the hard part for me and suggest a name? :)

alex:
How about cobblestone? You can make a long path out of many individual cobbles.

chunked-encryption.md
```diff
@@ -0,0 +1,288 @@
...
+implemented internally by higher-level libraries as a component of the
+higher-level protocol, or exposed as a “hazmat” primitive meant for implementing
+protocols.
+
+The ciphertext offset and validity formulas in the implementation notes must be
+adjusted to remove the 56-byte prefix when implementing raw mode.
+
+[STREAM]: https://eprint.iacr.org/2015/189
+[FLOE]: https://c2sp.org/FLOE
+[TLS 1.3]: https://datatracker.ietf.org/doc/html/rfc8446#section-5.2
```
str4d: This reference is only used to point to "the TLS 1.3 record layer design", but the record layer is specified in section 5.1. Section 5.2 specifies how a single fragment is encrypted, and the nonce XORing from section 5.3 is also adopted here. I think it makes more sense to reference section 5 as a whole:
```suggestion
[TLS 1.3]: https://datatracker.ietf.org/doc/html/rfc8446#section-5
```

chunked-encryption.md
```diff
@@ -0,0 +1,288 @@
...
+    info = "c2sp.org/chunked-encryption@v1+" || aead || 0x00 || salt || context)
+```
+
+The *input key* MUST be uniformly random, e.g. generated by a CSPRNG or output
+by a KDF.
+
+*aead* is the name of the underlying AEAD, from the [AEAD Algorithms IANA
+registry][], like `AEAD_AES_128_GCM`.
+
+*salt* is a sequence of 24 bytes that MUST NOT repeat for a certain *input key*,
```
str4d, line 56: Is `salt` the right term to use here, or will it be potentially misleading when it eventually appears as a parameter name in a library API? Salts are normally reusable values, whereas this is functionally a nonce.

str4d (commented):
Generally looks great to me!

chunked-encryption.md
FiloSottile: Yeah I don't love this but nonce is already used for the AEAD, and I like slightly incorrect better than overloading.

PR #233, opened 2026-04-01
Add tlog-policy spec

niels-moller:


tlog-policy.md
```diff
@@ -0,0 +1,241 @@
...
+
+A log is defined by a line
+```
+log <vkey> [<url>]
+```
+When the policy is used for verifying a proof, all of the
+listed logs are accepted. 
+
+The log's vkey includes a "key name". This name MUST correspond to the
+log's origin line, representing the identity of the log.
```
niels-moller, line 128: I think this is MUST-important, since there's no other way to specify the origin line.

tlog-policy.md
```diff
@@ -0,0 +1,241 @@
...
+group <name> <k> <name>...
+```
+
+All these defines a group, where the group is considered to witness a
+checkpoint if and only if at least k of its n members have witnessed
+that checkpoint, each group member being either a witness or another
+group. In this terminology, for a single witness, "witnessing" is the
+same as cosigning.
+
+At least one member must be listed (i.e., n >= 1). The `any` variant
```
niels-moller, line 197: Maybe require n >= 2? A single-member group is essentially a rename, and not terribly useful.

tlog-policy.md
```diff
@@ -0,0 +1,241 @@
...
+  consider the checkpoint to be valid.
+
+Only the quorum is mandatory. An application where the appropriate
+log(s) are known from other context may accept a policy with an empty
+set of logs. A policy without witness cosigning can be defined using a
+`quorum none` line, and in this case, the set of witnesses may be
+empty.
+
+The policy says that a checkpoint is considered valid if it is signed
+by any one of the listed logs, and it is cosigned according to the
```
BenBirt: nit: maybe "cosigned by witnesses" for clarity? (very minor, likely doesn't matter)

tlog-policy.md
```diff
@@ -0,0 +1,241 @@
...
+defined quorum rule.
+
+Both logs and witnesses are identified primarily by their public keys.
+Each log or witness may also have an associated URL; this is required
+for operations interacting with the log or witness, but no URLs are
+needed if the policy is used only for offline verification.
```
BenBirt: nit: s/needed/required/

tlog-policy.md
```diff
@@ -0,0 +1,241 @@
...
+the policy file in detail.
+
+## Example policy
+
+This is an example of a policy with a quorum defined using two levels of
+groups. Actual keys are elided, for brevity, and the optional URLs are
+omitted:
+```
+log <vkey>
+witness X1 <vkey>
```
BenBirt, line 60: supernit: maybe extra empty line after the `log` line, to indicate it is not part of - or anything related to - the `X-witnesses` group?

tlog-policy.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Public key representation
+
+Public keys are written in [vkey][] format. (The `sigsum-key` command
+can be used to convert public keys between raw hex, OpenSSH, and
+[vkey][] formats). Log keys must use signature type `0x01` (Ed25519),
+and witness keys must use signature type `0x04` (Timestamped Ed25519
+witness cosignatures). Two vkeys are considered *duplicate* if they
+wrap the same underlying Ed25519 public key, even if they differ by
+key name and key id.
```
BenBirt: what should happen if the log vkey underlying Ed25519 key matches a witness' vkey underlying Ed25519 key?

(presumably this should be rejected, but maybe worth spelling out?)

tlog-policy.md
BenBirt: +1 I came to the same conclusion in https://github.com/project-oak/oak/blob/main/tr/c2sp/policy.rs#L148-L152 (also see https://github.com/project-oak/oak/blob/main/tr/c2sp/lib.rs#L244-L253)

tlog-policy.md
```diff
@@ -0,0 +1,241 @@
...
+```
+witness <name> <vkey> [<url>]
+```
+The name is used to refer to this witness when defining the quorum, or
+when defining witness groups.
+
+Since only logs and possibly monitors interact directly with
+witnesses, most policy files will not need any witness URLs. The
+meaning of the URL is application specific, but if a URL is provided,
+it is recommended that it is the witness' "submission prefix" URL, as
```
BenBirt, line 152: do you see any reason _not_ to enforce this recommendation?

i don't feel particularly strongly, but interested if you have thoughts.

tlog-policy.md
```diff
@@ -0,0 +1,241 @@
...
+
+Defining a witness group is required for defining a quorum that is not
+a single witness. A group is defined by a line of one of these forms:
+```
+group <name> all <name>...
+group <name> any <name>...
+group <name> <k> <name>...
+```
+
+All these defines a group, where the group is considered to witness a
```
BenBirt: nit: s/All of these defines/Each of these defines/

tlog-policy.md
```diff
@@ -0,0 +1,241 @@
...
+and 0x80-0xFF.
+
+Non-ASCII octets may appear in names, URLs and comments. Using UTF-8
+for any non-ASCII text is strongly recommended, but with respect to
+the policy file syntax and semantics, all octets with the high bit set
+are opaque. In particular, names must be handled as opaque octet
+strings, e.g., the name 0x4B (ASCII "K") is different from the name
+0xE2 0x84 0xAA (a Kelvin sign encoded using UTF-8).
+
+The motivation for this way of allowing non-ASCII, but handle it as
```
BenBirt: nit: "The motivations for allowing non-ASCII, but handling it as opaque data, are to:"

BenBirt (commented):
Looks great!

tlog-policy.md
niels-moller: Done.

tlog-policy.md
niels-moller: Done.

tlog-policy.md
niels-moller: Done.

tlog-policy.md
niels-moller: My thinking is that it should be allowed. In https://git.glasklar.is/sigsum/project/documentation/-/blob/main/proposals/2025-10-stricter-policy-format.md, it was motivated like this "Note that it is still allowed to list a log and a witness that use the same public key. Motivation: When an organization operates both a log and a witness, there's sufficient domain separation between log and witness signatures to make it a technically valid option to use the same signing key. Furthermore, if a policy file includes a log and a witness with the same key, by mistake, that seems likely to make things break in obvious rather than subtle ways."

tlog-policy.md
niels-moller: For a bit of background, our log server (https://git.glasklar.is/sigsum/core/log-go) uses a policy file to configure the key and url for witnesses the log should talk to, and kind-of recently switch from a more Sigsum-specific protocol to the tlog-witness protocol. Using a policy file for this purpose is perhaps debatable, but so far that's the only use case I'm aware of for witness urls in the policy file. 

I'm thinking that tlog-witness is one way of collecting cosignatures, which helps interop between logs and witnesses. But it doesn't have to be the only way, and furthermore, that protocol is not very relevant for applications that use tlog-policy but aren't logs. So I'd tend to leave open for whatever applications find useful.

BenBirt (approved):

tlog-policy.md
```diff
@@ -0,0 +1,242 @@
...
+  consider the checkpoint to be valid.
+
+Only the quorum is mandatory. An application where the appropriate
+log(s) are known from other context may accept a policy with an empty
+set of logs. A policy without witness cosigning can be defined using a
+`quorum none` line, and in this case, the set of witnesses may be
+empty.
+
+The policy says that a checkpoint is considered valid if it is signed
+by any one of the listed logs, and it is cosigned by witnesses
```
mhutchinson, line 35: Did you consider options for N-of-M with the logs? Or something even richer, like the quorum rules for the witnesses? This "any single log" constraint means that this policy couldn't work for CT.

tlog-policy.md
```diff
@@ -0,0 +1,242 @@
...
+n. If k is given numerically, it must be written in decimal, and it is
+required that 1 <= k <= n.
+
+Like for the quorum definition, a group definition can only refer to
+names defined on preceding lines. (This also rules out circular group
+definitions).
+
+Each defined name can be listed as a group member at most once. This
+ensures that each witness can contribute to a group, or to the quorum
+in particular, in only one way.
```
mhutchinson, line 208: Does this rule apply to witnesses and groups? It's placed under the `Defining a witness group` heading so I assumed the latter, but as written, it sounds like it applies to anything with a name?

Why does it exist - is it load bearing? It seems to be enforcing a weak variant of boolean factoring. e.g.

```
(A AND B) OR (A AND C) // illegal, A is mentioned twice
A AND (B OR C)  // legal
```

Am I understanding this right?

tlog-policy.md
rgdd: Yes, but I don't think the m-of-n logs pattern is the one we want to encourage long-term.

If we want to make split views harder, it's just regular witness quorum using the tlog-cosignature base semantics. If we want to make it less likely that the logged data is lost and nowhere to be found, then we can additionally express and require a tlog-mirror cosigigning group.

So I'd say the pitch for not supporting m-of-n logs is 'more clear separation of concerns', and also 'less encouragement for duplicate entries that monitors have to download'.

WDYT, makes sense or are you still missing the ability to express more complicated log policies than 'any'? My understanding is that this fits well already for MTC, but you're right that it doesn't fit today's CT policies.

mhutchinson: Yeah, I'm happy with this as an outcome. I think the spec would benefit from a sentence or two to discuss this design decision.

tlog-policy.md
niels-moller: It is intended to apply to all names, i.e., both witnesses and groups. The way I think of it, it forces the graph of relations between groups and witnesses to be a proper tree.

tlog-policy.md
niels-moller: We can get into trouble if logs go offline or suffer data loss; witnesses still protect against split views, but if monitors can't get the data underlying the single view under those cosignatures, that's no so reassuring.

But I think that is somewhat out of scope for this spec; the point of the log lines here is merely to *identify* the logs that ought to be monitored in order to discover all items that are accepted under the policy.

Not sure how to express that in a concise way for this context, any suggestions?

rgdd: Without giving a text suggestion, I'd frame it based on my comment earlier and point towards tlog-mirror.

BenBirt:
Hey folks, what's the next step to getting this spec published?

tlog-policy.md
```diff
@@ -0,0 +1,242 @@
...
+only reference names of groups and witnesses defined on preceding
+lines. Similarly, the quorum line must specify a witness or group
+defined earlier.
+
+### Public key representation
+
+Public keys are written in [vkey][] format. (The `sigsum-key` command
+can be used to convert public keys between raw hex, OpenSSH, and
+[vkey][] formats). Log keys must use signature type `0x01` (Ed25519),
+and witness keys must use signature type `0x04` (Timestamped Ed25519
```
florolf: signed-note now also specifies timestamped ML-DSA cosignatures (which are now also supposed to be used by logs as per https://github.com/C2SP/C2SP/commit/fcbd7d398d44a9331933d87b31155362f25aae79), so maybe we should just defer to the other specs instead of tying this down explicitly here?

tlog-policy.md
```diff
@@ -0,0 +1,242 @@
...
+The name is used to refer to this witness when defining the quorum, or
+when defining witness groups.
+
+Since only logs and possibly monitors interact directly with
+witnesses, most policy files will not need any witness URLs. The
+meaning of the URL is application specific, but if a URL is provided,
+it is recommended that it is the witness' "submission prefix" URL, as
+defined by the [witness protocol][].
+
+Duplicate witnesses, i.e., multiple witness lines with the same public
```
florolf, line 155: As discussed offline last week, both @niels-moller and my gut feeling is that it is useful to allow unreferenced witnesses and groups (though having separate linting that warns about this would be useful too). Not sure if this is worth spelling out explicitly in the spec.

tlog-policy.md
florolf: I would be slightly in favor of allowing it, since it allows for adding/removing witnesses without changing the (upstream) structure of the policy, e.g. going from

```
group glasklar-test-witnesses all poc.sigsum.org/nisse rgdd.se/poc-witness
```

to

```
group glasklar-test-witnesses all poc.sigsum.org/nisse
```

or the other way around if a new witness gets set up.

tlog-policy.md
niels-moller: Maybe something like "signature type appropriate for log signing, e.g., 0x01 (Ed25519)", and similarly for witness keys?

tlog-policy.md
niels-moller: You still can't drop the last member of a group, so question is, should it be valid to drop the second to last one? One minor reason this is a bit of an itch to me is that I'd like to get the same byte code for
```
group all othername name
quorum othername
```
and just
```
quorum name
```
(Since I don't want naming to matter). And then group size 1 becomes a special case, implemented here: https://git.glasklar.is/sigsum/core/sigsum-c/-/blob/3e17924ae4f8ca291417a529a7c5c2e374e1d8f9/lib/policy-compile.c#L165 and here: https://git.glasklar.is/sigsum/apps/sign-if-logged/-/blob/2879aae3d75eb4db7a6c4c01d865ebe933c7abec/internal/compiler/policy.go#L113.

florolf: I have similar special case in my implementation and I agree that it's a bit of a wart. Still, I think it's a good thing to have ergonomics-wise. My takeaway from your argument would be that we maybe even should allow empty groups (though you can then shoot yourself in the foot when the upstream group uses `all`), but I don't want to open that can of worms for now.

Another argument would be that if we forbid single-member groups in the spec, the expectation would be that parsers must refuse such policies, so you wouldn't be able to get rid of the special case in the code either, it would just trigger different behavior (an error, rather than inlining).

tlog-policy.md
florolf: I think it makes sense to refer to the specs explicitly. E.g.

> Logs and witnesses must use signature types appropriate for their specific function, see tlog-checkpoint and tlog-witness, respectively. Supported key types depend on the specific application.

florolf:
Another thought: Other C2SP specs use RFC 2119 conventions (SHOULD, MUST, etc.), we currently don't. Should we?

I'm not familiar with how strong the preference is to keep (especially) related specs like the whole tlog-* world aesthetically aligned and if this is maybe a blocker?

Either way, personally I'm happy to just get this of this out so we have *something* to coordinate on rather than adding more bikeshedding on top myself.

niels-moller:
> Another thought: Other C2SP specs use RFC 2119 conventions (SHOULD, MUST, etc.), we currently don't. Should we?

I think it makes sense to use that terminology, for consistency with both c2sp and ietf specs.

rgdd:
> Another thought: Other C2SP specs use RFC 2119 conventions (SHOULD, MUST, etc.), we currently don't. Should we?

Makes sense to me as well. If not here, in a separate follow-up PR.

BenBirt:
Hey folks, where are we on this PR? Does anybody need to be poked for a review/LGTM?

florolf:
Not sure what the exact flow is here. Filippo has acked the spec here https://github.com/C2SP/C2SP/issues/226#issuecomment-4518361782.

Looking at the failing CI jobs I think this PR needs to add the spec to `.github/README.md` and `.github/policy.yml`, see e.g. #232. I think @niels-moller needs to do that as the original author of the PR.

niels-moller:
> Not sure what the exact flow is here. Filippo has acked the spec here [#226 (comment)](https://github.com/C2SP/C2SP/issues/226#issuecomment-4518361782).
> 
> Looking at the failing CI jobs I think this PR needs to add the spec to `.github/README.md` and `.github/policy.yml`, see e.g. #232. I think @niels-moller needs to do that as the original author of the PR.

I'm also not sure what next step is and I'm a bit lost with the various github actions involved. I'd expect that someone has to update the ultimate source of truth for MAINTAINERS.md and policy.yml, but not sure what that is, and if update can go in this PR. Please let me now what's needed on my part.

tlog-policy.md
niels-moller: I've pushed a change to drop the Ed25519 requirements.

niels-moller:
I think all procedural hoops are done, everything looks green and it looks like I have permission to press the merge button. I've marked a couple of threads as resolved. Open topics:

1. Do we need clarifying/motivating text on the policy semantics of the log lines?
2. Allow or disallow groups with a single member?
3. Reword to follow RFC conventions for SHOULD/MUST, etc?

I think I'd be happy to merge as is, and do incremental improvements as needed.

BenBirt:
As far as I'm concerned - I'm happy either way (perhaps with a slight preference for sorting out the MUST/SHOULD before merging - but only slight). LGTM!

florolf:
Likewise. I was going to agree with Ben on the RFC 2119 terminology but then I looked into rewording the document and it needs quite a lot of structural changes to make sense without adding much clarity so I'd now lean into the direction to merging it as is for the time being.

One minor thing I noticed though that I had missed before is that the "Structure" section mentions leading and trailing whitespace and comment lines, but this doesn't imply empty lines being allowed (which they in fact are). Maybe this is just common sense though.

Anyway, feel free to use your own judgement on whether to add this @niels-moller. Looks good to merge to me.

niels-moller:
> One minor thing I noticed though that I had missed before is that the "Structure" section mentions leading and trailing whitespace and comment lines, but this doesn't imply empty lines being allowed (which they in fact are).

I've reworded to make explicit that empty lines are allowed.

> Anyway, feel free to use your own judgement on whether to add this @niels-moller. Looks good to merge to me.

I'll merge this soon. One question as a github/c2sp newbie: I think I should press the "squash and merge" button. But what commit message will I get with that, and will github let me edit it after pressing the button? Looking at C2SP history, it seems I should use a basic commit message like "tlog-policy: new spec".

FiloSottile:
> One question as a github/c2sp newbie: I think I should press the "squash and merge" button. But what commit message will I get with that, and will github let me edit it after pressing the button? Looking at C2SP history, it seems I should use a basic commit message like "tlog-policy: new spec".

Yes the button will open a edit interface.

image

Up to you on the commit body, for the summary line please start with `tlog-policy: ` and include the final ` (#233)` so it's easy to go back to the discussion.

florolf (approved):

PR #236, opened 2026-04-08
.github: Set up C2SP OID registry with first assignment

str4d:
I've suggested a rough tree structure for OIDs we assign, that gives us flexibility to change our minds later.

/cc @C2SP/stewards

### DIFF
--- a/.github/OIDs.md
+++ b/.github/OIDs.md
@@ -1,0 +1,32 @@
+# OIDs
+
+C2SP has been assigned a Private Enterprise Number ([RFC 9371]). This PEN is the
+root of OIDs used as identifiers in C2SP specifications.
+
+C2SP OIDs are assigned by the [C2SP stewards].
+
+[RFC 9371]: https://www.rfc-editor.org/rfc/rfc9371.html
+[C2SP stewards]: MAINTAINERS.md#stewards
+
+## Requesting a C2SP OID assignment
+
+To request an OID for an entity, [create an issue] giving details about the
+nature or purpose of the entity, and referencing the specification that the
+entity is (or will be) defined in or related to.
+
+The stewards decide whether to grant the assignment, and what structure the OID
+will have.
+
+[create an issue]: https://github.com/C2SP/C2SP/issues/new?template=oid.md
+
+## Registry of assigned C2SP OIDs
+
+| OID | Entity |
+| --- | --- |
+| `1.3.6.1.4.1.64829` | [C2SP] |
+| `1.3.6.1.4.1.64829.1` | [age] |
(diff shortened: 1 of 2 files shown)


.github/OIDs.md
```diff
@@ -0,0 +1,33 @@
...
+The stewards decide whether to grant the assignment, and what structure the OID
+will have.
+
+[create an issue]: https://github.com/C2SP/C2SP/issues/new?template=oid.md
+
+## Registry of assigned C2SP OIDs
+
+| OID | Entity |
+| --- | --- |
+| `1.3.6.1.4.1.64829` | [C2SP] |
```
str4d, line 26: https://www.iana.org/assignments/enterprise-numbers/?q=C2SP

.github/OIDs.md
```diff
@@ -0,0 +1,33 @@
...
+| `1.3.6.1.4.1.64829.1` | parent specification |
```
FiloSottile: Do we need this intermediate arc? I am sure we'll never have so many specs we overflow a int32 (which is where there starts being compatibility issues), and this is adding one byte to all our OIDs. Also, 1.3.6.1.4.1.64829.1 would not be a meaningful standalone entity.

(I was tempted to argue not to have even the age arc and go straight to necessary codepoints, but I feel that's a weaker argument and I'm ambivalent about it.)

str4d: The only purpose this arc serves is letting us change our mind on categorisation later while retaining consistent format. If we don't care about that, then we can indeed remove this arc and then just add an equivalent one later if we determine we do need a different sub-category. (We could even do so and still retain consistency if we wrote a C2SP spec for that category.)

Re: the age arc, I feel it's beneficial to have C2SP OIDs be connected to C2SP specifications. It also means that one spec's potential assignment of thousands of OIDs doesn't materially impact the encoded length of other specs' OIDs, since I agree that the number of specs wanting OIDs will grow significantly more slowly.

FiloSottile: Yeah the arguments for per-spec OIDs are good enough to justify the extra byte/arc.

I agree we can change our mind later on the category arcs, and also there is probably not going to be ever anything that _requires_ all our top-level assignments to be specs, so we can always mix.

Let's go with 1.3.6.1.4.1.64829.1 for age and 1.3.6.1.4.1.64829.1.1 for the extension?

str4d: Done.

FiloSottile (approved):

FiloSottile:
Minor nit: I've been using `[prefix]: [lowercase]` for commit messages. (I mostly inherited that from the Go repo.)

str4d:
Updated the commit message, just for you. (Though I note my previous commit message did still match others that have been merged, and we aren't going to get consistency across all spec maintainers.)

FiloSottile (approved):

dconnolly (approved):

dconnolly:
lgtm !

PR #237, opened 2026-04-09
tlog-cosignature: add ML-DSA-44 subtree cosignatures

FiloSottile:
The quantum computer timelines are [getting more worrying](https://words.filippo.io/crqc-timeline/), so we need a post-quantum signature algorithm for the tlog ecosystem. ML-DSA seems to be the algorithm all the implementation, analysis, and deployment energy is concentrating on. ML-DSA-44 is a Category 2 parameter set, which provides some margin over the 128-bit security level.

This gives us an opportunity to apply a few lessons learned and extend compatibility:

1. scope this signature algorithm for both log and witness signatures, since there is no reason for them to use different ones
2. include the cosigner name in the signature, to make it safe to reuse the same key across two different cosigners
3. use a binary encoding for the signed message, to ease compatibility with the Merkle Tree Certificate ecosystem

### DIFF
--- a/signed-note.md
+++ b/signed-note.md
@@ -121,10 +121,13 @@ runtime if possible.
 
 * `0x03` — Reserved.
 
-* `0x04` — Timestamped Ed25519 witness cosignatures, as specified by
+* `0x04` — Timestamped Ed25519 checkpoint cosignatures, as specified by
   c2sp.org/tlog-cosignature.
 
 * `0x05` — RFC 6962 `TreeHeadSignature`s, as specified by c2sp.org/static-ct-api.
+
+* `0x06` — Timestamped ML-DSA-44 (sub)tree cosignatures, as specified by
+  c2sp.org/tlog-cosignature.
 
 * `0xfa`–`0xfe` — Reserved for future use.
 
--- a/tlog-cosignature.md
+++ b/tlog-cosignature.md
@@ -1,7 +1,7 @@
 # Transparency Log Cosignatures
 
 A cosignature is a statement by a *cosigner* that it verified
-the consistency of a [checkpoint][]. Log clients can verify a quorum of
+the consistency of a [checkpoint][] or [subtree][]. Log clients can verify a quorum of
 cosignatures to prevent split-view attacks before trusting an inclusion proof.
 A cosigner may make additional statements relating to a checkpoint.  Log clients
 that know about this can then be assured of additional cosigning properties.
--- a/tlog-cosignature.md
+++ b/tlog-cosignature.md
@@ -17,6 +17,12 @@ CsUYapGGPo4dkMgIAUqom/Xajj7h2fB2MPA3j2jxq2I=
 — witness.example.com/w1 jWbPPwAAAABkGFDLEZMHwSRaJNiIDoe9DYn/zXcrtPHeolMI5OWXEhZCB9dlrDJsX3b2oyin1nPZqhf5nNo0xUe+mbIUBkBIfZ+qnA==
 ```
 
+This document specifies two cosignature types: one based on Ed25519, and one
+based on ML-DSA-44. The ML-DSA version SHOULD be used for new deployments, and
+it's extended to also commit to the cosigner's name, and to allow signing
+subtrees that don't start at zero. (Those subtrees don't currently have a
+[checkpoint][] representation.)
+
 ## Conventions used in this document
 
 Data structures are defined according to the conventions laid out in Section 3
@@ -52,40 +58,51 @@ Per the signed note format, a note signature line is
 
 The key name SHOULD be a schema-less URL that identifies the cosigner. Like the
 checkpoint origin line, this is for disambiguation, and MAY match a publicly
-reachable endpoint or not.
+reachable endpoint or not. For ecosystems that use OIDs for identification, the
+key name MAY be the string `oid/` followed by an OID in dotted decimal form.
 
-The key ID MUST be
+The key ID for Ed25519 cosignatures MUST be computed as
 
     SHA-256(<name> || "\n" || 0x04 || 32-byte Ed25519 cosigner public key)[:4]
+
+The key ID for ML-DSA-44 cosignatures MUST be computed as
+
+    SHA-256(<name> || "\n" || 0x06 || 1312-byte ML-DSA-44 cosigner public key)[:4]
 
 Clients are configured with tuples of (cosigner name, public key, supported
 cosignature version) and based on that they can compute the expected name and
 key ID, and ignore any signature lines that don't match the name and key ID.
 
-Public keys MAY be encoded as [vkeys][] with signature type 0x04 and the 32-byte
-Ed25519 cosigner public key as the public key material.
+Ed25519 public keys MAY be encoded as [vkeys][] with signature type 0x04 and the
+32-byte Ed25519 cosigner public key as the public key material.
+
+ML-DSA-44 public keys MAY be encoded as [vkeys][] with signature type 0x06 and the
+1312-byte ML-DSA-44 cosigner public key as the public key material.
 
 Future cosignature formats MAY reuse the same cosigner public key with a
-different key ID algorithm byte (and a different signed message header line).
+different key ID algorithm byte (and a different newline-terminated prefix).
 
-The signature MUST be a 72-byte `timestamped_signature` structure.
+The signature MUST be a `timestamped_signature` structure.
 
     struct timestamped_signature {
         u64 timestamp;
-        u8 signature[64];
+        select (signature_algorithm) {
+            case ed25519: opaque ed25519_signature[64];
+            case ml-dsa-44: opaque ml_dsa_44_signature[2420];
+        } signature;
     }
 
 "timestamp" is the time at which the cosignature was generated, as a POSIX
 timestamp.  It MUST NOT exceed 2^63 - 1, and verifiers MAY reject cosignatures
 with timestamps in the future.
 
-"signature" is an Ed25519 ([RFC 8032][]) signature from the cosigner public key
-over the message defined in the next section.
+"signature" is an Ed25519 ([RFC 8032][]) or ML-DSA-44 ([FIPS 204][]) signature
+from the cosigner public key over the message defined below.
 
 Per [RFC 8446][], Section 3.3, these are serialized in sequence, with the
 timestamp encoded in big-endian order.
 
-## Signed message
+## Ed25519 signed message
 
 The signed message MUST be two newline (U+000A) terminated lines (one header
 line and one timestamp line) followed by the whole note body of the cosigned
@@ -115,20 +132,87 @@ the log based on the first three lines of the checkpoint; consensus between
 cosigners on the extension lines SHALL NOT be assumed, and no semantic statement
 is made about any extension lines unless the cosigner's operator says otherwise.
 
+A cosigner operator that operates multiple Ed25519 cosigners (e.g. with distinct
+additional statements, see below) MUST use distinct public keys for each
+cosigner. The Ed25519 signed message format doesn't commit to the cosigner name,
+so the same public key can't be used across multiple cosigners.
+
+## ML-DSA-44 signed message
+
+The signed message MUST be a `cosigned_message` structure.
+
+    struct cosigned_message {
+        uint8 label[12] = "subtree/v1\n\0";
+        opaque name<1..2^8-1>;
+        uint64 timestamp;
+        opaque origin<1..2^8-1>;
+        uint64 start;
+        uint64 end;
+        uint8 root_hash[32];
+    }
+
+`name` is the cosigner name. If the name starts with `oid/`, it MUST be encoded
+as specified below in OID Encoding.
+
+`timestamp` is `timestamped_signature.timestamp`. These two values MAY be zero
+if the cosigner doesn't make any statement as to the tree being the largest
+observed at time of signing.
+
+`origin` is the log's origin, as represented in a checkpoint's origin line
+without the final newline. If the origin starts with `oid/`, it MUST be encoded
+as specified below in OID Encoding.
+
+`start` is the index of the first leaf included in the [subtree][] being signed.
+If signing a [checkpoint][], it MUST be zero. If `start` is not zero,
+`timestamp` MUST be zero.
+
+`end` is the index of the last leaf included in the [subtree][] being signed,
+plus one. If signing a [checkpoint][], it is the size of the tree.
+
+`root_hash` is the root hash of the subtree being signed.
+
+Semantically, a v1 subtree cosignature is a statement that the subtree with the
+specified root hash is consistent with all other historical views observed by
+the cosigner of the log identified by the origin line. If the timestamp is not
+zero, it is also a statement that, as of the specified time, this is the
+consistent tree head with the largest size the cosigner has observed for the log.
+
+Note that checkpoint extension lines are not included in the signed message for
+ML-DSA-44 cosignatures, and no statement is made about them. Subtrees with start
+different from zero currently don't have a checkpoint representation.
+
+### OID encoding
+
+If the cosigner name or log origin line starts with `oid/`, the rest of the
+string MUST be an OID in dotted decimal form, and it MUST be replaced with its
+DER encoding without tag and length, preserving the `oid/` prefix.
+
+For example, the name `oid/2.999.42.1337` would be encoded as
+
+    6f 69 64 2f 88 37 2A 8A 39
+
+This encoding is meant to disambiguate OID-based names from domain-based ones,
+while letting verifiers in OID-based ecosystems avoid carrying a decimal OID
+encoder. It is not meant to be represented on the wire: the ASCII name SHOULD be
+used wherever both types may occur, and an appropriate OID encoding (such as DER
+or even a RELATIVE OID encoding) can be used in contexts where only OID-based
+names are expected.
+
+## Additional statements
+
 A cosigner MAY make additional statements about a checkpoint.  These
 additional statements need to be communicated out of band to those defining
 trust policies based on tuples of (public key, supported cosignature version).
 A given tuple MUST imply a single set of statements.  These statements MUST
-include the cosignature/v1 semantics, and MAY include other statements that
+include the base cosignature semantics, and MAY include other statements that
 are non-conflicting.  Examples of non-conflicting statements include "I also
-mirrored the log up until the checkpoint size" and "I also reproduced the sparse
-Merkle tree root in the extension lines".  See [tlog-mirror][] for an example.
-
-A cosigner operator that operates multiple cosigners with distinct additional
-statements MUST use distinct public keys for each cosigner. In cosignature/v1,
-it is not sufficient to use a distinct cosigner name.
+mirrored the log up until the checkpoint size", "I also reproduced the sparse
+Merkle tree root in the extension lines", and "I certify the SAN ←→ public key
+associations in the log leaves".  See [tlog-mirror][] for an example.
 
 [note signature]: https://c2sp.org/signed-note@v1.0.0
 [vkeys]: https://c2sp.org/signed-note@v1.0.0#verifier-keys
 [checkpoint]: https://c2sp.org/tlog-checkpoint@v1.0.0
 [tlog-mirror]: https://c2sp.org/tlog-mirror
+[FIPS 204]: https://csrc.nist.gov/pubs/fips/204/final
+[subtree]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-02.html#name-subtrees
--- a/tlog-cosignature.md
+++ b/tlog-cosignature.md
@@ -18,10 +18,13 @@ CsUYapGGPo4dkMgIAUqom/Xajj7h2fB2MPA3j2jxq2I=
 ```
 
 This document specifies two cosignature types: one based on Ed25519, and one
-based on ML-DSA-44. The ML-DSA version SHOULD be used for new deployments, and
-it's extended to also commit to the cosigner's name, and to allow signing
-subtrees that don't start at zero. (Those subtrees don't currently have a
-[checkpoint][] representation.)
+based on ML-DSA-44. The ML-DSA version SHOULD be used for new deployments.
+
+Unlike the Ed25519 type, the ML-DSA-44 type is secure against quantum computers.
+Moreover, it commits to the cosigner's name, and supports signing [subtrees][]
+in addition to [checkpoints][checkpoint]. The ML-DSA-44 parameter set was
+selected because at NIST Level 2 it provides some margin beyond the 128-bit
+security level.
 
 ## Conventions used in this document
 
@@ -84,19 +87,19 @@ different key ID algorithm byte (and a different newline-terminated prefix).
 
 The signature MUST be a `timestamped_signature` structure.
 
-    struct timestamped_signature {
+    struct {
         u64 timestamp;
         select (signature_algorithm) {
             case ed25519: opaque ed25519_signature[64];
             case ml-dsa-44: opaque ml_dsa_44_signature[2420];
         } signature;
-    }
+    } timestamped_signature;
 
-"timestamp" is the time at which the cosignature was generated, as a POSIX
+`timestamp` is the time at which the cosignature was generated, as a POSIX
 timestamp.  It MUST NOT exceed 2^63 - 1, and verifiers MAY reject cosignatures
 with timestamps in the future.
 
-"signature" is an Ed25519 ([RFC 8032][]) or ML-DSA-44 ([FIPS 204][]) signature
+`signature` is an Ed25519 ([RFC 8032][]) or ML-DSA-44 ([FIPS 204][]) signature
 from the cosigner public key over the message defined below.
 
 Per [RFC 8446][], Section 3.3, these are serialized in sequence, with the
@@ -141,45 +144,46 @@ so the same public key can't be used across multiple cosigners.
 
 The signed message MUST be a `cosigned_message` structure.
 
-    struct cosigned_message {
+    struct {
         uint8 label[12] = "subtree/v1\n\0";
-        opaque name<1..2^8-1>;
+        opaque cosigner_name<1..2^8-1>;
         uint64 timestamp;
-        opaque origin<1..2^8-1>;
+        opaque log_origin<1..2^8-1>;
         uint64 start;
         uint64 end;
-        uint8 root_hash[32];
-    }
+        uint8 hash[32];
+    } cosigned_message;
 
-`name` is the cosigner name. If the name starts with `oid/`, it MUST be encoded
-as specified below in OID Encoding.
+`cosigner_name` is the cosigner name. If the name starts with `oid/`, it MUST be
+encoded as specified below in [OID Encoding](#oid-encoding).
 
 `timestamp` is `timestamped_signature.timestamp`. These two values MAY be zero
 if the cosigner doesn't make any statement as to the tree being the largest
-observed at time of signing.
+observed at time of signing. If `start` is not zero, these values MUST be zero.
 
-`origin` is the log's origin, as represented in a checkpoint's origin line
+`log_origin` is the log's origin, as represented in a checkpoint's origin line
 without the final newline. If the origin starts with `oid/`, it MUST be encoded
-as specified below in OID Encoding.
+as specified below in [OID Encoding](#oid-encoding).
 
 `start` is the index of the first leaf included in the [subtree][] being signed.
 If signing a [checkpoint][], it MUST be zero. If `start` is not zero,
 `timestamp` MUST be zero.
 
-`end` is the index of the last leaf included in the [subtree][] being signed,
-plus one. If signing a [checkpoint][], it is the size of the tree.
+`end` is the exclusive upper bound of the indexes of the leaves in the
+[subtree][] being signed: the index of the last included leaf plus one. If
+signing a [checkpoint][], it is the size of the tree.
 
-`root_hash` is the root hash of the subtree being signed.
+`hash` is the root hash of the subtree being signed.
 
 Semantically, a v1 subtree cosignature is a statement that the subtree with the
 specified root hash is consistent with all other historical views observed by
 the cosigner of the log identified by the origin line. If the timestamp is not
-zero, it is also a statement that, as of the specified time, this is the
-consistent tree head with the largest size the cosigner has observed for the log.
+zero, it is also a statement that, as of the specified time, this is the largest
+consistent tree the cosigner has observed for the log.
 
 Note that checkpoint extension lines are not included in the signed message for
-ML-DSA-44 cosignatures, and no statement is made about them. Subtrees with start
-different from zero currently don't have a checkpoint representation.
+ML-DSA-44 cosignatures, and no statement is made about them. Subtrees with
+non-zero start values currently don't have a checkpoint representation.
 
 ### OID encoding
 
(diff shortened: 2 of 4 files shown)


tlog-cosignature.md
```diff
@@ -17,6 +17,12 @@ CsUYapGGPo4dkMgIAUqom/Xajj7h2fB2MPA3j2jxq2I=
 — witness.example.com/w1 jWbPPwAAAABkGFDLEZMHwSRaJNiIDoe9DYn/zXcrtPHeolMI5OWXEhZCB9dlrDJsX3b2oyin1nPZqhf5nNo0xUe+mbIUBkBIfZ+qnA==
 ```
 
+This document specifies two cosignature types: one based on Ed25519, and one
+based on ML-DSA-44. The ML-DSA version SHOULD be used for new deployments, and
```
AlCutter: I think this para could do with a very minor tweak - the key points here are that the doc contains two schemes, and that one of them is strongly recommended for new deployments.

The follow-on text feels like it's either a justification for that recommendation(?), in which case it should probably mention the Q word and could be structured more as a justification, e.g.

```suggestion
based on ML-DSA-44. 

The ML-DSA version SHOULD be used for new deployments. Its signatures:
    - Are quantum-resistant,
    - Commit to the cosigner's name (earlier schemes did not do this),
    - Allow for signing subtrees which don't start at zero (such subtrees don't curr...)
```

Alternatively, perhaps it's conveying detail which might be better located later in the doc?

tlog-cosignature.md
```diff
@@ -115,20 +132,87 @@ the log based on the first three lines of the checkpoint; consensus between
...
+        opaque origin<1..2^8-1>;
+        uint64 start;
+        uint64 end;
+        uint8 root_hash[32];
+    }
+
+`name` is the cosigner name. If the name starts with `oid/`, it MUST be encoded
+as specified below in OID Encoding.
+
+`timestamp` is `timestamped_signature.timestamp`. These two values MAY be zero
```
AlCutter, line 159:
```suggestion
`timestamp` is `timestamped_signature.timestamp`. This value MAY be zero
```

It's the same value in two places, right?

tlog-cosignature.md
```diff
@@ -115,20 +132,87 @@ the log based on the first three lines of the checkpoint; consensus between
...
+`origin` is the log's origin, as represented in a checkpoint's origin line
+without the final newline. If the origin starts with `oid/`, it MUST be encoded
+as specified below in OID Encoding.
+
+`start` is the index of the first leaf included in the [subtree][] being signed.
+If signing a [checkpoint][], it MUST be zero. If `start` is not zero,
+`timestamp` MUST be zero.
+
+`end` is the index of the last leaf included in the [subtree][] being signed,
+plus one. If signing a [checkpoint][], it is the size of the tree.
```
AlCutter: This _", plus one"_ is a bit trixy - as I was reading through I was mentally preparing to write a comment that it was off-by-one, until I got to the "HA - plus one!" punch line :)

This feels like it'd be more naturally expressed by talking about the closed-open `[start, end)` range

tlog-cosignature.md
```diff
@@ -115,20 +132,87 @@ the log based on the first three lines of the checkpoint; consensus between
...
+
+`root_hash` is the root hash of the subtree being signed.
+
+Semantically, a v1 subtree cosignature is a statement that the subtree with the
+specified root hash is consistent with all other historical views observed by
+the cosigner of the log identified by the origin line. If the timestamp is not
+zero, it is also a statement that, as of the specified time, this is the
+consistent tree head with the largest size the cosigner has observed for the log.
```
AlCutter: How about _".. this is the largest consistent tree [state] the cosigner has observed for the log"_? Otherwise we probably need to talk about what a `tree head` is and that feels like it'll add a bit of muddiness.

tlog-cosignature.md
```diff
@@ -115,20 +132,87 @@ the log based on the first three lines of the checkpoint; consensus between
...
+
+Note that checkpoint extension lines are not included in the signed message for
+ML-DSA-44 cosignatures, and no statement is made about them. Subtrees with start
+different from zero currently don't have a checkpoint representation.
```
AlCutter:
```suggestion
ML-DSA-44 cosignatures, and no statement is made about them. Subtrees with non-zero start
values currently don't have a checkpoint representation.
```

tlog-cosignature.md
```diff
@@ -115,20 +132,87 @@ the log based on the first three lines of the checkpoint; consensus between
...
+
+If the cosigner name or log origin line starts with `oid/`, the rest of the
+string MUST be an OID in dotted decimal form, and it MUST be replaced with its
+DER encoding without tag and length, preserving the `oid/` prefix.
+
+For example, the name `oid/2.999.42.1337` would be encoded as
+
+    6f 69 64 2f 88 37 2A 8A 39
+
+This encoding is meant to disambiguate OID-based names from domain-based ones,
```
AlCutter: "URL-based"?

tlog-cosignature.md
```diff
@@ -115,20 +132,87 @@ the log based on the first three lines of the checkpoint; consensus between
...
+while letting verifiers in OID-based ecosystems avoid carrying a decimal OID
+encoder. It is not meant to be represented on the wire: the ASCII name SHOULD be
+used wherever both types may occur, and an appropriate OID encoding (such as DER
+or even a RELATIVE OID encoding) can be used in contexts where only OID-based
```
AlCutter:
```suggestion
used wherever both types may occur, however an appropriate OID encoding (such as DER
or even a RELATIVE OID encoding) may be used in contexts where only OID-based
```

tlog-cosignature.md
```diff
@@ -115,20 +132,87 @@ the log based on the first three lines of the checkpoint; consensus between
...
+include the base cosignature semantics, and MAY include other statements that
 are non-conflicting.  Examples of non-conflicting statements include "I also
-mirrored the log up until the checkpoint size" and "I also reproduced the sparse
-Merkle tree root in the extension lines".  See [tlog-mirror][] for an example.
-
-A cosigner operator that operates multiple cosigners with distinct additional
-statements MUST use distinct public keys for each cosigner. In cosignature/v1,
-it is not sufficient to use a distinct cosigner name.
+mirrored the log up until the checkpoint size", "I also reproduced the sparse
+Merkle tree root in the extension lines", and "I certify the SAN ←→ public key
```
AlCutter: Should we remove the example based on extension lines since the recommended-for-new-deployments scheme cannot commit to those lines in its signature?

tlog-cosignature.md
```diff
@@ -115,20 +132,87 @@ the log based on the first three lines of the checkpoint; consensus between
...
+        opaque name<1..2^8-1>;
+        uint64 timestamp;
+        opaque origin<1..2^8-1>;
+        uint64 start;
+        uint64 end;
+        uint8 root_hash[32];
+    }
+
+`name` is the cosigner name. If the name starts with `oid/`, it MUST be encoded
+as specified below in OID Encoding.
```
davidben:
```suggestion
as specified below in [OID Encoding](#oid-encoding).
```

tlog-cosignature.md
```diff
@@ -115,20 +132,87 @@ the log based on the first three lines of the checkpoint; consensus between
...
+
+`timestamp` is `timestamped_signature.timestamp`. These two values MAY be zero
+if the cosigner doesn't make any statement as to the tree being the largest
+observed at time of signing.
+
+`origin` is the log's origin, as represented in a checkpoint's origin line
+without the final newline. If the origin starts with `oid/`, it MUST be encoded
+as specified below in OID Encoding.
```
davidben:
```suggestion
as specified below in [OID Encoding](#oid-encoding).
```

tlog-cosignature.md
```diff
@@ -115,20 +132,87 @@ the log based on the first three lines of the checkpoint; consensus between
...
+string MUST be an OID in dotted decimal form, and it MUST be replaced with its
+DER encoding without tag and length, preserving the `oid/` prefix.
+
+For example, the name `oid/2.999.42.1337` would be encoded as
+
+    6f 69 64 2f 88 37 2A 8A 39
+
+This encoding is meant to disambiguate OID-based names from domain-based ones,
+while letting verifiers in OID-based ecosystems avoid carrying a decimal OID
+encoder. It is not meant to be represented on the wire: the ASCII name SHOULD be
```
davidben: Hehehe. That works for me, though I'm also happy to convert to decimal. (MTC is somewhat stuck using decimal in the issuer names right now, though that was mostly because of https://github.com/golang/go/issues/75260. Dunno if other stacks have a similar limitation.)

tlog-cosignature.md
```diff
@@ -115,20 +132,87 @@ the log based on the first three lines of the checkpoint; consensus between
...
+A cosigner operator that operates multiple Ed25519 cosigners (e.g. with distinct
+additional statements, see below) MUST use distinct public keys for each
+cosigner. The Ed25519 signed message format doesn't commit to the cosigner name,
+so the same public key can't be used across multiple cosigners.
+
+## ML-DSA-44 signed message
+
+The signed message MUST be a `cosigned_message` structure.
+
+    struct cosigned_message {
```
davidben: Hmm, this is an existing issue elsewhere in the document, but the TLS presentation language is kinda goofy. Structs are written:

```
struct {
   ...
} cosigned_message;
```

https://www.rfc-editor.org/rfc/rfc8446#section-3.6

tlog-cosignature.md
Hayden-IO: I'd consider justifying why ML-DSA-44 over one of the other parameter choices.

tlog-cosignature.md
```diff
@@ -115,20 +132,87 @@ the log based on the first three lines of the checkpoint; consensus between
...
+        uint8 label[12] = "subtree/v1\n\0";
```
Hayden-IO, line 148: Why `subtree/v1` for the identifier rather than `cosignature/v2`, since this can be used for witnessing or MTC?

tlog-cosignature.md
```diff
@@ -17,6 +17,12 @@ CsUYapGGPo4dkMgIAUqom/Xajj7h2fB2MPA3j2jxq2I=
 — witness.example.com/w1 jWbPPwAAAABkGFDLEZMHwSRaJNiIDoe9DYn/zXcrtPHeolMI5OWXEhZCB9dlrDJsX3b2oyin1nPZqhf5nNo0xUe+mbIUBkBIfZ+qnA==
 ```
 
+This document specifies two cosignature types: one based on Ed25519, and one
+based on ML-DSA-44. The ML-DSA version SHOULD be used for new deployments, and
+it's extended to also commit to the cosigner's name, and to allow signing
+subtrees that don't start at zero. (Those subtrees don't currently have a
```
Hayden-IO: Should we define subtree linking to PLANTS earlier on?

I also find it a little strange to refer to non-MTC logs as subtrees as well. Maybe stating `...and to allow the signing of [subtrees](https://datatracker.ietf.org/doc/html/draft-ietf-plants-merkle-tree-certs-02#name-subtrees) in addition to [checkpoints](http://c2sp.org/tlog-checkpoint)`?

tlog-cosignature.md
```diff
@@ -115,20 +132,87 @@ the log based on the first three lines of the checkpoint; consensus between
...
+The signed message MUST be a `cosigned_message` structure.
+
+    struct cosigned_message {
+        uint8 label[12] = "subtree/v1\n\0";
+        opaque name<1..2^8-1>;
+        uint64 timestamp;
+        opaque origin<1..2^8-1>;
+        uint64 start;
+        uint64 end;
+        uint8 root_hash[32];
```
davidben: Couple other minor suggestions as I put together an MTC counterpart:

* Perhaps s/root_hash/subtree_hash/. root_hash could be read either as "the root of the specified subtree" or "the root of the whole tree", which is a bit confusing. (Or just hash?)

* Perhaps s/name/cosigner_name/ and s/origin/log_origin/ so it's more obvious which one is which.

davidben:
I tossed together a first iteration of an MTC version of this at https://github.com/ietf-plants-wg/merkle-tree-certs/compare/tlog-compatible

tlog-cosignature.md
AlCutter: +1 on `cosigner_name` and `log_origin`.

I don't find `root_hash` confusing _personally_, but `hash` seems ok too.

tlog-cosignature.md
FiloSottile: Because it's a more general message than cosignature/v1, and because it's weird to put in MTC a v2 of something that doesn't have a v1 in the lineage of that document's normative references.

Anyway, it's just a label, it just needs to be unique and well-defined.

tlog-cosignature.md
FiloSottile: Yes, but "This value" can be misinterpreted as `timestamp` can be zero if [...] even when `timestamped_signature.timestamp` is not zero, while I can't think of a way to misinterpret the current text even if it reads a bit more weirdly. Picking weird over potentially unclear.

niels-moller (commented):
This MR does several different things. Adding a new (hopefully) quantum-resistant signature scheme, adding support for subtree signatures (where I'm not at all familiar with the details), and a completely new approach to constructing the message to be signed (if I understood it correctly, old way prepends namespace string and a few text lines to the checkpoint, new way is a binary format that doesn't include any of the checkpoint literally).

For the choice of signature algorithm, I think that is worthy of some discussion independently of the other parts of the MR. E.g., SLH-DSA-*-128S might be a conservative choice, but at a cost of a factor 3 in signature size (and there are like other tradeoff factors too). So it would be nice with a rationale.

tlog-cosignature.md
niels-moller: I find it a bit confusing both to treat extension lines differently depending on signature type. If we don't want to sign them here, given that there's no deployment of extension lines, can we retrofit the old cosignature type to also not sign extension lines, to get a bit more consistency? I take it log signatures (which witnesses are expected to verify) will still include the extension lines.

I also find it a bit weird to on one hand piggy-back on the signed-note spec, and on the other hand attach a signature line to the note that actually doesn't cover all of the signed-node body.

If we foresee more tweaks to the sign message, would make sense to include some kind of id to the timestamped_signature struct which goes into the signature line? The signature type byte will start to feel small if we add a few more post-quantum algorithms as well as additional flavors of witness claims.

FiloSottile:
> SLH-DSA-*-128S might be a conservative choice, but at a cost of a factor 3 in signature size (and there are like other tradeoff factors too). So it would be nice with a rationale.

Due to a combination of factors, ML-DSA is getting all the deployment attention, which is a virtuous cycle of implementation availability, testing, and optimization. Go is not planning any other PQ signature in the standard library. Wycheproof doesn't have test vectors for any other PQ signatures and no one asked for them.

Some of the reasons are the tightening deadlines that make it necessary to pick one answer, the generally sweet(er) spot of size/security/performance tradeoffs, the CNSA 2.0 regulatory pressure, the direction of the PKIs, the ease of implementation, and the lack of a specification for Falcon.

I added a bit of rationale about the parameter set choice to the document, but I think a long discussion on the merits of different PQ signatures is out of scope. There is definitely demand for a ML-DSA cosignature from the MTC and Go ecosystems, at least. If there is strong demand for a SLH-DSA cosignature, it can be added, but I don't see it replacing the ML-DSA one, so it shouldn't block it.

FiloSottile:
I have addressed all the review comments, PTAL.

niels-moller:
> I added a bit of rationale about the parameter set choice to the document, but I think a long discussion on the merits of different PQ signatures is out of scope.

I agree a long discussion is out of place. I was thinking more of something laying out which properties (like signature size, pubkey size, sign/verify performance) seem most important in this context. But if the argument boils down to ML-DSA is what everyone else wants, then I guess that may be enough motivation.
 
> If there is strong demand for a SLH-DSA cosignature, it can be added, but I don't see it replacing the ML-DSA one, so it shouldn't block it.

Fair enough. Using SLH-DSA would be a bit fun though, since there a signature is essentially the concatenation of a bunch of merkle-tree inclusion proofs, and the pubkey is the expected root hash.

tlog-cosignature.md
FiloSottile: Note that semantically they are treated the same way. Also, `subtree/v1` is intended to be also adopted as a log signature.

> If we don't want to sign them here, given that there's no deployment of extension lines, can we retrofit the old cosignature type to also not sign extension lines, to get a bit more consistency? I take it log signatures (which witnesses are expected to verify) will still include the extension lines.

Unless you think it should block this PR, do you mind opening an issue? We should do at least some due diligence on whether truly there are no deployments that would break.

I honestly don't have an opinion either way, I remember this was discussed at length and IMHO it was discussed for longer than its importance justifies already. (I don't even remember on what side of that I was!)

> I also find it a bit weird to on one hand piggy-back on the signed-note spec, and on the other hand attach a signature line to the note that actually doesn't cover all of the signed-node body.

That's why cosignature/v1 included them, IIRC, absent a strong reason to do otherwise. The strong reason for subtree/v1 is that it doesn't reuse the tlog-checkpoint format at all for the signed body, to make cryptographic-layer interop with binary-only ecosystems like MTC easier.

Another way to look at it is that signed-note is a general way to serialize signatures that don't need to be signing the message body 1:1. There is precedent in c2sp.org/static-ct-api.

> If we foresee more tweaks to the sign message, would make sense to include some kind of id to the timestamped_signature struct which goes into the signature line? The signature type byte will start to feel small if we add a few more post-quantum algorithms as well as additional flavors of witness claims.

I suggest we start worrying about that when we cross 128 assigned/reserved codepoints. Until then, it's nice to have only one joint, less switching, less confusion, fewer opportunities for mistakes.

FiloSottile:
> Using SLH-DSA would be a bit fun though, since there a signature is essentially the concatenation of a bunch of merkle-tree inclusion proofs, and the pubkey is the expected root hash.

Heh, yeah, as a Merkle tree enthusiast I was pretty excited about SLH-DSA, too. Alas, it doesn't look like it's "happening" and the implementation maturity and availability effects easily outweigh the rest even if you only consider the security consequences.

tlog-cosignature.md
niels-moller: > Note that semantically they are treated the same way. Also, `subtree/v1` is intended to be also adopted as a log signature.

Ah, I didn't know that. Fun that we're going full circle and back to signing binary messages.
 
> Unless you think it should block this PR, do you mind opening an issue? We should do at least some due diligence on whether truly there are no deployments that would break.

Filed https://github.com/C2SP/C2SP/issues/238 for reference. (Not entirely sure if that change is a good idea, but I dislike the inconsistency).

tlog-cosignature.md
```diff
@@ -115,20 +135,87 @@ the log based on the first three lines of the checkpoint; consensus between
...
+
+## ML-DSA-44 signed message
+
+The signed message MUST be a `cosigned_message` structure.
+
+    struct {
+        uint8 label[12] = "subtree/v1\n\0";
+        opaque cosigner_name<1..2^8-1>;
+        uint64 timestamp;
+        opaque log_origin<1..2^8-1>;
```
niels-moller, line 151: Would it make sense to use hashes for cosigner_name and log_origin? 

Pro: 
 * We get a fixed-size signed message.
 * We don't need to introduce a new arbitrary limit on the length of names.
 
Cons:
* A little extra processing.
* We have to specify the hash function to use.

tlog-cosignature.md
niels-moller: Do you expect to have to encode/decode the oids? I would have expected that oids are treated as constant magic strings, and then encoding doesn't matter much.

I think it's a bit unfortunate if the spec requires witnesses to know about the special oid/ prefix. A fully opaque string would be friendlier, no matter if the contents of that string is ascii, raw binary, or whatever.

FiloSottile: The options are doing the mapping like this in cosigners, or making clients that deal in binary OIDs also carry either a decimal encoder or two magic strings for each cosigner. (There is no parsing involved in either case.)

There are many orders of magnitude more clients than cosigner, so I optimized for complexity in clients.

niels-moller: I would strongly prefer to to keep complexity down in the highly trusted witnesses, even if they are fewer. That said, I think two magic strings would be ok, can you elaborate on how that would work on the witness side? I'm thinking of a witness signer running on a device like a TKey or USB Armory, getting an add-checkpoint request (or some equivalent representation) from the untrusted outside, checking consistency with internal state, and producing a cosignature if everything is ok. Would that witness signer get both representations of a log's name from outside, or which one? Are there any security implications if mapping is done outside of the trusted device that holds the signing key?

To be clear, I'm concerned mainly with *log* names and how to handle them when cosigning. For the witness' own name, I'd expect it's easy (only slightly annoying) for the witness to keep one magic string representing its name in the signed data, and a different magic string sent over the wire.

tlog-cosignature.md
```diff
@@ -115,20 +135,87 @@ the log based on the first three lines of the checkpoint; consensus between
...
+    6f 69 64 2f 88 37 2A 8A 39
+
+This encoding is meant to disambiguate OID-based names from URL-based ones,
+while letting verifiers in OID-based ecosystems avoid carrying a decimal OID
+encoder. It is not meant to be represented on the wire: the ASCII name SHOULD be
+used wherever both types may occur, however an appropriate OID encoding (such as
+DER or even a RELATIVE OID encoding) may be used in contexts where only
+OID-based names are expected.
+
+## Additional statements
```
mhutchinson, line 186: I can't tell from this PR whether logs that want to sign in a PQ-resistant way should be using this cosignature scheme, or another scheme that hasn't been added yet (see my other comment).

If they are using this scheme:
 - I suggest that we explicitly call out here what additional statements logs make:
    - The log hasn't  _published_ anything inconsistent (rather than just "observed").
    - The log makes all the data implied by this size available (modulo tlog-tiles pruning).
 - Logs themselves no longer sign extension lines. Seems significant to mention that... somewhere?
 - The name "**co**signature" is stretched a little.

signed-note.md
```diff
@@ -121,11 +121,14 @@ runtime if possible.
...
 * `0x03` — Reserved.
 
-* `0x04` — Timestamped Ed25519 witness cosignatures, as specified by
+* `0x04` — Timestamped Ed25519 checkpoint cosignatures, as specified by
   c2sp.org/tlog-cosignature.
 
 * `0x05` — RFC 6962 `TreeHeadSignature`s, as specified by c2sp.org/static-ct-api.
 
+* `0x06` — Timestamped ML-DSA-44 (sub)tree cosignatures, as specified by
+  c2sp.org/tlog-cosignature.
```
mhutchinson, line 130: Is there going to be a more generic ML-DSA-44 (or other PQ resistant scheme) added here for generic notes (e.g. the example message at the bottom of this file) ?

As this section grows, some structured data to help select a sig type would probably be useful (e.g. some will be insecure, some are domain-specific). No action needed on this point right now though.

mhutchinson (commented):
TL;DR: of my comments: does this apply to the sigs that logs themselves put on checkpoints?

The PR description is currently empty. I think it would benefit from the context on _why_ this change is added (I worked it out, but it took several steps).

tlog-cosignature.md
davidben: I don't feel strongly either way. Would we feel less anxious about the limit if this were a 2-byte length prefix instead of a 1-byte prefix?

tlog-cosignature.md
lukevalenta: This paragraph should probably be updated as well to account for subtree signing.
```suggestion
the consistency of a [checkpoint][] or [subtree][]. Log clients can verify a quorum of
```

tlog-cosignature.md
niels-moller: I think a one byte prefix (max length 255) should be enough in practice. My itch is that this doesn't look like the right place to *introduce* such a limit.

tlog-cosignature.md
niels-moller: I think it would be best treat log signature semantics as out of scope in this spec. But let me outline my understanding of log signatures: They state that a given tree head belongs to a log with the given origin line (which is why management of origin line - log pubkey is important; and sigsum made the shortcut of using the pubkey hash as the log identity). The log signing key represents the authority to decide what goes into a log identified by an origin line. Sure, there are some additional claims made by the log, like the signed tree being consistent, and the underlying data being available, but to me that is only weakly related to a particular log signature: we're not going to trust the log on those claims based on a log's signature on some tree head, they have to be backed up by other parties such as witnesses, monitors and mirrors. 

One point of view is that log signatures are merely an authentication mechanism for the witness' add-checkpoint api; from this point of view, we don't even need to include log signatures in a cosigned checkpoint distributed to verifiers. It's important that monitors and verifiers agree on the origin line(s) of relevant logs, but they could delegate log signature verification to witnesses entirely, and not care about log pubkeys and signatures.

But on the other hand, including the log signature proves to verifiers and monitors that the log was involved in the process; the witness quorum is not given authority to make up new tree heads for the origin line all on their own. This is a somewhat obscure scenario, but to me, this is why I think it's prudent to keep the log signature on checkpoints and tlog proofs, and require that it is verified.

davidben: I'm not sure I see why this is inconsistent with unifying them. They both sign the same kinds of stuff, and they're both making the same kinds of statements. "This is on my particular view of the tree" plus the additional statements specific to the role.

It happens that one role is "this is on my particular view of the tree... because I am authoritative for what the tree even is". That is indeed quite an important role! And almost certainly the signature consumer is going to treat that role as special in its policy, but that's fine. We can still unify signature formats and have different roles.

MTC unifies the two and it's been quite tidy IMO.

Though in terms of defining the semantics of the "log owner" (co)signer role and this spec, my inclination is that something like tlog-checkpoint is a good place to list those out.

tlog-cosignature.md
davidben: I mean we can use this thing if we really care:
https://www.rfc-editor.org/rfc/rfc9420.html#section-2.1.2

But I think, if we believe 255 is enough, that's fine. If we're unsure, we could just bump it to 2^16-1 and not worry about it. I guess I'm just used to TLS stuff but I'm generally not too concerned about just picking an arbitrary limit in this sort of thing.

(My weak preference would be to either leave it alone or bump to two bytes. But my strong preference is to just pick something, so we can go ahead and nail down the MTC format. I can live with any of these. 🙂)

niels-moller: I strongly prefer to not blow up the worst-case size. I want to be able to use a small simple hardware device doing all the cryptographic parts of the witness cosigning.

davidben: I don't see how any of these blow up sizes. What are you referring to?

The existing format has no size limits at all. Small hardware devices were already prepared to verify a payload that contains the log names directly. If those devices depended on a bound on the payload, they would have had to set some bound on the log origin on their own. (This is very reasonable!) They can set that same bound here. Making the binary format use 1 or 2 or 8 bytes of length prefix doesn't obligate you to work with that size of string.

Plus the maximum sizes here don't come from the format. They come from:
* The size of your own name. That's a constant.
* The largest log origin in your config.

Also keep in mind that ML-DSA (unlike Ed25519!) is streamable and supports externally computing hashes, so they now have even more options to work with. But given the incoming requests themselves need to say the log origin, I find it hard to believe the witness would truly never just have that available in memory. It is ultimately not a long string.

So I think we shouldn't hash it and just leave it as is. This doesn't seem to buy anything.

niels-moller: The current PR has a max name length of 255, and a worst case max `cosigned_message` size of 588 bytes (if I count the fields correctly). Increasing max name length to 65535 increases worst case size considerably. Using a hash (my suggestion that started this thread) would give a fixed size of 132 bytes (with 32-byte hash digests) or 196 for 64-byte digests.

So if people don't buy the name hash suggestion, please just leave as is.

(Sure, my witness implementation can set its own arbitrary limits, but then that may surprise a log operator that has chosen some overly verbose origin line and discover later that their choice breaks interop with me. If it's easy to support *everything* that is within spec, that is preferable. And we also aim to auto configure witnesses by subscribing to the appropriate witness-network.org lists, which means witness operators are not in tight control of the configured log origin lines).

niels-moller: And you're right that current wire format versions (both tlog-checkpoint and tlog-witness) set no maximum size for names. I think that is a weakness; implementation are bound to have their own arbitrary limits, which will differ and may lead to interop surprises.

tlog-cosignature.md
mhutchinson: It's becoming clear that this is a PR-scoped comment. With hindsight perhaps this isn't the ideal location to have attached the comment, but at least we're having the conversation :-)

I agree that:
  - this signature scheme does appear suitable for log signing
  - `tlog-checkpoint` is probably a better place to document it, and then link it from this doc perhaps.

My understanding is that with this PR merged and all specs considered at HEAD, a tlog-checkpoint _could_ be signed with one of these signatures. i.e. the checkpoint spec says "Logs SHOULD use Ed25519 signatures to sign the checkpoint, but MAY use any note signature algorithm based on the ecosystem they operate in.". As this is now listed in `signed-note`, I think that opens the door?

Let's ignore the claims stuff for a minute. I believe that these changes to the body of specs mean that a log that uses extension lines and upgrades from Ed25519 to ML-DSA-44 does so in a way that is technically permitted by the specs. But in doing this, their extension lines are no longer signed. I'm not saying that this is wrong or bad, but I don't think it's clear that picking a different signature fundamentally changes what is signed in your note. It's also possible that downstream consumers of checkpoints would be surprised to find the log signature being a cosignature.

In the interests of not blocking this PR, I'm happy to accept an issue against tlog-checkpoint to say that we should address this documentation before cutting another release of that spec. Or that signed-note should be clear about the different features of each signature - some are useful for any bytes, where some require structured data, and some only sign a subset of that data.

signed-note.md
mhutchinson: This comment is largely vestigial, but I'll leave it here as a forward pointer to the thread that has taken its place: https://github.com/C2SP/C2SP/pull/237#discussion_r3085767779

tlog-cosignature.md
FiloSottile: I have a preference against hashes simply because it leads to a discussion on what hash function to use, and there are two "obvious" answers: SHA-256 like the Merkle tree or SHA-3 like ML-DSA.

I agree this should not be the place where log origin size limits stem from long term, but we can go back and fix that in tlog-checkpoint. I have opened #242. (tlog-cosignature is the right place to limit the size of the cosigner name, though.)

Sounds like we're generally happy with a 255 max, so let's keep that.

signed-note.md
FiloSottile: I am not sure what other ecosystems use signed-note, if there's adoption we should indeed define an ML-DSA signature type soon, but it doesn't block this PR.

I don't want the list of codepoints to become load-bearing beyond avoiding collisions. The signed-note document can recommend the right generic one, and leave the ecosystem-specific recommendations to other documents.

tlog-cosignature.md
FiloSottile: I updated the recommendation in tlog-checkpoint and mentioned the fact that extension lines are not covered.

FiloSottile:
Addressed the comments, this is ready to merge, @C2SP/tlog-cosignature and @C2SP/tlog-checkpoint please ACK.

AlCutter (approved):

davidben (approved):

lukevalenta (approved):

FiloSottile:
@rgdd added a line to tlog-witness to recommend these new cosignatures and to specify that when obtained over the tlog-witness protocol, they must include the whole tree and the timestamp.

rgdd (approved):
LGTM, thanks everyone for your work and input!

(On a separate timeline that doesn't block merge, it would be good to write down
a concise summary of the thoughts that went into these changes.  So that others
who look at this now or in the future doesn't have to re-discover the thought
process.  I'll see if I can help with this; perhaps around next week or so.)

tlog-cosignature.md
```diff
@@ -17,6 +17,15 @@ CsUYapGGPo4dkMgIAUqom/Xajj7h2fB2MPA3j2jxq2I=
...
 ```
 
+This document specifies two cosignature types: one based on Ed25519, and one
+based on ML-DSA-44. The ML-DSA version SHOULD be used for new deployments.
+
+Unlike the Ed25519 type, the ML-DSA-44 type is secure against quantum computers.
+Moreover, it commits to the cosigner's name, and supports signing [subtrees][]
+in addition to [checkpoints][checkpoint]. The ML-DSA-44 parameter set was
+selected because at NIST Level 2 it provides some margin beyond the 128-bit
+security level.
```
thomwiggers, line 27: You're leaving unsaid here that ML-DSA-44 is the smallest/most efficient option.

PR #239, opened 2026-04-12
age-plugin: mention "mlkem768x25519" as a native recipient stanza

olastor:
I believe mentioning the new type here would make sense because "x25519" is explicitly mentioned, as well.


FiloSottile (approved):
Thank you!

str4d (approved):

PR #244, opened 2026-04-23
WIP: Identity-based transparency log spec

Hayden-IO:
An early exploration for creating a unified leaf format between Sigstore and Sigsum logs for identity-based/key usage/signature transparency (name TBD).

### DIFF
--- a/identity-transparency.md
+++ b/identity-transparency.md
@@ -1,0 +1,80 @@
+# Identity-based Transparency Logging Spec
+
+TODO: Add section on background and motivation, including:
+
+* Background on existing solutions, with an opportunity to share tooling and have redundant log operators
+* Minimizing the entry size, especially given ML-DSA's large public key and signature size
+* Removing log poisoning risk
+* Providing ecosystem-specific context strings
+
+## Conventions used in this document
+
+The base64 encoding used throughout is the standard Base 64 encoding specified
+in [RFC 4648][], Section 4. The hex encoding of a positive integer is the fixed
+length lowercase Base 16 encoding of its zero-padded big endian representation.
+
+`U+` followed by four hexadecimal characters denotes a Unicode codepoint, to be
+encoded in UTF-8. `0x` followed by two hexadecimal characters denotes a byte
+value in the 0-255 range. `||` denotes concatenation.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in [BCP 14][] [RFC 2119][] [RFC
+8174][] when, and only when, they appear in all capitals, as shown here.
+
+[RFC 4648]: https://www.rfc-editor.org/rfc/rfc4648.html
+[BCP 14]: https://www.rfc-editor.org/info/bcp14
+[RFC 2119]: https://www.rfc-editor.org/rfc/rfc2119.html
+[RFC 8174]: https://www.rfc-editor.org/rfc/rfc8174.html
+[RFC 8032]: https://www.rfc-editor.org/rfc/rfc8032.html
+
+## Leaf Entry Format
+
+The leaf entry is a sequence of bytes. The leaf entry MUST start with a single
+byte version identifier, and MUST be followed by repeated 32-byte SHA-256
+digests.
+
+Example leaf entry, where `||` denotes concatenated values.
+
+```
+0x01 || Hash(message) || Hash(Root of trust) || Hash(key1) || Hash(value1) || Hash(key2) || Hash(value2) || ... || Hash(keyN) || Hash(valueN)
+```
+
+The first byte MUST be the byte `0x01`. Future versions of this leaf format
+MUST increment this byte value.
+
+The next 32 bytes MUST be a digest of the `message` submitted to the log.
+`message` MUST be the digest of the original data. `message` is hashed again to
+prevent log poisoning where entries contain spam, e.g. spam is split into byte
+arrays with each chunk uploaded as a message. Logs MUST reject any messages
+that are not 32 bytes.
+
+The next 32 bytes MUST be the digest of a root of trust. The root of trust
+SHOULD be a public key, for example the signer's public key or a public key
+from a root certificate. The encoding of the public key is ecosystem-dependent.
+
+The remainder of the leaf entry is optional.
+
+The following bytes are a key-value mapping of ecosystem-specific strings. A
+32-byte digest of a key name MUST be followed by the 32-byte digest of a value.
+It is recommended to apply a restriction on the number of key-value pairs to
+prevent large leaf entries.
+
+TODO: Should we document why the leaf doesn't contain the signature, and the
+(optional?) responsibility by the log operator to persist signatures for
+non-repudiation
+
+## Signature Format
+
+For `data` to be signed, the client MUST use `SHA256(data)` as the `message` to
+be submitted to the log.
+
+The `signature` is computed over the concatenation of the log entry version
+`0x01` (as a domain separator demonstrating the intent to log signed data), the
+NUL character `0x00`, and the checksum `SHA256(message)=SHA256(SHA256(data))`.
+
+Any signature algorithm may be used.
+
+## Verification
+
+TODO: Do we need a section on how to verify an entry?
--- a/identity-transparency.md
+++ b/identity-transparency.md
@@ -67,9 +67,11 @@ The following bytes are a key-value mapping of ecosystem-specific strings. A
 32-byte digest of the digest of a key name MUST be followed by the 32-byte
 digest of the digest of a value. The key-value strings are double-hashed to
 allow the log to accept either the key/value, or the digest of the key/value
-if the key or value is large.
+if the key or value is large or should not be revealed to the log.
 It is recommended to apply a restriction on the number of key-value pairs to
 prevent large leaf entries.
+Key/value pairs MUST be sorted lexicographically by the double-digest of the
+key.
 
 The last 32 bytes MUST be the digest of a _receipt_. The receipt provides
 non-repudiation for the log operator, proving the log received a valid
@@ -93,8 +95,6 @@ context strings
 The signature algorithm MUST be one of the following:
 
 * ML-DSA-44
-* ML-DSA-65
-* ML-DSA-87
 * Ed25519
 
 ## Root of Trust & Receipt
@@ -110,8 +110,6 @@ and [RFC 9881][] for Ed25519 and ML-DSA respectively. The `root of trust` MUST
 be prefixed by a signature algorithm identifier, either:
 
 * `ML-DSA-44`
-* `ML-DSA-65`
-* `ML-DSA-87`
 * `Ed25519`
 
 The `receipt` is the signature, whose encoding is defined in [RFC 8410][] and
--- a/identity-transparency.md
+++ b/identity-transparency.md
@@ -43,10 +43,10 @@ repeated values:
 0x01 ||
 Hash(root of trust) ||
 Hash(message) ||
-Hash(key1) || Hash(value1) ||
-Hash(key2) || Hash(value2) ||
+Hash(Hash(key1)) || Hash(Hash(value1)) ||
+Hash(Hash(key2)) || Hash(Hash(value2)) ||
 ... ||
-Hash(keyN) || Hash(valueN) ||
+Hash(Hash(keyN)) || Hash(Hash(valueN)) ||
 H(receipt)
 ```
 
@@ -64,7 +64,10 @@ SHOULD be a public key, for example the signer's public key or a public key
 from a root certificate. The encoding of the public key is ecosystem-dependent.
 
 The following bytes are a key-value mapping of ecosystem-specific strings. A
-32-byte digest of a key name MUST be followed by the 32-byte digest of a value.
+32-byte digest of the digest of a key name MUST be followed by the 32-byte
+digest of the digest of a value. The key-value strings are double-hashed to
+allow the log to accept either the key/value, or the digest of the key/value
+if the key or value is large.
 It is recommended to apply a restriction on the number of key-value pairs to
 prevent large leaf entries.
 
@@ -77,15 +80,15 @@ time limit on how long a receipt is retained.
 
 ## Signature Format
 
-For `data` to be signed, the client MUST use `SHA256(data)` as the `message` to
-be submitted to the log.
+For `data` to be signed, the client MUST use `H(data)` as the `message` to
+be submitted to the log, where `H` is `SHA-256`.
 
 The `signature` is computed over the concatenation of the specification
 identifier `c2sp.org/identity-transparency/v1` (as a domain seperator
 demonstrating the intent to log signed data), the NUL character `0x00`,
-the checksum `SHA256(message)=SHA256(SHA256(data))`, and the ecosystem-specific
+the checksum `H(message)=H(H(data))`, and the ecosystem-specific
 context strings
-`SHA256(key1)||SHA256(value1)||...||SHA256(keyN)||SHA256(valueN)`.
+`H(H(key1))||H(H(value1))||...||H(H(keyN))||H(H(valueN))`.
 
 The signature algorithm MUST be one of the following:
 
--- a/identity-transparency.md
+++ b/identity-transparency.md
@@ -26,18 +26,28 @@ document are to be interpreted as described in [BCP 14][] [RFC 2119][] [RFC
 [BCP 14]: https://www.rfc-editor.org/info/bcp14
 [RFC 2119]: https://www.rfc-editor.org/rfc/rfc2119.html
 [RFC 8174]: https://www.rfc-editor.org/rfc/rfc8174.html
-[RFC 8032]: https://www.rfc-editor.org/rfc/rfc8032.html
+[RFC 8410]: https://www.rfc-editor.org/rfc/rfc8410.html
+[RFC 9881]: https://www.rfc-editor.org/rfc/rfc9881.html
 
 ## Leaf Entry Format
 
 The leaf entry is a sequence of bytes. The leaf entry MUST start with a single
-byte version identifier, and MUST be followed by repeated 32-byte SHA-256
-digests.
+byte version identifier.
 
-Example leaf entry, where `||` denotes concatenated values.
+### Version 0x01
 
-```
-0x01 || Hash(message) || Hash(Root of trust) || Hash(key1) || Hash(value1) || Hash(key2) || Hash(value2) || ... || Hash(keyN) || Hash(valueN)
+Example leaf entry, where `||` denotes concatenated values and `...` denotes
+repeated values:
+
+```text
+0x01 ||
+Hash(root of trust) ||
+Hash(message) ||
+Hash(key1) || Hash(value1) ||
+Hash(key2) || Hash(value2) ||
+... ||
+Hash(keyN) || Hash(valueN) ||
+H(receipt)
 ```
 
 The first byte MUST be the byte `0x01`. Future versions of this leaf format
@@ -45,35 +55,78 @@ MUST increment this byte value.
 
 The next 32 bytes MUST be a digest of the `message` submitted to the log.
 `message` MUST be the digest of the original data. `message` is hashed again to
-prevent log poisoning where entries contain spam, e.g. spam is split into byte
-arrays with each chunk uploaded as a message. Logs MUST reject any messages
-that are not 32 bytes.
+prevent log poisoning where entries contain unwanted bytes, e.g. meaningful
+content is split into byte arrays with each chunk uploaded as a message. Logs
+MUST reject any messages that are not 32 bytes.
 
 The next 32 bytes MUST be the digest of a root of trust. The root of trust
 SHOULD be a public key, for example the signer's public key or a public key
 from a root certificate. The encoding of the public key is ecosystem-dependent.
-
-The remainder of the leaf entry is optional.
 
 The following bytes are a key-value mapping of ecosystem-specific strings. A
 32-byte digest of a key name MUST be followed by the 32-byte digest of a value.
 It is recommended to apply a restriction on the number of key-value pairs to
 prevent large leaf entries.
 
-TODO: Should we document why the leaf doesn't contain the signature, and the
-(optional?) responsibility by the log operator to persist signatures for
-non-repudiation
+The last 32 bytes MUST be the digest of a _receipt_. The receipt provides
+non-repudiation for the log operator, proving the log received a valid
+_credential_ to upload to the log. The receipt MUST be stored outside of the
+entry bundle, due to its size. The log MAY NOT generate or the receipt, and in
+this case, the 32 bytes MUST only be NUL characters `0x00`. The log MAY set a
+time limit on how long a receipt is retained.
 
 ## Signature Format
 
 For `data` to be signed, the client MUST use `SHA256(data)` as the `message` to
 be submitted to the log.
 
-The `signature` is computed over the concatenation of the log entry version
-`0x01` (as a domain separator demonstrating the intent to log signed data), the
-NUL character `0x00`, and the checksum `SHA256(message)=SHA256(SHA256(data))`.
+The `signature` is computed over the concatenation of the specification
+identifier `c2sp.org/identity-transparency/v1` (as a domain seperator
+demonstrating the intent to log signed data), the NUL character `0x00`,
+the checksum `SHA256(message)=SHA256(SHA256(data))`, and the ecosystem-specific
+context strings
+`SHA256(key1)||SHA256(value1)||...||SHA256(keyN)||SHA256(valueN)`.
 
-Any signature algorithm may be used.
+The signature algorithm MUST be one of the following:
+
+* ML-DSA-44
+* ML-DSA-65
+* ML-DSA-87
+* Ed25519
+
+## Root of Trust & Receipt
+
+The root of trust and receipt depends on the _credential_ submitted to the log
+to demonstrate ownership of an identity. Additional credentials may be added
+in future revisions.
+
+### Credential: Public Key
+
+The `root of trust` is the PKIX ASN.1 `SubjectPublicKeyInfo` as per [RFC 8410][]
+and [RFC 9881][] for Ed25519 and ML-DSA respectively. The `root of trust` MUST
+be prefixed by a signature algorithm identifier, either:
+
+* `ML-DSA-44`
+* `ML-DSA-65`
+* `ML-DSA-87`
+* `Ed25519`
+
+The `receipt` is the signature, whose encoding is defined in [RFC 8410][] and
+[RFC 9881][].
+
+### Credential: OpenID Connect Identity Token
+
+The `root of trust` is the issuer of the token, as a string. The value SHOULD
+come from the `iss` JWT claim or MAY come from a specified claim populated by a
+federated issuer.
+
+The receipt is a zero-knowledge proof over the token signature.
+
+## Context Key-Value Pairs
+
+TODO: Add Sigsum context string
+
+TODO: Populate from [Sigstore OIDs](https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#directory)
 
 ## Verification
 
--- a/identity-transparency.md
+++ b/identity-transparency.md
@@ -53,15 +53,15 @@ H(receipt)
 The first byte MUST be the byte `0x01`. Future versions of this leaf format
 MUST increment this byte value.
 
+The next 32 bytes MUST be the digest of a root of trust. The root of trust
+SHOULD be a public key, for example the signer's public key or a public key
+from a root certificate. The encoding of the public key is ecosystem-dependent.
+
 The next 32 bytes MUST be a digest of the `message` submitted to the log.
 `message` MUST be the digest of the original data. `message` is hashed again to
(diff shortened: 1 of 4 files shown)


identity-transparency.md
```diff
@@ -0,0 +1,80 @@
...
+## Signature Format
+
+For `data` to be signed, the client MUST use `SHA256(data)` as the `message` to
+be submitted to the log.
+
+The `signature` is computed over the concatenation of the log entry version
+`0x01` (as a domain separator demonstrating the intent to log signed data), the
+NUL character `0x00`, and the checksum `SHA256(message)=SHA256(SHA256(data))`.
+
+Any signature algorithm may be used.
```
Hayden-IO: From @kommendorkapten: "From a specification perspective, I think it makes sense to be agnostic here. But for real world deployments and interoperability I think a more prescriptive list of algorithms would make sense."

identity-transparency.md
```diff
@@ -0,0 +1,80 @@
...
+arrays with each chunk uploaded as a message. Logs MUST reject any messages
+that are not 32 bytes.
+
+The next 32 bytes MUST be the digest of a root of trust. The root of trust
+SHOULD be a public key, for example the signer's public key or a public key
+from a root certificate. The encoding of the public key is ecosystem-dependent.
+
+The remainder of the leaf entry is optional.
+
+The following bytes are a key-value mapping of ecosystem-specific strings. A
```
Hayden-IO, line 66: We should be opinionated on the sorting of the ecosystem-specific strings. Sort by pre-image or sort by digest?

identity-transparency.md
```diff
@@ -0,0 +1,80 @@
...
+Example leaf entry, where `||` denotes concatenated values.
+
+```
+0x01 || Hash(message) || Hash(Root of trust) || Hash(key1) || Hash(value1) || Hash(key2) || Hash(value2) || ... || Hash(keyN) || Hash(valueN)
+```
+
+The first byte MUST be the byte `0x01`. Future versions of this leaf format
+MUST increment this byte value.
+
+The next 32 bytes MUST be a digest of the `message` submitted to the log.
```
Hayden-IO, line 60: Note, `message MUST be the digest of the original data` is currently a recommendation in the Sigsum specification. Can we make this a requirement instead? This is of course not verifiable, but any well-formed client should follow the spec.

identity-transparency.md
```diff
@@ -0,0 +1,80 @@
...
+TODO: Should we document why the leaf doesn't contain the signature, and the
+(optional?) responsibility by the log operator to persist signatures for
+non-repudiation
+
+## Signature Format
+
+For `data` to be signed, the client MUST use `SHA256(data)` as the `message` to
+be submitted to the log.
+
+The `signature` is computed over the concatenation of the log entry version
```
Hayden-IO: Note: This doesn't work for DSSEs which [define](https://github.com/secure-systems-lab/dsse/blob/master/protocol.md) a signature over a PAE.

Hayden-IO: From @kommendorkapten: "Do we wanna create an extension for DSSE, or a transparency flavoured DSSE version?

0x1 || 0x00 || sha256(PAE)?"

Hayden-IO: Ideally we don't have to involve the DSSE project to support a transparency-related extension.

Generalizing, the issue is that an ecosystem has a specific format for what to sign over. One idea is we could have a leaf entry identifier that signals to the log that the signature is computed directly over the digest, which would be `SHA256(PAE)`.

identity-transparency.md
Hayden-IO: Agreed, and I know this flexibility is in contrast to the tlog-cosignature specification. I'd say it's up to the ecosystem to determine how to specify this. Maybe we can formalize the algorithm registry proposal?

identity-transparency.md
```diff
@@ -0,0 +1,80 @@
...
+## Leaf Entry Format
+
+The leaf entry is a sequence of bytes. The leaf entry MUST start with a single
+byte version identifier, and MUST be followed by repeated 32-byte SHA-256
+digests.
+
+Example leaf entry, where `||` denotes concatenated values.
+
+```
+0x01 || Hash(message) || Hash(Root of trust) || Hash(key1) || Hash(value1) || Hash(key2) || Hash(value2) || ... || Hash(keyN) || Hash(valueN)
```
rgdd: As mentioned in https://git.glasklar.is/sigsum/project/documentation/-/blob/main/archive/2025-09-03-leaf-context-for-sigsum-v2.md (opt#1) and discussed a bit already, sigsum is missing a signing context. 

This signing context needs to be part of both the leaf and the signed message. I see several ways of making that happen. Most straight forward would be to have it as a mandatory value, i.e., always have it in the leaf and the signed message. The other option is to have it in extension lines, and ensure that there are semantics for knowing that how the message is signed depends on that extension.

Any initial thoughts on how you'd prefer to see this integrated? Before I write a wall I'll make one pass with a few comments, so moving on for now!

identity-transparency.md
rgdd: I think it's hard to phrase it as a MUST when there's no way of enforcing? But 100% agree that any well behaved client/lib should be sending a hash, and not 32 arbitrarily chosen bytes even though the log can't (generally) distinguish the two.

identity-transparency.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+The first byte MUST be the byte `0x01`. Future versions of this leaf format
+MUST increment this byte value.
+
+The next 32 bytes MUST be a digest of the `message` submitted to the log.
+`message` MUST be the digest of the original data. `message` is hashed again to
+prevent log poisoning where entries contain spam, e.g. spam is split into byte
```
rgdd: I think of spam as: the log getting many entries. And poison as: bad/unwanted bytes in a leaf. So we need both anti-poison (the hashing) and anti-spam (e.g. filter when we get too many submissions from someone that is filling up the log with many entries). OK if we rephrase spam here?

identity-transparency.md
rgdd: What are the extensions that we intent to go here right now? I think it would make sense to enumerate them as a starting point, regardless of if they should go in the spec or not (will help us and others who are chipping in with input).

identity-transparency.md
```diff
@@ -0,0 +1,80 @@
...
+The remainder of the leaf entry is optional.
+
+The following bytes are a key-value mapping of ecosystem-specific strings. A
+32-byte digest of a key name MUST be followed by the 32-byte digest of a value.
+It is recommended to apply a restriction on the number of key-value pairs to
+prevent large leaf entries.
+
+TODO: Should we document why the leaf doesn't contain the signature, and the
+(optional?) responsibility by the log operator to persist signatures for
+non-repudiation
```
rgdd: WDYT about a way for the leaf to encode + commit to there being (or not being) a signature? I think that would make things more robust in the case of the log being expected to store the signature as extra data outside of the leaf. I also have a gut feeling that it might make sense to then also commit H(sig) to the Merkle tree leaf if storage is expected. Might fit as an extension line, where the value is H(sig). Another variation/flavor of this could be: storage is expected for a duration T, so that it's possible to show that the log didn't fabricate the entry for at least T time units before it's dropped.

I see how this doesn't make sense in the case of a root of trust that's an OIDC sigstore CA; but I think it makes a lot of sense in the case of just a free standing key pair.

identity-transparency.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Signature Format
+
+For `data` to be signed, the client MUST use `SHA256(data)` as the `message` to
+be submitted to the log.
+
+The `signature` is computed over the concatenation of the log entry version
+`0x01` (as a domain separator demonstrating the intent to log signed data), the
```
rgdd: Version byte is a bit too vague to be a strong domain separator. More robust option: "c2sp.org/<spec-name>\0" + leaf version byte or similar. Exactly which constant doesn't matter as long as it's more tied to this spec and version of the semantics we're describing.

identity-transparency.md
rgdd: This needs to be specified somewhere, but I'm not 100% sure where 'somewhere' is yet. But I'd say this should most likely be tied to the root of trust hash, i.e., if we have Ed25519 signing we would probably compute the root of trust as H(AlgID=Ed25519 || 32 Ed25519 pubkey bytes) or similar. This also encourages keeping pubkey + sigalg in the same place and tied together, as opposed to the sig alg being something that is 'discovered' as part of an unauthenticated message somewhere.

Can we start small and enumerate here to begin with, and then we see where that takes us? My preference would be to keep this to a minimum and expand as necessary. ML-KEM (same variation as tlog-cosignature) seems like the obvious thing to specify and then add everything else on a case-by-case basis?

identity-transparency.md
```diff
@@ -0,0 +1,80 @@
...
+[RFC 4648]: https://www.rfc-editor.org/rfc/rfc4648.html
+[BCP 14]: https://www.rfc-editor.org/info/bcp14
+[RFC 2119]: https://www.rfc-editor.org/rfc/rfc2119.html
+[RFC 8174]: https://www.rfc-editor.org/rfc/rfc8174.html
+[RFC 8032]: https://www.rfc-editor.org/rfc/rfc8032.html
+
+## Leaf Entry Format
+
+The leaf entry is a sequence of bytes. The leaf entry MUST start with a single
+byte version identifier, and MUST be followed by repeated 32-byte SHA-256
```
rgdd:
```suggestion
byte version identifier. For version 0x01, the version byte MUST be followed by repeated 32-byte SHA-256
```

(or similar -- i.e., let's phrase it a bit better than this uninspired clarification.)

rgdd (commented):
Dropped a few in-line comments and suggestions. Happy to help with edits, but step one is probably a bit of discuss so starting with comments.

Another comment/question, I'm pondering how extension key-value pairs should be framed.

Right now it seems like they are unsigned? That makes me view it as 'claims' by the log, e.g., "I store the signature with this hash -- just ask me and I'll prove it". Or in the case of a log that is also doing OIDC stuff, "this comes from this particular github user" or similar.

I like the above framing.

Another framing is "it's arbitrary metadata and the log just puts it there, possibly tampering with it before putting it into the leaf". Feels worse.

Another one is "it's arbitrary metadata put by the signer and the log can't tamper with it", e.g., by having all key-value extension pairs signed. Again, then it feels better. But I'm not sure if it's warranted to be that permissive (and instead cap by some suitable limit). So I'm curious if there are real uses for this already, and if so, what are they?

As for the question I saw in 1-2 two places -- should we document 'why'. Yes, somewhere. Maybe not in the spec long-term, but I'd say it makes sense to put such pointers here for now at least.

Have to sign off for now but back tomorrow -- so starting with this initial round of comments to get the conversation going here!

identity-transparency.md
kommendorkapten: Thinking more about this, I think it would be disruptive to each ecosystem to require a signature to be domain separated to be added before computing the signature?

The verification flow for a client thus needs to branch of depending on if the artifact was intended to be tlogged or not (e.g. in Sigstore there _may_ be private instances that don't operate a tlog), and as stated certain ecosystems already define the signature format (like in-toto/DSSE). How do we feel about dropping this requirement?

rgdd: > I think it would be disruptive to each ecosystem to require a signature to be
> domain separated to be added before computing the signature?

Is it not already disruptive regardless of if there is a namespace or not?

I.e., isn't DSSE's current signature convention something different than "given
input message m, sign H(H(m))".  If it's already disruptive, then we might as
well have a strong namespace (and a poison-free signing context)?

> The verification flow for a client thus needs to branch of depending on if the
> artifact was intended to be tlogged or not

Is this also not already the case?  Either the tool is used in transparency-log
enforcing mode and configured with a trust policy for this, or it isn't.

Or if the tooling is not intended for transparency, why take it into account?

> (e.g. in Sigstore there may be private instances that don't operate a tlog)

My understanding is that tlog would be mandatory in the new architecture, since
the leaves would be signature-less and the sigstore CA would just sign the root?

> and as stated certain ecosystems already define the signature format (like
> in-toto/DSSE). How do we feel about dropping this requirement?

Awaiting some replies to threads before I weight in with an actual opinion.
But a warning, I do see the risk that we're talking about different things.
So it would be good to clarify if we're talking about things that mainly affect
sigstore CA write path API (ACL), or the leaf and monitors (e.g., because
we're talking about key-value extensions related to storing and verifying
signatures; which is the lense I look at by default because...sigsum :-)).

identity-transparency.md
```diff
@@ -0,0 +1,80 @@
...
+The next 32 bytes MUST be a digest of the `message` submitted to the log.
+`message` MUST be the digest of the original data. `message` is hashed again to
+prevent log poisoning where entries contain spam, e.g. spam is split into byte
+arrays with each chunk uploaded as a message. Logs MUST reject any messages
+that are not 32 bytes.
+
+The next 32 bytes MUST be the digest of a root of trust. The root of trust
+SHOULD be a public key, for example the signer's public key or a public key
+from a root certificate. The encoding of the public key is ecosystem-dependent.
+
```
Hayden-IO: From a few offline chats: Should we record the hash of the signature? While it's not a value that would be monitored, it would be a commitment from the log that it received a valid signature. If a signer were to claim they didn't generate a signature, then the log could produce a signature (stored off-tree) using that hash value as a lookup into CAS.

identity-transparency.md
loosebazooka: I guess on the topic of sha256, should this be configurable at all? Is sha256 safe for the foreseeable future (for this usecase)?

I imagine we can just change the whole leaf format if the hashing function changes.

identity-transparency.md
loosebazooka: I like the idea of being opinionated, or even strict per ecosystem. Even with a well defined json structure we ran into some issues with sigstore early on. Allowing anyone to define anything here puts in a precarious inter-op position. This is probably especially hard on client tooling developers.

The leaf format may be independent of the user's persistence format (ex: sigstore bundle) though, where the type is specified there instead of by the log -- and the whole (in this example) sigstore ecosystem will be responsible for defining its leaf format in its own spec. Although, this might end up worsening the problem for monitors. 

The opaqueness of this format has the potential to it harder for someone somewhere, but I'm having trouble defining that for now.

rgdd:
Last week we had a Sigsum community gathering in Stockholm. Among other things we discussed this PR.

TL;DR notes:
https://git.glasklar.is/sigsum/project/documentation/-/blob/main/archive/2026-05-12-sigsum-next-meetup-notes

identity-transparency.md
Hayden-IO: From chat with Filippo: Call this a "receipt", make it ecosystem-dependent and not specified here. e.g. for keys, it's a hash of a signature, for tokens, a zkp, for certs, the full chain.

identity-transparency.md
Hayden-IO: Hm, I had initially thought that the context would be included only in the log entry, not as part of the signature input. I was going to then suggest that the context be an ecosystem-dependent string, same as Sigstore will have provenance values as optional strings.

Could we have different signature format identifiers? For example `0x01` is intent to log in a transparency log, `0x02` is intent to log and use an additional value (maybe provided in `extra` data in tlog-proof?). We were thinking about supporting `0x00` as well, which would be a signature with no transparency.

identity-transparency.md
Hayden-IO: I'm good with keeping this set minimal. Maybe we start with Ed25519 and ML-DSA, to align with the tlog-cosignature spec? For our use case, we're only planning to support this spec with ML-DSA, but I could see someone asking for a classical-crypto-flavored variant in the near term to help ease the transition to PQC.

identity-transparency.md
Hayden-IO: For Sigstore, the extensions will be from https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#directory. The log will map provider-specific claims from an identity token to this standardized list. 

I didn't include this list here because I didn't want to couple Sigstore-specific values to the more general leaf format spec. I was thinking we would have a Sigstore leaf specification and a Sigsum leaf spec that extends this specification. Where those would live long-term, I'm not sure - could be in C2SP or in each project's spec repo.

We also may want to handle certificates at some point, for example for logging Authenticode certs, but this is very much TBD. I'd want to define a set of OIDs to extract from the certificates, rather than extracting all OIDs. Even if we constrain the max number of extensions, it still muddies the monitoring story if we don't have a well-defined set.

identity-transparency.md
Hayden-IO: I thought this could be `MUST` to signal that a well-behaving client should send a hash. Though I wrote "should" in my comment above (and just now!), so maybe we relax this since it's not enforceable. No strong opinion here.

identity-transparency.md
Hayden-IO: Yes, because tiles-tlog hardcodes SHA256. If it breaks, we'd need revisions of many of these specs.

identity-transparency.md
Hayden-IO: Yep, happy with those definitions, will rephrase!

rgdd:
Top-posting, TL;DR: hayden and I have been going through the comments and tried
to figure out what we think is the current wip sketch for the upcoming leaf.

    0x01          // leaf version byte -- defines what byte 2 and forward is
	H(RoT)        // Root of Trust -- namespaces different submitters
    H(H(data))    // double hash -- what is the thing being logged
	[H(H(key_1))] // double hash -- optional extension key
	[H(H(val_1))] // double hash -- optional extension value
	...
	[H(H(key_n))]
	[H(H(val_n))]
	H(receipt)    // receipt is, e.g., a signature or a ZK proof

Whether the receipt is stored or not is "log policy".

At least two RoT types:

* "Signature / pubkey" --> recepit is a signature, RoT is probably something
  like H("ML-DSA-44" + pubkey_bytes)
* "Signature-less" --> receipt is empty string for now, maybe ZK in the future

In the case of RoT being "Signature / pubkey", signed message is roughly:

    C2SP.org/<spec-name>/v1  // protocol namespace
    H(H(data))               // what is signed
    [H(H(key_1))]            // optional key, double hashed (conservative)
    [H(H(val_1))]            // optional value
    ...
    [H(H(key_n))]
    [H(H(val_n))]

Next steps:

* Hayden will take a stab at updating based on what we were sketching here
* New top-level passes, i.e., starting new threads based on where we're at now
* RoT definitions we're happy with + framing on how to add more later if needed?
* Initial set of extensions we're happy with + framing on how to add more later if needed?
  * signing context
  * sigstore's oid extensions
  * (we think we can maybe frame extensions as: should be useful for monitor,
	otherwise out of scope. But still discussing / very early thoughts here.)

rgdd:
Carve out from a conversation in the sigsum room, related to the above TL;DR and RoT. Just to keep track of this design note.

> rgdd:
> My thinking was that the log has a policy for storing the receipts for all leaves, and whether they are signatures, ZK proofs, or something else doesn't matter that much?
I.e., a monitor can download these opaque bytes, and check that H(opaque bytes) matches H(receipts) for the log's retention policy. The monitor anyway won't know if it's a valid ML-DSA-44 signature since it doesn't have the pubkey to verify.
> If we want to do better then some upper max limit that is 'large enough', i.e., something more finer grained, then yes we would need to move those 'type bytes' from RoT to the leaf.
> And monitors that don't know newly introduced 'type bytes' wouldn't know how to deal with the extra data pertaining to these entries (or would have to have some upper limit, like now)
Not sure what's best here -- think it warrants a bit more discussion and thinking
>
> nisse:
> If a "type byte" is moved from inside the RoT to an unhashed leaf header, another consequence is that it becomes obvious to outsiders which types are accepted by the log, and distribution of types, usage evolving over time, etc. Not really clear to me how useful that is or if it for some reason is undesirable to be transparent about that metadata, but my gut feeling is that it would be nice to make that visible. We'd need a process for managing the type namespace, but I think we'd need that anyway (since at a minimum each tlog needs to agree with its users, and the id also needs to be expressed in some way in the submission api).

identity-transparency.md
Hayden-IO: We decided to use the DSSE's pre-auth encoding as the message, so a signature would be over `c2sp.org/<id>/v1 || 0x00 || Hash(Hash("DSSEv1" + SP + LEN(type) + SP + type + SP + LEN(body) + SP + body))`, so this spec does not need to be aware of other domains.

It's not a perfect solution, since an off-the-shelf DSSE verifier won't work. DSSE has an experimental concept of extensions, so we could define a transparent flavor of a DSSE `signature`. But this is out of scope for this spec.

identity-transparency.md
Hayden-IO: Added a subheader for `0x01`, so we can append new versions in the future.

identity-transparency.md
Hayden-IO: As discussed offline, added as a receipt.

identity-transparency.md
Hayden-IO: Added!

Hayden-IO:
I've updated the spec! I went through all of the previous feedback, all of it should now be addressed. I've also updated the leaf format based on https://github.com/C2SP/C2SP/pull/244#issuecomment-4615035651. And as I type this, I realize I didn't include the double-hashed context strings. I'll add that shortly.

identity-transparency.md
```diff
@@ -0,0 +1,136 @@
...
+The following bytes are a key-value mapping of ecosystem-specific strings. A
+32-byte digest of the digest of a key name MUST be followed by the 32-byte
+digest of the digest of a value. The key-value strings are double-hashed to
+allow the log to accept either the key/value, or the digest of the key/value
+if the key or value is large.
```
Hayden-IO: Note to self: We need to define an ordering.

identity-transparency.md
```diff
@@ -0,0 +1,136 @@
...
+content is split into byte arrays with each chunk uploaded as a message. Logs
+MUST reject any messages that are not 32 bytes.
+
+The next 32 bytes MUST be the digest of a root of trust. The root of trust
+SHOULD be a public key, for example the signer's public key or a public key
+from a root certificate. The encoding of the public key is ecosystem-dependent.
+
+The following bytes are a key-value mapping of ecosystem-specific strings. A
+32-byte digest of the digest of a key name MUST be followed by the 32-byte
+digest of the digest of a value. The key-value strings are double-hashed to
```
kommendorkapten, line 68: What is the rationale behind the double hashing? 

From a log poisoning vector, one hash should be enough. Double hashing can be useful to prevent length extension attacks, but that's not needed here as the digests shouldn't be used as MACs.

rgdd: I see how the double hash doesn't add anything in the case of a Sigstore CA that is also the log, but in other cases there can be distinctions that are meaningful.  For example, a Sigsum submitter that provides a signing context should not have to reveal what the signing context is to the log.

So: from a poisoning perspective you're correct if the preimage is provided to the log, and then the log computes the hash. But it would be an unfortunate shortcoming if this spec forced the log operator to observe all preimages to mitigate poisoning attacks. And spec wise, it seems simpler to avoid branches and always compute the extension key-value pairs in the same way. So that's how I'd explain the rationale!

Bonus answer/context: for values double hashing seems essential. For keys, I'm less convinced double hashing is essential and it would likely be fine with a single hash; or even the raw value itself. Especially if we want to frame keys as being a slowly evolving semi-closed set, as opposed to something that can arbitrarily be set by anyone that submits to the log. But even if the key is not meant to be a value worth protecting, a hash might still be a good idea because it makes it easier to deal with the leaf format (i.e., fixed size vs arbitrary size). FWIW I'd be more than open to revisit the 'key' part here, but preferably after making some more progress on the framing wrt. the set of keys / their intended use.

identity-transparency.md
```diff
@@ -0,0 +1,136 @@
...
+demonstrating the intent to log signed data), the NUL character `0x00`,
+the checksum `H(message)=H(H(data))`, and the ecosystem-specific
+context strings
+`H(H(key1))||H(H(value1))||...||H(H(keyN))||H(H(valueN))`.
+
+The signature algorithm MUST be one of the following:
+
+* ML-DSA-44
+* ML-DSA-65
+* ML-DSA-87
```
rgdd: Suggestion: let's keep down the scope and expand it if we must? I'd say let's start with ML-DSA-44, similar to tlog-cosignature.

identity-transparency.md
kommendorkapten: Ok, I think I get it:

1. Submitter computes `d = h(data)` and submits that to the log.
2. Log computes `d' = h(d)` and includes in the leaf. 

Seems right?

rgdd: Yes :+1:

kommendorkapten: That makes perfect sense.

> The following bytes are a key-value mapping of ecosystem-specific strings. A 32-byte digest of the digest of a key name MUST be followed by the 32-byte digest of the digest of a value. The key-value strings are double-hashed to allow the log to **accept either the key/value, or the digest of the key/value if the key or value is large**.

Note: emphasis mine. 

Can this be clarified (that no key/value should be sent, only digests) as now it's a bit ambiguous I think (as if you send the key/value it wouldn't be double hashed).

identity-transparency.md
Hayden-IO: I do think we'll get a request to expand this, but I'm OK with starting with `-44` since the witness network will use those parameters as well.

identity-transparency.md
Hayden-IO: And just to add a bit more about Sigstore - for transparent signatures based on OIDC tokens, we'll double-hash the claims, which yes, isn't necessary for preventing poisoning.

I'm not yet sure what the upload API should look like for context strings for signing with keys, this will have to be hashed out as part of a future discussion on the upload API. Sigstore will want to replicate whatever Sigsum supports, which I assume will just be the digest of the context string, never the context string.

For keys - Yea, I'm also open to revisiting this as well. I'd lean towards a single hash for uniform length. For now, let's stick with double hashing for uniformity with the values.

Hayden-IO: > Can this be clarified (that no key/value should be sent, only digests) as now it's a bit ambiguous I think (as if you send the key/value it wouldn't be double hashed).

I think this wouldn't be a part of the leaf spec, but a part of the upload spec. For Sigstore, we will receive the key/values unhashed, since they're in the OIDC token and we need the prehash values to verify the token signature.

identity-transparency.md
```diff
@@ -0,0 +1,134 @@
...
+The first byte MUST be the byte `0x01`. Future versions of this leaf format
+MUST increment this byte value.
+
+The next 32 bytes MUST be a digest of the `message` submitted to the log.
+`message` MUST be the digest of the original data. `message` is hashed again to
+prevent log poisoning where entries contain unwanted bytes, e.g. meaningful
+content is split into byte arrays with each chunk uploaded as a message. Logs
+MUST reject any messages that are not 32 bytes.
+
+The next 32 bytes MUST be the digest of a root of trust. The root of trust
```
Hayden-IO: Another name than 'root of trust'?

identity-transparency.md
```diff
@@ -0,0 +1,134 @@
...
+come from the `iss` JWT claim or MAY come from a specified claim populated by a
+federated issuer.
+
+The receipt is a zero-knowledge proof over the token signature.
+
+## Context Key-Value Pairs
+
+TODO: Add Sigsum context string
+
+TODO: Populate from [Sigstore OIDs](https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#directory)
```
Hayden-IO, line 130: This list should be standardized and ordered

identity-transparency.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Root of Trust & Receipt
+
+The root of trust and receipt depends on the _credential_ submitted to the log
+to demonstrate ownership of an identity. Additional credentials may be added
+in future revisions.
+
+### Credential: Public Key
+
+The `root of trust` is the PKIX ASN.1 `SubjectPublicKeyInfo` as per [RFC 8410][]
```
Hayden-IO, line 108: Look into an alternative encoding - RFC8032 for Ed25519, review RFC9881 or FIPS204

identity-transparency.md
Hayden-IO: Trust anchor?

niels-moller:
Terminology suggestion: Can we talk about name/value pairs rather then key/value pairs? It's gets confusing to discuss htese things when we have a key/value, where what's stored in the leaf is a "key hash" (i.e., H(key)), while the value can be a different kind of "key hash" (e.g., H(submitter public key)).

identity-transparency.md
```diff
@@ -0,0 +1,134 @@
...
```
Leont, line 108: It may be helpful to specifically mention it should be DER encoded, not just any ASN.1 encoding.

identity-transparency.md
```diff
@@ -0,0 +1,134 @@
...
+and [RFC 9881][] for Ed25519 and ML-DSA respectively. The `root of trust` MUST
+be prefixed by a signature algorithm identifier, either:
+
+* `ML-DSA-44`
+* `Ed25519`
```
Leont, line 113: `SubjectPublicKeyInfo` already encodes an algorithm identifier. This seems unnecessary.

identity-transparency.md
```diff
@@ -0,0 +1,134 @@
...
+
+```text
+0x01 ||
+Hash(root of trust) ||
+Hash(message) ||
+Hash(Hash(key1)) || Hash(Hash(value1)) ||
+Hash(Hash(key2)) || Hash(Hash(value2)) ||
+... ||
+Hash(Hash(keyN)) || Hash(Hash(valueN)) ||
+H(receipt)
```
Leont, line 50: Why this mix of `HASH()` and `H()`? Is that an accident, or am I missing some distinction?

PR #245, opened 2026-04-24
tlog-witness: add optional sign-subtree API

FiloSottile:


davidben (approved):
LGTM!

tlog-witness.md
```diff
@@ -172,6 +180,141 @@ checkpoint must be performed atomically, otherwise the following race can occur:
...
+
+ - With access to the full tree state, by checking that the checkpoint size is
+   less than or equal to the size of the latest checkpoint it cosigned for the
+   checkpoint's origin, and that a consistency proof can be constructed from the
+   latest checkpoint to the submitted checkpoint.
+
+Note that the stateful verification mechanisms above only require read access to
+the witness state and can operate on stale state safely.
+
+If all the checks above pass, the witness MUST respond with a "200 Success" HTTP
```
lukevalenta: > If all the checks above pass

This could be interpreted as doing _all_ of the three checks for verifying that the checkpoint is consistent with the witness's view of the log. (I think it's _probably_ clear enough already that you just need to do one of those checks, but just flagging in case we can tighten it up.)

lukevalenta: Also probably need to work in a "403 Forbidden" if any of the verification checks fail (and we didn't respond with a "409 Conflict").

tlog-witness.md
```diff
@@ -172,6 +180,141 @@ checkpoint must be performed atomically, otherwise the following race can occur:
...
+   submitted checkpoint against those. If the checkpoint is not known, the
+   witness MAY respond with a "409 Conflict" HTTP status code, and the response
+   body MUST consist of a known checkpoint for the same origin. In this case,
+   the witness SHOULD remember and accept the last few signed checkpoints, to
+   minimize conflicts.
+
+ - With access to the full tree state, by checking that the checkpoint size is
+   less than or equal to the size of the latest checkpoint it cosigned for the
+   checkpoint's origin, and that a consistency proof can be constructed from the
+   latest checkpoint to the submitted checkpoint.
```
lukevalenta: I think this is backwards?
```suggestion
   submitted checkpoint to the latest checkpoint.
```

tlog-witness.md
```diff
@@ -172,6 +180,141 @@ checkpoint must be performed atomically, otherwise the following race can occur:
...
+The witness MUST verify that the checkpoint is consistent with its view of the
+log identified by the origin line. It MAY do so in multiple ways:
+
+ - Statelessly, by checking a cosignature from its own key on the submitted
+   checkpoint.
+
+ - With access to one or more recently signed checkpoints, by comparing the
+   submitted checkpoint against those. If the checkpoint is not known, the
+   witness MAY respond with a "409 Conflict" HTTP status code, and the response
+   body MUST consist of a known checkpoint for the same origin. In this case,
```
lukevalenta: Do we need to specify Content-Type of `text/x.tlog.signed-checkpoint` or something here?

tlog-witness.md
```diff
@@ -172,6 +180,141 @@ checkpoint must be performed atomically, otherwise the following race can occur:
...
+Unprocessable Entity" HTTP status code.
+
+The witness MUST verify that the checkpoint is consistent with its view of the
+log identified by the origin line. It MAY do so in multiple ways:
+
+ - Statelessly, by checking a cosignature from its own key on the submitted
+   checkpoint.
+
+ - With access to one or more recently signed checkpoints, by comparing the
+   submitted checkpoint against those. If the checkpoint is not known, the
```
lukevalenta: Do we need to cache the cosignatures along with these and return them in the "409 Conflict" response? (I think probably not)

tlog-witness.md
lukevalenta: Also maybe want to add a "If the submitted checkpoint size matches the latest checkpoint size, the witness MUST check that the root hashes are identical." Maybe this is implied since the witness can verify equality by constructing an empty consistency proof.

tlog-witness.md
```diff
@@ -172,6 +180,141 @@ checkpoint must be performed atomically, otherwise the following race can occur:
...
+The consistency proof lines MUST encode a Subtree Consistency Proof from the
+subtree to the checkpoint according to
+[draft-ietf-plants-merkle-tree-certs-03][], Section 4.4. If the Merkle
+Consistency Proof doesn't verify, the witness MUST respond with a "422
+Unprocessable Entity" HTTP status code.
+
+The witness MUST verify that the checkpoint is consistent with its view of the
+log identified by the origin line. It MAY do so in multiple ways:
+
+ - Statelessly, by checking a cosignature from its own key on the submitted
```
lukevalenta:
```suggestion
 - Statelessly, by checking a cosignature from one of its own keys on the submitted
```

tlog-witness.md
```diff
@@ -172,6 +180,141 @@ checkpoint must be performed atomically, otherwise the following race can occur:
...
+h/ghL5nyaXTIBAklqDnisM+SFdbA3izoGZ5oRYMZzc0=
+8qtnAAU9sTKQx/zt3E4v8Jt2a9IMQUf0QUawvePF9y0=
+KuO5BNOmginG34Lm5G3e2GuiWliGo3K0ebzGpvI0fEg=
+
+example.com/behind-the-sofa
+14
+/CcKVZM9n65aH7jLaIZuUB4/MSlEFQ4+ldwUC21rDHM=
+```
+
+start MUST be less than end, and end MUST be less than or equal to the
```
lukevalenta: I suppose proof verification will fail if this isn't the case, but we might want to just do all the subtree checks here.
```suggestion
The half-open interval `[start, end)` MUST be a valid subtree per [draft-ietf-plants-merkle-tree-certs-03][] section 4.1, and end MUST be less than or equal to the
```

tlog-witness.md
```diff
@@ -172,6 +180,141 @@ checkpoint must be performed atomically, otherwise the following race can occur:
...
+
+Note that the stateful verification mechanisms above only require read access to
+the witness state and can operate on stale state safely.
+
+If all the checks above pass, the witness MUST respond with a "200 Success" HTTP
+status code. The response body MUST be a sequence of one or more [note][]
+signature lines for the subtree, each starting with the `—` character (U+2014)
+and ending with a newline character (U+000A). The signatures SHOULD be ML-DSA-44
+[cosignatures][], and SHOULD be from one or more of the same witness key(s) that
+signed the checkpoint. If the cosignature format supports timestamps, the
```
lukevalenta:
```suggestion
signed the checkpoint. The cosignature format MUST support subtree cosigning. If the cosignature format supports timestamps, the
```

tlog-witness.md
```diff
@@ -172,6 +180,141 @@ checkpoint must be performed atomically, otherwise the following race can occur:
...
+cosignatures from unknown keys. The client MUST NOT send more than 8 subtree
+cosignature lines.
+
+Each consistency proof line MUST encode a single hash in base64. The client MUST
+NOT send more than 63 consistency proof lines.
+
+The client MUST NOT send more than 8 checkpoint signatures. The client MAY omit
+the signatures entirely, if the witness is expected to statefully verify the
+checkpoint. If signatures are omitted, the body MUST terminate with the newline
+character that concludes the checkpoint body, without empty lines.
```
lukevalenta:
```suggestion
character that concludes the checkpoint body, without a trailing empty line.
```

tlog-witness.md
```diff
@@ -172,6 +180,141 @@ checkpoint must be performed atomically, otherwise the following race can occur:
...
+checkpoint size. Otherwise, the witness MUST respond with a "400 Bad Request"
+HTTP status code.
+
+The consistency proof lines MUST encode a Subtree Consistency Proof from the
+subtree to the checkpoint according to
+[draft-ietf-plants-merkle-tree-certs-03][], Section 4.4. If the Merkle
+Consistency Proof doesn't verify, the witness MUST respond with a "422
+Unprocessable Entity" HTTP status code.
+
+The witness MUST verify that the checkpoint is consistent with its view of the
```
lukevalenta:
```suggestion
If the checkpoint origin is unknown, the witness MUST respond with a "404
Not Found" HTTP status code. The witness MUST verify that the checkpoint is consistent with its view of the
```

rgdd:
Top-posting a few higher-level comments (also sync:ed with @niels-moller):

* It's nice that the add-checkpoint API is used to move the witness state
  forward, whereas the cosign-subtree API is used to just collect subtree
  cosignatures.  And the timestamp semantics are then clearly defined as
  always add timestamp (add-checkpoint) vs never add timestestamp (subtree).
* The overall shape of the HTTP POST API is mostly looking good.  We have
  one comment regarding the optional subtree cosignature lines (see below),
  but other than that the inputs feel consistent for cosign-subtree compared to
  the inputs we have in add-checkpoint.  The semantics are also easy to explain:
  someone wants to request a cosignature on a specified subtree, and that
  subtree has to be contained in a specified reference checkpoint.
* It stands out to define *a little bit of DoS mechanims* directly in the
  HTTP POST input lines, while then also saying there are other DoS mechanisms
  and they are out of scope for the purpose of this document.  And it seems a
  bit unfortunate to define something that *only* works for the cosign-subtree
  API.  How about we define an Appendix for this instead, which applies for both
  add-checkpoint and cosign-subtree?  And if anyting new needs to be defined,
  then we define it there.  Otherwise just list options + link.  For example,
  the subtree cosignatures DoS mechanism could be an HTTP header; similar to how
  Sigsum moved this from the input key-value pairs to be optional in headers.
  FWIW sigsum's [rate-limit mechanism][] would probably work here, except that
  maybe it would be a bit weird to use `_sigsum_v1.` as the left-most label.
* What breaks / who will be sad if we simplify the stateful-processing
  description?  The two options we think make the most sense are: (1) witness
  operates stateless, needs to see its own cosignature, or (2) witness keeps
  track of the latest cosigned checkpoint, and if no cosignature is provided
  then the reference checkpoint must correspond to the witness' latest state.
  And it is up to the log to remember this, and to retry if it got out of sync.
  This pattern of the second option is similar to "old" + HTTP 409 status code.
* For the DoS subtree cosiging cosignature construct: it makes sense to have a
  limit.  But maybe it should be 1?  I.e., the expectation is to know which
  "ticket" to send, and if it's wrong retry with a valid ticket?
* For the upper limit of checkpoint signatures: not sure if it is necessary
  to cap it to 8.  How about something like "SHOULD only be cosignatures made by
  the witness", since the point is to prove to the witness it already cosigned?
  Then the number of cosignatures that can be sent for that witness is more or
  less capped already based on how many keys the witness is cosigning with.
  * (By not capping it, it could also allow for sign-if-logged type setups in
    the future, i.e., where a witness cosigns iff there's a broader quorum.
    Motivation would be to reduce the number of cosignatures, to save space.)
  * (And if we want a limit defined, maybe that's for the tlog-checkpoint spec?)

A few nit-picks:

* "If the cosignature format supports timestamps, the timestamp MUST be zero."
  Can we instead clearly refer to using tlog-cosignature, and drop the "if"
  part and go straight to clarifying that the timestamp MUST be zero here?
* API name: maybe s/sign-subtree/cosign-subtree/

[rate-limit mechanism]: https://git.glasklar.is/sigsum/project/documentation/-/blob/main/log.md

davidben:
(I don't have direct experience running any of these services, so take these as thoughts from someone who was involved in sketching out the original version of this, but does not have strong feelings and probably is not going to die on any particular hill.)

> It stands out to define a little bit of DoS mechanims directly in the
> HTTP POST input lines, while then also saying there are other DoS mechanisms
> and they are out of scope for the purpose of this document. And it seems a
> bit unfortunate to define something that only works for the cosign-subtree
> API.

There's some difference between `cosign-subtree` and `add-checkpoint` in that `add-checkpoint` is naturally bound by the origin log. If a witness is up-to-date, only the origin log can cause the witness to update its state and sign something, AIUI. Whereas `cosign-subtree` is an oracle that can sign whatever old subtree it likes. And so some of those mitigations were to try to mitigate the subtree-oracle-specific DoS concerns and try to make it look more like `add-checkpoint`.

(How much that matters, no idea!)

> What breaks / who will be sad if we simplify the stateful-processing
> description? The two options we think make the most sense are: (1) witness
> operates stateless, needs to see its own cosignature, or (2) witness keeps
> track of the latest cosigned checkpoint, and if no cosignature is provided
> then the reference checkpoint must correspond to the witness' latest state.
> And it is up to the log to remember this, and to retry if it got out of sync.
> This pattern of the second option is similar to "old" + HTTP 409 status code.

Tracking the _very latest_ checkpoint feels like it might get you into a livelock. Suppose you're not the log operator and you're trying to get this one subtree signed, but the log operator and witness keep churning the log forward several times a seconds. In order to win the race, you would need to get your 409 and reply in between one of these updates. 

`add-checkpoint` is different in that, every time you lose the race, forward progress has been made. So for any given target tree position, either you will eventually win the race, or the other guy will catch the witness up to where you wanted anyway.

Also if we define it to be truly the _latest_ checkpoint, you arguably need to take a lock on the witness state, but actually a stale one is also fine. Since stale ones are actually fine, that suggests that the witness could track the last handful and increase the window for the race condition.

The third option ("With access to the full tree state...") is redundant with the others, but feels like an easy option for mirrors and log operators who might want to implement this protocol. Then we don't have to think about this at all.

> For the upper limit of checkpoint signatures: not sure if it is necessary to cap it to 8. [...] (And if we want a limit defined, maybe that's for the tlog-checkpoint spec?)

FWIW, `add-checkpoint` also carries a checkpoint and we seem not to have felt a need to define a limit there.

rgdd:
> There's some difference between cosign-subtree and add-checkpoint in that
> add-checkpoint is naturally bound by the origin log. If a witness is
> up-to-date, only the origin log can cause the witness to update its state and
> sign something, AIUI.

No, anyone with access to the add-checkpoint endpoint can replay the most recent
checkpoint and get back a fresh cosignature (i.e., timestamp updated).  So they
are not so different after all if someone runs a witness with a public endpoint.

> Suppose you're not the log operator and you're trying to get this one subtree
> signed

Is that a deployment model you are expecting?

It's not how deployment works today, hence why I'm asking these questions to
figure out what breaks for you if we would trim a bit. To be clear I am also not
strongly opinionated here, but before giving a thumbs up to something that
looks a bit more complicated it's important to have the motivation down.

(The easiest option to spec and implement would be to always just send a
reference checkpoint with a witness cosignature. No race to speak of, but
I take it you are hoping to not have to send signatures to save bandwidth.)

> add-checkpoint is different in that, every time you lose the race, forward
> progress has been made. So for any given target tree position, either you will
> eventually win the race, or the other guy will catch the witness up to where
> you wanted anyway.

If it's different or not depends on who's expected to query.

In the case where the log operator collects cosignature this is a non-problem
for both endpoints.  In the case that someone else than the log operator
collects, then we have the same challenge with both endpoints.  The collector
has a target tree and they want enough cosignature to reach a quorum for it.

davidben:
> No, anyone with access to the add-checkpoint endpoint can replay the most recent
> checkpoint and get back a fresh cosignature (i.e., timestamp updated). So they
> are not so different after all if someone runs a witness with a public endpoint.

...oh! Okay, then I agree there's no need for any subtree-specific DoS considerations here.

> Is that a deployment model you are expecting?

For MTC, it's the moral equivalent of you submitting your certificate to some specific CT log yourself, rather than expecting the CA to do it. I don't expect it to be common, but I think it's useful to have in a pinch. (E.g. maybe if some old relying party really needs this witness, but your CA of choice has stopped bothering for you.)

(I mean, why have the conflict protocol at all if only log operators might use the endpoint?)

> (The easiest option to spec and implement would be to always just send a
> reference checkpoint with a witness cosignature. No race to speak of, but
> I take it you are hoping to not have to send signatures to save bandwidth.)

Eh, I'm not worried about the bandwidth cost here. I don't feel particularly strongly and a model where you just expect to check your own signature seems fine? Also the third option (just check your copy of the tree) makes this moot for mirrors at least.

rgdd:
> (I mean, why have the conflict protocol at all if only log operators might use
> the endpoint?)

Example: log sends a valid request to a witness and dies/aborts before it gets
back a response. The witness however persisted and cosigned the new log state.
The next time the log sends a request but for an even later tree state, they
will be out of sync because the witness is further ahead than the log knows.

So it's here to be able to recover from this kind of out-of-sync scenarios.

> Also the third option (just check your copy of the tree) makes this moot for
> mirrors at least.

But then it is not just the signature which is moot. In other words, for a mirror
it would be sufficient to get a log origin and a subtree range as input.

---

Do I understand correctly that you'd be happy if the API always just required at
least one checkpoint cosignature from the witness when using `cosign-subtree`?

Motivated by:

* We avoid a branch in how witnesses MAY verify the incoming `cosign-subtree`
  request.  This makes it easier to implement for most requesters and witnesses,
  and still doesn't rule out having a witness that due to other reasons are
  operating in a stateful manner (e.g., there might be extra cosigning claims).
* We avoid having conflict resolution in the `cosign-subtree` API.  Either we
  have a valid reference checkpoint cosigned by the witness, or we don't.
* The benefit of supporting the stateful option(s) is to save a signature, but we
  are not worried about the extra bandwidth/compute in this particular context.
  Or a different take might be: if this turns out to be a problem, then we can
  just extend this specification when it's needed. It's backwards compatible.
* We're anyway not removing enough of the inputs as worded right now,
  e.g., for a mirror consistency proof and other inputs are redundant too.

davidben:
> Do I understand correctly that you'd be happy if the API always just required at
> least one checkpoint cosignature from the witness when using cosign-subtree?

Yup, that seems reasonable! I don't feel very strongly about anything here. Mostly showed up to recount the original motivations. They may well have been wrong or unnecessary. :-)

rgdd:
> Yup, that seems reasonable! I don't feel very strongly about anything here.
> Mostly showed up to recount the original motivations. They may well have been
> wrong or unnecessary. :-)

Great, much appreaciated and thanks for the good work you're doing in the space!

tlog-witness.md
FiloSottile: I doubt anyone uses it, but done for consistency.

tlog-witness.md
FiloSottile: Probably not, but also probably fine to leave up to the implementation, based on what it'd like to get back.

FiloSottile:
I landed all the review suggestions, and dropped the DoS mechanism and the stateful verification option. The latter, we can always add back if someone wants to use it.

The only reason I worry about the DoS part is that CT was harmed by requiring O(n) signatures, which is more than some secure hardware solutions can handle. MTC needs O(1) signatures from CAs, which is great. add-checkpoint can be used to get a fresh signature over the same tree, but the response can still be cached for e.g. 1 second, capping the number of signatures at O(1) again. sign-subtree doesn't allow caching, so instead the "DoS" mechanism was relying on CAs to produce O(1) subtrees. Maybe we shouldn't have called it DoS.

Anyway, not a very strong opinion, we're probably going to solve high-speed signatures for our witnesses anyway.

Merging to make progress, if there's anything to change before tagging a release, please open an issue!

PR #249, opened 2026-05-06
tlog-mirror: allow client to truncate the request body

lukevalenta:
Relates to #219.

A single add-entries request whose body fits a fixed cap may not fit
the entire range `[upload_start, upload_end)`. This is concretely a
problem on HTTP platforms with per-request body limits (Cloudflare
Workers caps at 100 MB on standard plans, GCP Cloud Functions at 32
MB, AWS Lambda at 6 MB, etc.) when mirrors are catching up to logs
that have advanced significantly since the mirror last cosigned.

Allow the client to send a prefix of the canonical entry-package
sequence. The mirror responds with 409 carrying the advanced next
entry whenever any package was saved, so the client always learns
where to resume. No wire format additions; the existing cryptographic
contract and "first missing entry" computation already cover partial
updates.

### DIFF
--- a/tlog-mirror.md
+++ b/tlog-mirror.md
@@ -147,29 +147,31 @@ checkpoints.
 `upload_start` MUST be less or equal to `upload_end`. It MUST also be less or
 equal to the mirror's next entry for the origin.
 
-The request body uploads the log entries whose indices are in
-`[upload_start, upload_end)`. Entries are grouped into bounded-size entry
-packages. Each package has a [subtree consistency proof][] that allows the
-mirror to verify and commit the entries in the package without buffering the
+The request body uploads log entries from the interval
+`[upload_start, upload_end)`. The interval determines a fixed *canonical
+sequence* of entry packages, aligned at multiples of 256. Each package
+carries a bounded number of entries with a [subtree consistency proof][],
+allowing the mirror to verify and commit the package without buffering the
 entire request body.
 
-Each entry package is determined by a half-open interval `[start, end)` of log
-indices. The request MUST contain entry packages whose intervals' disjoint
-union, in order, is the interval `[upload_start, upload_end)`. The overall
-interval MUST be divided into packages at multiples of 256, to align with the corresponding entry bundles.
+The request MUST contain a prefix of the canonical sequence: zero or more
+packages, each exactly as specified below, in order. If the client sends a
+strict prefix, the mirror processes the prefix and advances its next entry
+accordingly; the client issues further `add-entries` requests to complete
+the upload.
 
-More precisely, if `upload_start` is equal to `upload_end`, there are no
-packages. Otherwise, let `rounded_start` be `upload_start` rounded down to a
-multiple of 256, and let `rounded_end` be `upload_end` rounded up to a
-multiple of 256. The request MUST contain
+If `upload_start` is equal to `upload_end`, the canonical sequence is empty.
+Otherwise, let `rounded_start` be `upload_start` rounded down to a multiple
+of 256, and let `rounded_end` be `upload_end` rounded up to a multiple of
+256. The canonical sequence consists of
 `num_packages = (rounded_end - rounded_start) / 256` packages. Entry
-package `i`, for `0 <= i < num_packages`, MUST be computed from the interval
-`[start, end)` where:
+package `i`, for `0 <= i < num_packages`, is determined by the half-open
+interval `[start, end)` of log indices where:
 
     start = max(upload_start, rounded_start + i * 256)
     end = min(upload_end, rounded_start + (i + 1) * 256)
 
-The package MUST contain the following values, concatenated.
+Each package MUST contain the following values, concatenated.
 
 * The log entries in `[start, end)`, each with a big-endian uint16 length prefix
 * 1 byte, encoding an 8-bit unsigned integer, `num_hashes`, which MUST be at
@@ -241,6 +243,18 @@ If this verification process fails, it MUST respond with a
 "422 Unprocessable Entity" HTTP status code and end processing. Otherwise, it
 saves the entries as pending entries. If some entry has already been written to
 the log or the pending entry list, the mirror MUST skip saving that entry.
+
+The mirror discards any partial bytes after the last successfully
+authenticated entry package. If at least one entry package was authenticated
+and saved, the mirror MUST respond with a "409 Conflict" carrying the
+advanced next entry, as described above. The client retries with
+`upload_start` set to the advertised next entry value, repeating until the
+mirror has received all packages covering `[upload_start, upload_end)`.
+
+If no entry package was authenticated and saved before the body ended (for
+example, the request header itself was malformed, or the first package's
+bytes were truncated mid-package), the mirror MUST respond with a
+"400 Bad Request" HTTP status code.
 
 Once all expected entry packages are successfully validated and committed, the
 next entry will be greater or equal to `upload_end`. The mirror then finishes
@@ -319,6 +333,13 @@ or update its mirror checkpoint until all entries are durably committed. If the
 mirror commits entries out of order, it MUST correctly compute the next entry to
 be the *first* missing entry, even if some subsequent entries have been
 committed. Mirror clients will then reupload the subsequent entries.
+
+To work around per-request body size limits on common hosting platforms,
+clients SHOULD limit each `add-entries` request to at most 32 entry
+packages (8192 entries). When `(rounded_end - rounded_start) / 256`
+exceeds 32, the client sends the first 32 packages of the canonical
+sequence as a prefix, receives a 409 with the advanced next entry, and
+issues further requests to complete the upload.
 
 A mirror MAY additionally implement other update processes, provided it continues
 to correctly operate `add-entries` and never violates its cosigner requirements


tlog-mirror.md
```diff
@@ -154,9 +154,22 @@ mirror to verify and commit the entries in the package without buffering the
...
-interval MUST be divided into packages at multiples of 256, to align with the corresponding entry bundles.
+indices. The request MAY contain entry packages whose intervals' disjoint
+union, in order, is a prefix of the interval `[upload_start, upload_end)`.
+That is, the client MAY truncate the request body after some whole number of
+entry packages, sending fewer packages than would cover the full interval. The
+client MUST NOT truncate mid-package; each package the client sends MUST be
+well-formed. The overall interval MUST be divided into packages at multiples
+of 256, to align with the corresponding entry bundles.
+
+Truncation is intended for clients uploading to mirrors deployed on platforms
```
davidben: What happens on these platforms if you cross this limit? Since the limit exists on the server, but it's the client choosing how much to upload, I'm wondering what ends up happening if the client is unaware. Does the request fail completely before the server is able to ingest stuff, or do we end up consuming the bytes and the server just stops at some point?

tlog-mirror.md
```diff
@@ -154,9 +154,22 @@ mirror to verify and commit the entries in the package without buffering the
 entire request body.
 
 Each entry package is determined by a half-open interval `[start, end)` of log
-indices. The request MUST contain entry packages whose intervals' disjoint
-union, in order, is the interval `[upload_start, upload_end)`. The overall
-interval MUST be divided into packages at multiples of 256, to align with the corresponding entry bundles.
+indices. The request MAY contain entry packages whose intervals' disjoint
```
davidben: Think this should still be a MUST, now that it says it's a prefix? Otherwise this sentence says "oh yeah the body MAY be this... but I guess it can also be a picture of an giraffe since we didn't really say anything". :-)

tlog-mirror.md
```diff
@@ -154,9 +154,22 @@ mirror to verify and commit the entries in the package without buffering the
...
+union, in order, is a prefix of the interval `[upload_start, upload_end)`.
+That is, the client MAY truncate the request body after some whole number of
+entry packages, sending fewer packages than would cover the full interval. The
+client MUST NOT truncate mid-package; each package the client sends MUST be
+well-formed. The overall interval MUST be divided into packages at multiples
+of 256, to align with the corresponding entry bundles.
```
davidben: The flow of this paragraph is a little odd to me:

What you're trying to say is that you divide `[upload_start, upload_end)` into packages and then you're allowed to cut it off at a package boundary, right? But that you can't do something like take take [256, 1024) and then choose as your prefix `[256, 600)` and then get packages `[256, 512)` and `[512, 600)` because the receiver is not going to expect a partial package there.

Is that right? ISTM this paragraph is saying:

* Entry packages are intervals.
  * Got it. Cool.
* The request MAY (MUST?) contain packages that together give _some prefix_ of the advertised interval.
  * What prefix? I guess any prefix? It didn't say.
* _That is_, the client MAY truncate after a whole number of packages.
  * Hmm. The sentence we're clarifying with "That is" didn't actually say anything about whole number of packages. For that matter, we don't really know what are the constraints on packages yet.
* The client MUST NOT truncate mid-package
  * Okay, but can I use packages [256, 257), [257, 258), etc?
* Packages must be divided at multiples of 256
  * Oh! Okay, that's how packages work. But, hmm, this still allows  `[256, 512)` and `[512, 600)` in the example above, right? Those cover `[256, 600)`, which is indeed a prefix of `[256, 1024)`. No one said that my prefix had to end at a multiple of 256 too.

I think this is partly the fault of the original paragraph being barely understandable for what it was trying to convey. Now that we added yet another concept, it's getting too big.

Don't have a concrete suggestion on how to phrase this though. Maybe something where we start by talking about how to divide the target interval into entry packages, and then separately say "actually you can just cut it off if you want", but you have to cut it at a whole entry package.

tlog-mirror.md
davidben: Nit: It's not actually uploading those log entries anymore now that it's truncated, right?

tlog-mirror.md
```diff
@@ -242,6 +255,18 @@ If this verification process fails, it MUST respond with a
...
+(see "Request Body" above), the mirror MUST treat the received packages as an
+authenticated update, save their entries as pending entries, and respond with
+a "409 Conflict" carrying the advanced next entry, as described above. The
+client retries with `upload_start` set to the advertised next entry value to
+make further progress, repeating until the mirror has received all packages
+covering `[upload_start, upload_end)`.
+
+If the request body is malformed (for example, truncated mid-package), the
+mirror MUST respond with a "400 Bad Request" HTTP status code. The mirror MAY
+still persist any whole packages received before the malformed point.
```
davidben: If the client were unaware of the server's limit and the stream just got cut off partway through (is that what happens?), it may not necessarily be truncated mid package. Maybe 409 would still be better?

Or should the size limit be something the client is more explicitly aware of?

Or maybe we should just upload each entry package in a separate request and not do this big streaming thing, if it's a huge pain?

davidben (commented):
Ah fun. Generally on board with the change, though some comments on the specific phrasing.

tlog-mirror.md
lukevalenta: In the Workers case (the only one I tested), if the client uploads more than 100MB the platform terminates the request immediately and returns a [413 Request entity too large](https://developers.cloudflare.com/workers/platform/limits/#request-and-response-limits). I'm using a Durable Object (single-threaded stateful worker) as the synchronization primitive, and when the 413 gets triggered the entire request context gets rolled back (pending Durable Object Storage DB writes) because of the failed request. I could probably persist some of the data processed by the Durable Object if I used a different database primitive.

davidben: Hmm. I guess that means the client needs to be aware of the limit. That's annoying. :-/

tlog-mirror.md
lukevalenta: Good call. In the updated version I changed it to 409 if at least one entry package was successfully processed, and 400 otherwise (since that is more likely due to a malformed header, etc.). wdyt?

tlog-mirror.md
lukevalenta: Did a cleanup pass and tried to address this by defining a 'canonical sequence' of entry packages determined by `[upload_start, upload_end)` and then saying the request body must contain a prefix of that canonical sequence.

tlog-mirror.md
lukevalenta: Obsoleted latest pass at the text.

tlog-mirror.md
lukevalenta: Yeah.. I suppose doing some sort of binary search or backoff to find the limit wouldn't be too hard though. Mirrors can also make their limits available at some metadata endpoint. It's a little annoying but I don't think in itself a reason to reject the streaming design if it seems to work well otherwise.

tlog-mirror.md
```diff
@@ -242,6 +248,18 @@ If this verification process fails, it MUST respond with a
 saves the entries as pending entries. If some entry has already been written to
 the log or the pending entry list, the mirror MUST skip saving that entry.
 
+The mirror discards any partial bytes after the last successfully
+authenticated entry package. If at least one entry package was authenticated
```
phbnf, line 248: Am I missing something, or does it mean that he mirror returns a 4XX when everything is working as intended?

lukevalenta: You're understanding correctly. It is a little funky, because a deliberate prefix isn't a "client error".

I think there are three main cases to consider here:
1) The client tries to do an upload with upload_start too far away from the mirror's next entry. The mirror returns a '409 Conflict' with an up-to-date next entry, and the client uses that to reconstruct its next request.
2) The client does a deliberate prefix upload and successfully updates the mirror. If the mirror sends a 200 response with an empty body in this case, the client should still be able to infer (based on how much data it uploaded) what the mirror's next entry is. The client can then use that for the next add-entries request. (We could also return the next entry explicitly in the response.)
3) There is some connection issue and the client's request gets cut off partway through. The mirror might still commit the authenticated entry packages that it has received. Not sure it matters what response it sends to the client, since if the connection is broken the client won't receive it. The mirror also might not have a choice in what's returned if the platform returns the response.

Since the client does the same thing (calculate the next entry and move to the next add-entries request) in all of these cases, it might simplify processing if we just return a 409, but I'm also OK to make it a 200 for case (2) if folks prefer the cleaner semantics. Thoughts?

phbnf: Thanks for the details! I'm mostly concerned about a client that would always ignore  a log's max request size. If they did, that could lead to a steady traffic of 4XX, which wouldn't make alerting straight forward.
Maybe the 409 answer should include the mirror's max request size? That would incentivize clients to align their next request accordingly.

(Another scenario that is similar to 3 would be a mirror restart. In which case a 409 would 100% do, and I like this PR because it makes it explicit that the mirror can respond with such a 409 now.)

lukevalenta: In the Cloudflare Workers case (which is what I encountered and led to this PR), the platform returns the `413 Payload Too Large` directly to the client when the streamed payload hits 100MB, so the mirror software itself doesn't get a say. I tried having the mirror set a much lower limit (like 10MB) and return a 429 to the client before it hits the platform limit, but that still didn't work to bypass the platform 413 response.

phbnf: Fair enough, and I guess this will be the same for any mirror behind a load balancer. On the plus side, that should only happen when a mirror catches up, and shouldn't really be an issue when a mirror is following a log as it grows.

davidben (approved):

lukevalenta:
I'll go ahead and merge, but we can tweak later if needed.

tlog-mirror.md
AlCutter: @lukevalenta sorry for the late response, just picking up on the semantics bit in here for (2), it does feel a bit cheeky to +/- "tell the client off" for doing something we explicitly permit in the spec, I think a `2xx` would be a better here (as you mention above).

I wonder if `202 Accepted` (also returning the next expected range) might be a good fit, saving the `200` for the final chunk which "completes" the range of entries under a checkpoint and garners a cosig?

`201 Created` may also work, although I have a nagging demon on my shoulder saying there may be implementations which might prefer not to return the `Location:` URL immediately (e.g. because they have internal/async processing to do before making these resources actually available: maybe replication or something?)

lukevalenta: Works for me! I put up a PR at https://github.com/C2SP/C2SP/pull/253 and I think 202 works well here.

PR #253, opened 2026-05-21
tlog-mirror: distinguish deliberate prefix upload with 202 Accepted

lukevalenta:
Previously, after a successful prefix upload (whether deliberate or due to mid-stream truncation), the mirror responded with 409 Conflict. This makes it difficult for clients to differentiate state mismatches vs successful prefix uploads, which is awkward semantically.

Introduce 202 Accepted for the case where the mirror authenticated and committed at least one entry package but has not yet received all packages covering `[upload_start, upload_end)`. The response body format is identical to the existing 409 body, and client retry behavior is identical, so client logic only adds one branch.

409 Conflict is now reserved for state mismatches: unrecognized `upload_end`, stale `upload_start`, or mirror checkpoint having advanced past `upload_end`.

Addresses the discussion at https://github.com/C2SP/C2SP/pull/249/changes#r3265632279.

### DIFF
--- a/tlog-mirror.md
+++ b/tlog-mirror.md
@@ -214,9 +214,9 @@ The mirror SHOULD send these error responses without waiting for the entire
 request body to be available. Conversely, the client SHOULD be prepared to
 receive an error response before the request body is fully sent.
 
-When sending a 409 response, the response body MUST have a `Content-Type` of
-`text/x.tlog.mirror-info` and consist of three lines, each followed by a
-newline (U+000A):
+When sending a 409 Conflict or 202 Accepted response, the response body MUST
+have a `Content-Type` of `text/x.tlog.mirror-info` and consist of three lines,
+each followed by a newline (U+000A):
 
 * The tree size of a valid pending checkpoint, in decimal
 * The next entry, in decimal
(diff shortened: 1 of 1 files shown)


tlog-mirror.md
```diff
@@ -214,9 +214,9 @@ The mirror SHOULD send these error responses without waiting for the entire
 request body to be available. Conversely, the client SHOULD be prepared to
 receive an error response before the request body is fully sent.
 
-When sending a 409 response, the response body MUST have a `Content-Type` of
-`text/x.tlog.mirror-info` and consist of three lines, each followed by a
-newline (U+000A):
+When sending a 409 Conflict or 202 Accepted response, the response body MUST
```
AlCutter: This is not wrong, but the flow is a little bumpy as fact that a `409` could be returned was introduced in the list above, however no mention of `202` has been made yet.

Maybe there's a way to restructure to smooth it out? Perhaps an introductory para before this one, or moving this para about forming a proper `409` or `202` response and the bits about how a client reacts to it further down?

Alternatively, maybe it's fine to leave it like this for now, maybe leave a TODO to, er, do a restructure/tidy up before `v1.0` in case there are more tweaks coming which may also affect flow.

I'm happy either way.

AlCutter (approved):

tlog-mirror.md
lukevalenta: Good catch! I shuffled the paragraphs around to describe when the status codes should be returned first, and then what should be in the response body.

davidben (approved):

PR #263, opened 2026-06-12
mtc-tlog: Add an X.509 extension for the CA base URL

davidben:
OID arc is the same arc that all the other MTC experiment OIDs are being allocated from. Just because it would be more procedurally annoying to have this document grab an OID from one of the IETF PKIX registries, and because OIDs are generally functionally equivalent, I'm proposing that this OID just be _the_ OID for this document.

### DIFF
--- a/mtc-tlog.md
+++ b/mtc-tlog.md
@@ -23,7 +23,24 @@ An MTC CA following this profile has, in addition to [CA parameters][] defined
 in the MTC specification, a *CA prefix URL*. The CA prefix URL determines the
 serving URL for each issuance URL, as described below.
 
+When such a CA is [represented as an X.509 certificate][], the certificate has a
+non-critical X.509 extension with OID 1.3.6.1.4.1.44363.47.3 and syntax an
+IA5String, as defined below. The IA5String's contents are the CA prefix URL.
+Presence of this extension indicates that the certificate subject follows this
+specification.
+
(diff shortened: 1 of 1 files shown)


FiloSottile:
C2SP has [its own PEN arc](https://github.com/C2SP/C2SP/blob/main/.github/OIDs.md). Would you like 1.3.6.1.4.1.64829.2.1 for this?

davidben:
Oh awesome. Sure, that looks more official. :-)

mtc-tlog.md
```diff
@@ -23,7 +23,24 @@ An MTC CA following this profile has, in addition to [CA parameters][] defined
 in the MTC specification, a *CA prefix URL*. The CA prefix URL determines the
 serving URL for each issuance URL, as described below.
 
+When such a CA is [represented as an X.509 certificate][], the certificate has a
+non-critical X.509 extension with OID 1.3.6.1.4.1.44363.47.3 and syntax an
```
jdeblasio: Need it be non-critical?

jdeblasio (approved):

mtc-tlog.md
davidben: Just seemed unnecessary to give a choice and I don't see a reason to make it critical. If you know MTC but don't know tlog, you presumably don't care if it implements this profile at all.

PR #266, opened 2026-06-12
mtc-tlog: update pruning text so RPs control when pruning happens

AGWA:
For context, my concerns about the current text are based on the following experiences from the WebPKI:

1. On many occasions, CAs have misissued certificates with lifetimes that were longer than permitted. This has never led to a CA being distrusted.

2. Several CT logs have incorporated certificates which expired after the log's intended expiry range.

3. Multiple RPs/UAs regularly take weeks to months to update their CT log lists.

### Concern 1: CAs cannot prune unexpired certificates with erroneously long lifetimes

This could lead to a single unexpired certificate followed by a great many expired certificates, causing the log to grow much larger than expected, increasing costs to multiple parties in the ecosystem.

### Concern 2: Slow log list updates cause alert fatigue for monitors

Consider the following sequence of events:

1. CA prunes their log in accordance with RP policy.
2. Weeks to months pass.
3. RP updates the minimum index in their log list to reflect the log's new minimum index.

During those weeks/months, monitors are unable to retrieve entries from the log despite those entries being trusted by the RP. A decent monitor would complain to its operator, leading to alert fatigue.

### The fix

We _really_ want the order of operations to be:

1. RP updates the minimum index in their log list per their desired policy.
2. CA notices that the minimum index for their log has increased, and prunes their log if desired.

This is analogous to Chrome's existing requirement that CT logs remain operational until they are removed from the log list, even if the log is past its expiry range and contains only expired certificates.

Both concerns are addressed:

1. If an RP decides they don't care about breaking an unexpired certificate with an overly-long lifetime, they can bump the minimum index beyond the certificate, allowing the log to be pruned. This is analogous to Chrome's existing practice of removing CT logs after their expiry range ends, even if the log still contains unexpired certificates.

2. If an RP is slow to update its log list, CAs are not allowed to prune entries, so monitors won't experience errors trying to download them.

### DIFF
--- a/mtc-tlog.md
+++ b/mtc-tlog.md
@@ -69,9 +69,8 @@ a [transparency log cosigner][] as described below. Issuance logs MAY serve
 additional cosignatures, including ones from cosigners that are not
 [MTC cosigners][MTC cosigner].
 
-If [pruned][], an issuance log MUST set its minimum index such that only expired
-entries are pruned. Relying parties MAY set further restrictions, such as
-requiring that an entry be expired for at least 6 months before pruning.
+If [pruned][], an issuance log MUST NOT set its minimum index larger than the
+smallest minimum index trusted by relying parties.
 
 An issuance log with a landmark sequence MUST [publish active landmarks][] at
 the following URL:
--- a/mtc-tlog.md
+++ b/mtc-tlog.md
@@ -88,9 +88,9 @@ a [transparency log cosigner][] as described below. Issuance logs MAY serve
 additional cosignatures, including ones from cosigners that are not
 [MTC cosigners][MTC cosigner].
 
-If [pruned][], an issuance log MUST set its minimum index such that only expired
-entries are pruned. Relying parties MAY set further restrictions, such as
-requiring that an entry be expired for at least 6 months before pruning.
+Relying parties SHOULD set restrictions on [pruning][], such as requiring
+that the log's minimum index be at most the minimum trusted index in
+up-to-date copies of the relying party's trust anchors
 
 An issuance log with a landmark sequence MUST [publish active landmarks][] at
 the following URL:
@@ -104,7 +104,7 @@ the following URL:
 [MTC cosigner]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-04.html#name-cosigners
 [note signature]: http://c2sp.org/signed-note
 [prefix URL]: https://c2sp.org/tlog-tiles#parameters
-[pruned]: https://c2sp.org/tlog-tiles#pruning
+[pruning]: https://c2sp.org/tlog-tiles#pruning
 [publish active landmarks]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-04.html#name-publishing-landmarks
 [transparency log cosigner]: https://c2sp.org/tlog-cosignature
 


davidben:
Ah sorry, missed this. (Never got an email notification for this. Just saw it in the PR list on accident. Not sure how tooling here works... is there supposed to be a way to auto-CC maintainers for PRs on drafts?)

I agree with this change in spirit, but I'm not sure how to word it. If we say "smallest minimum index trusted by relying parties", what does that mean for, say, an install of Chrome in a closed laptop that hasn't gotten updates for the last year? Or a random Python venv I made last year with a random snapshopt of certifi? Is a CA obligated to query the state of every install of every RP that might exist? Conversely, if the Chrome Root Store updated its minimum index just now, can the CAs prune now, or should they be asked to wait a couple days first?

My workaround for this mess previously to say that the CA ultimately decides when to prune (I mean, that's unavoidably true because the CA is in control of that field of its config file), but they are subject to whatever policies imposed by root programs. And then the fuzziness between "random old installs of this software" and "root program's current view of the world" is captured by the root program imposing some appropriate policy.

But yeah the text doesn't do a good job of explaining this. WDYT about something like this?

> If [pruned][], an issuance log MUST set its minimum index such that only expired entries are pruned. Relying parties SHOULD set further restrictions, such as:
> * requiring that an entry be expired for at least 6 months before pruning
> * requiring that the log's minimum index be at most than the minimum trusted index in up-to-date copies of the relying party's trust anchors

AGWA:
Good point about out-of-date clients; CAs definitely shouldn't be required to consider them. I think "minimum trusted index in up-to-date copies of the relying party's trust anchors" is good wording and addresses my second concern.

However, my first concern is still not addressed. CAs need to be able to prune expired certificates, because otherwise a single misissued certificate could inhibit pruning for a long time. A blanket prohibition against pruning expired certificates, either in the spec or in root program policy, is too inflexible.

How about this:

> Relying parties SHOULD set restrictions on pruning, such as requiring that the log's minimum index be at most the minimum trusted index in up-to-date copies of the relying party's trust anchors

In general, relying parties would not bump up the minimum trusted index until all of the certificates below that index are expired for some period of time, but could still do so in exceptional circumstances.

davidben:
> However, my first concern is still not addressed. CAs need to be able to prune expired certificates, because otherwise a single misissued certificate could inhibit pruning for a long time. A blanket prohibition against pruning expired certificates, either in the spec or in root program policy, is too inflexible.

Ah good point. Don't know how I missed that with how clearly you wrote it. 🤦 

> How about this:
>
>> Relying parties SHOULD set restrictions on pruning, such as requiring that the log's minimum index be at most the minimum trusted index in up-to-date copies of the relying party's trust anchors

SGTM!

Does lose any discussion of expiry, but that's probably fine. I guess if I wanted to fully tease apart the dependency order, it's:

* Certificate lifetimes are capped at X
* Thus all violations of that are misissuances
* Thus RPs should revoke them
* Once everything is expired _or revoked_, RPs should move their minimum index forward
* RPs should demand that the CA's minimum index lag behind the RP's minimum index by some appropriate amount

But how much of this is worth spelling out here, I dunno. I'm happy with your suggested text.

AGWA:
Agree with your bullet list, though I'm also not sure how much of it is worth spelling out here :sweat_smile:. I've updated the PR with my latest proposed text.

davidben (approved):

PR #267, opened 2026-06-14
chunked-encryption: add optional message padding scheme

FiloSottile:


FiloSottile:
A note to future me: here's an efficient implementation of Padmé.

```
// paddingLength returns the number of padding bytes to add to a message of
// length msgLen, implementing Padmé as described in Algorithm 1 of
// https://arxiv.org/abs/1806.03160v4.
func paddingLength(msgLen uint64) uint64 {
	if msgLen == 0 {
		return 0
	}
	E := bits.Len64(msgLen) - 1
	S := bits.Len(uint(E))
	z := E - S
	m := (uint64(1) << z) - 1
	L := (msgLen + m) & ^m
	return L - msgLen
}
```

PR #269, opened 2026-06-16
mtc-tlog: describe client-constructed landmark-relative certificates

lukevalenta:
Document that clients can construct landmark-relative certificates for themselves by fetching tiles and landmarks directly from the CA issuance log. Tracked by https://github.com/ietf-plants-wg/merkle-tree-certs/issues/246.

Separately, I filed https://github.com/ietf-plants-wg/merkle-tree-certs/issues/253 to add some more detail on the exact algorithm for determining a landmark subtree/landmark number for a given entry index and landmark sequence.

### DIFF
--- a/mtc-tlog.md
+++ b/mtc-tlog.md
@@ -99,6 +99,16 @@ the following URL:
 <CA prefix URL>/<log number>/landmarks
 ```
 
+This endpoint is mutable, but updates infrequently. Responses SHOULD
+include a [`Cache-Control: max-age=<seconds>`][cache-control] header whose
+value approximates the number of seconds the CA expects until it next
+allocates a landmark. CAs SHOULD include a small buffer in this value to
+ensure the next landmark is allocated before the cache expires, avoiding
+unnecessary retries from clients waiting on it. A client waiting for a
+future landmark can then re-fetch the resource after its cached copy
+expires to pick up the latest landmark sequence.
+
+[cache-control]: https://www.rfc-editor.org/rfc/rfc9111#section-5.2.2.1
 [CA cosigner]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-04.html#name-certification-authority-cos
 [log ID]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-04.html#name-issuance-logs
 [MTC cosigner]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-04.html#name-cosigners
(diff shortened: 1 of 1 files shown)


mtc-tlog.md
```diff
@@ -142,3 +142,111 @@ configured to accept the next few unused log numbers.
 [MTC-compatible]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-04.html#name-signature-format
 [sign-subtree]: https://github.com/C2SP/C2SP/pull/245
 [standalone certificates]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-04.html#name-standalone-certificates
+
+## Client-Constructed Landmark-Relative Certificates
+
+Because each issuance log is served as a [tiled transparency log][], an
```
davidben:
```suggestion
Because each issuance log is served as a [tiled transparency log][tiled transparency logs], an
```

This linking convention is... a little error-prone. :-)

mtc-tlog.md
```diff
@@ -142,3 +142,111 @@ configured to accept the next few unused log numbers.
...
+[authenticating party][] holding a [standalone certificate][standalone certificates]
+MAY construct a [landmark-relative certificate][] for the same entry by
```
davidben:
```suggestion
can construct a [landmark-relative certificate][] for the same entry by
```

Not sure if this is right or not. MAY is a little odd because you're really just observing a fact. (Really not just the authenticating party but anyone at all can do it.)

mtc-tlog.md
```diff
@@ -142,3 +142,111 @@ configured to accept the next few unused log numbers.
...
+   `<CA prefix URL>/<log number>/landmarks`.
+
+2. Derive the landmark subtree and landmark number covering the entry index
+   from the landmark sequence (see
+   [Constructing Landmark-Relative Certificates][constructing landmark-relative]).
+
+   If no active landmark covers the entry, the authenticating party cannot
+   construct a landmark-relative certificate for it. This can happen because
+   the entry is not yet covered by an allocated landmark, in which case the
+   authenticating party SHOULD retry later, or because every landmark that once
```
davidben: How does the authenticating party learn when "later" is?

Should we recommend that this resource contain a `Retry-After` header of when the CA expects to next issue a landmark?

mtc-tlog.md
```diff
@@ -142,3 +142,111 @@ configured to accept the next few unused log numbers.
...
+   covered the entry has aged out of the active landmark window, in which case
+   retrying will not help.
+
+3. Fetch the [Merkle Tree tiles][merkle tree tiles] containing the hashes
+   needed to generate the [subtree inclusion proof][] for the entry index
+   in the landmark subtree, and generate the proof. A party that is aware
+   of mirrors of the issuance log MAY fetch tiles from those mirrors
+   instead of, or in addition to, the CA prefix URL; this document does
+   not define a discovery mechanism for such mirrors, and mirror URLs are
+   obtained out of band.
```
davidben: Nit: Think we don't need the second clause. (I dunno, maybe the document that defines a discovery mechanism will do it in-band! Who knows!)

```suggestion
   not define a discovery mechanism for such mirrors.
```

mtc-tlog.md
```diff
@@ -142,3 +142,111 @@ configured to accept the next few unused log numbers.
...
+   `signatureValue`. The new `signatureValue` contains an
+   [`MTCProof`][certificate format] whose `extensions` field is copied from
+   the standalone certificate's `MTCProof`, whose `start`, `end`, and
+   `inclusion_proof` describe the landmark subtree and the inclusion proof
+   from above, and whose `signatures` field is empty.
+
+5. Compute the trust anchor ID for the resulting
+   [landmark-relative certificate][] from the CA ID, log number, and
+   landmark number (see
+   [Landmark-Relative Certificates in TLS][landmark-relative tls]).
```
davidben:
```suggestion
5. Configure certificate selection for the resulting
   [landmark-relative certificate][] based on the CA ID, log number, and
   landmark number (see
   [Landmark-Relative Certificates in TLS][landmark-relative tls]).
```

(There's a couple of IDs, and this doesn't say what you do with it. I figure we can just cite the section and say "go figure it out from this".)

mtc-tlog.md
lukevalenta: I had some text in there about any party being able to construct it, but then dropped it since it felt a little more concrete talking about the authenticating party, but probably worth at least a parenthetical.

mtc-tlog.md
lukevalenta: Oh yeah I was thinking that ideally the landmark sequence would be published to the S3 bucket with a cache TTL set to whenever the CA expects to publish the next landmark, so that could work as a signaling mechanism. I guess we could add the bit about cache TTL up above.

lukevalenta: Added some text about using cache control to signal when the next landmark will be published.

mtc-tlog.md
```diff
@@ -99,6 +99,16 @@ the following URL:
...
 ```
 
+This endpoint is mutable, but updates infrequently. Responses SHOULD
+include a [`Cache-Control: max-age=<seconds>`][cache-control] header whose
+value approximates the number of seconds the CA expects until it next
+allocates a landmark. CAs SHOULD include a small buffer in this value to
+ensure the next landmark is allocated before the cache expires, avoiding
+unnecessary retries from clients waiting on it. A client waiting for a
+future landmark can then re-fetch the resource after its cached copy
+expires to pick up the latest landmark sequence.
```
davidben: Would a `Retry-After` make more sense? Then the cache doesn't need to over-commit, while the `Retry-After` can include the buffer. Or would that confuse people with an HTTP 200?

mtc-tlog.md
```diff
@@ -142,3 +152,112 @@ configured to accept the next few unused log numbers.
...
+   `signatureValue`. The new `signatureValue` contains an
+   [`MTCProof`][certificate format] whose `extensions` field is copied from
+   the standalone certificate's `MTCProof`, whose `start`, `end`, and
+   `inclusion_proof` describe the landmark subtree and the inclusion proof
+   from above, and whose `signatures` field is empty.
+
+5. Configure certificate selection for the resulting
+   [landmark-relative certificate][] based on the CA ID, log number, and
+   landmark number (see
+   [Landmark-Relative Certificates in TLS][landmark-relative tls]).
```
davidben: Maybe worth mentioning that, if your PKI uses timestamped groups, you may also need some information about those.
https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#name-timestamped-landmark-groups

mtc-tlog.md
```diff
@@ -142,3 +152,112 @@ configured to accept the next few unused log numbers.
...
+## ACME Issuance for MTC-Aware Clients
+
+A CA following this profile and issuing certificates over [ACME][RFC 8555]
+MUST convey its CA prefix URL to MTC-aware ACME clients at certificate
+issuance time, so that the client can locate the issuance log. One use is
+described in
+[Client-Constructed Landmark-Relative Certificates](#client-constructed-landmark-relative-certificates).
+
+An ACME client signals MTC-awareness by including an MTC media type in the
+`Accept` header of its certificate-download request, as defined by the MTC
```
davidben: Hmm, there isn't actually an MTC media type. (Although does the client need to signal? You can just send new link relations.)

mtc-tlog.md
```diff
@@ -142,3 +152,112 @@ configured to accept the next few unused log numbers.
...
+[ACME extension][acme extensions]. When such a request results in the CA
+returning a [standalone certificate][standalone certificates], the response
```
davidben: Which response? (I'd expect the individual certificate URL or alternate URL that served the certificate.)

mtc-tlog.md
```diff
@@ -142,3 +152,112 @@ configured to accept the next few unused log numbers.
...
+[authenticating party]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-04.html#name-terminology-and-roles
+[CA ID]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-04.html#name-certification-authority-ide
+[certificate format]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-04.html#name-certificate-format
+[constructing landmark-relative]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-04.html#name-constructing-landmark-relat
+[landmark-relative certificate]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-04.html#name-landmark-relative-certifica
+[landmark-relative tls]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-04.html#section-8.2
+[merkle tree tiles]: https://c2sp.org/tlog-tiles#merkle-tree
+[subtree inclusion proof]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-04.html#name-subtree-inclusion-proofs
+
+## ACME Issuance for MTC-Aware Clients
```
davidben: Hmm. I do worry that this might leave readers confused about which mechanism to implement. (I suspect almost all folks will want to just have the ACME server do it. Not sure how best to phrase it.)

mtc-tlog.md
```diff
@@ -142,3 +152,112 @@ configured to accept the next few unused log numbers.
...
+  [Certificate Format][certificate format]).
+* The CA prefix URL of the issuance log. The CA conveys this URL to
+  [MTC-aware ACME clients](#acme-issuance-for-mtc-aware-clients) at
+  certificate issuance time, as described below. It may also be obtained
+  from the CA's issuance-log X.509 extension (see
+  [Parameters](#parameters)), or out of band. The CA prefix URL identifies
+  the log's serving location as defined in
+  [Serving Issuance Logs](#serving-issuance-logs).
+
+Construction proceeds as follows:
```
davidben: Would it make sense to flesh out https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#name-constructing-landmark-relat and cite that? Maybe with some text to say "oh yeah you can just fetch from the log to get the landmarks, inclusion proof, etc. Feel free to use a mirror if you know about it."

jdeblasio: Oh, I like the idea of, as much as possible, pushing this text into a generic "how to construct landmark-relative certs". Since the approach this section is describing won't be the path for most consumers, I'm a little wary of spending too many words outlining out to do it lest it look like a recommended path. I do think it's reasonable to mention how to do this operation though, so offloading as much as possible to a shared algorithm keeps the knowledge, but reduces the focus on a weird scenario.

jdeblasio (commented):
Strong endorsement of most of David's questions.

mtc-tlog.md
lukevalenta: One argument for `Cache-Control` is that it seems like major cloud providers support adding it as a stored header on objects, so you don't need to add a proxy in front of the object storage. (I'm definitely not tied to `Cache-Control`, but that's why I originally picked it.)

* [Cloudflare](https://developers.cloudflare.com/r2/api/s3/extensions/#extended-metadata-using-unicode)
* [AWS S3](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#API_PutObject_RequestSyntax)
* [GCS](https://docs.cloud.google.com/storage/docs/metadata#editable)
* [Azure](https://learn.microsoft.com/en-us/rest/api/storageservices/set-blob-properties?tabs=microsoft-entra-id#request-headers-all-blob-types)

mtc-tlog.md
lukevalenta: Sounds good! I can update the other PR to add some more detail and subsequently cut the detail out here.

mtc-tlog.md
lukevalenta: Left this as `Cache-Control`. I'll go ahead and resolve this, but we can re-open if this needs more discussion!

mtc-tlog.md
lukevalenta: Clarified with "Most deployments obtain the landmark-relative certificate directly from the CA"

mtc-tlog.md
lukevalenta: Oh oops, I was referring to the TAI extension, but yeah we can just omit and send the link relation unconditionally.

mtc-tlog.md
lukevalenta: Ah yep this would be the certificate-download response from https://datatracker.ietf.org/doc/html/rfc8555/#section-7.4.2.

mtc-tlog.md
lukevalenta: Deferring the TLS identifier construction to the MTC draft

mtc-tlog.md
lukevalenta: Details on constructing landmark-relative certs from standalone + landmark sequence + issuance log are now included in https://github.com/ietf-plants-wg/merkle-tree-certs/pull/254 instead. This PR should be ready for another round of review, now!

lukevalenta:
I cleaned up the text a bit more and also dropped the requirement that the ACME server includes the `c2sp.org/mtc-tlog/prefix-url` link relation when clients download the standalone cert since it seems that most clients will just use the standard ACME path, and the DIY landmark creation will only be used by clients that are OK with figuring out what the log's prefix URL is out of band.

mtc-tlog.md
```diff
@@ -111,6 +103,16 @@ the following URL:
 <CA prefix URL>/<log number>/landmarks
 ```
 
+This endpoint is mutable, but updates infrequently. Responses SHOULD
+include a [`Cache-Control: max-age=<seconds>`][cache-control] header whose
+value approximates the number of seconds the CA expects until it next
```
davidben, line 108: There's a nuisance here which is that `Cache-Control` serves two purposes. Here, we want "I expect the next landmark will be after this time", but `Cache-Control` skips HTTP fetches generically, so it means "I _almost certainly_ will not have allocated a landmark past this point". I.e. I expect the buffer to go in the other direction.

My guess is you'd set `max-age` to like 5-10 minutes and say, yeah, that's good enough to capture that it's mostly static, but also means that you won't be more than 5-10 minutes late in discovering a landmark.

Perhaps `Retry-After`?

davidben: Oh I guess we already had a thread. GitHub... I'll close this and move to the other one.

mtc-tlog.md
davidben: (Copying from elsewhere.)

So my worry about `Cache-Control` is that it has a different purpose originally. It is a directive to the low-level HTTP cache that you _do not have to hit the server at all_ for the next while. That's quite aggressive. If you estimate and overshoot, folks like relying parties will not pick up your landmark if they have an HTTP cache. That will happen _even if_ the relying parties are actually just operating a separate poll loop.

My guess is a CA would set `max-age` to like 5-10 minutes and say, yeah, that's good enough to get some amount of caching, but avoids any over-caching. Setting a buffer so it _exceeds_ the time to the next landmark sounds very aggressive.

`Retry-After` feels a little more robust in that it *only* impacts the retry hint. It still allows someone to fetch more frequently if they think it's useful, etc. It also avoids impacting anyone who isn't looking at `Retry-After`.

mtc-tlog.md
davidben: I left a comment in https://github.com/ietf-plants-wg/merkle-tree-certs/pull/254#discussion_r3807212577 but I think this suggestion here may have misfired. :-) The main aim was to reduce the focus on the weird scenario, but I think we ended up making it more prominent instead.

lukevalenta: Ah yep, makes sense! I pared the text way down in the latest version, but maybe we just don't need it at all if there's no benefit to 99% of implementors? (can move discussion to https://github.com/ietf-plants-wg/merkle-tree-certs/pull/254#discussion_r3874091257)

davidben (approved):

PR #273, opened 2026-06-24
tlog-witness: add monitoring API

FiloSottile:
It matches tlog-mirror's prefixing of tlog-tiles.

/cc @C2SP/tlog-witness @C2SP/tlog-mirror

### DIFF
--- a/tlog-mirror.md
+++ b/tlog-mirror.md
@@ -119,6 +119,9 @@ a separate witness service using its pending checkpoint. If so, this witness
 service MUST be a distinct cosigner from the mirror cosigner, with a distinct
 name. The mirror's signature is computed later, as described below.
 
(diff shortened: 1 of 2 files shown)


lukevalenta (approved):

davidben (approved):

AlCutter (approved):
This seems reasonable, I like that it lines up well with `tlog-mirror` and provides for offloading the lift for lower-powered witnesses.

Should we add a bit of text about how monitors should deal with the situation where there's a spread of checkpoint sizes across the witnesses for a given log?

FiloSottile:
> Should we add a bit of text about how monitors should deal with the situation where there's a spread of checkpoint sizes across the witnesses for a given log?

It feels a bit out of scope for tlog-witness, but I don’t know where else it would fit.

PR #276, opened 2026-06-25
tlog-tiles: Upgrade and soften the Ed25519 MTI

### LINKED ISSUE #275: tlog-tiles: Ed25519 requirement is too strong mid PQ transition
tlog-tiles currently says:

> If the log is public, or is interacting in any way with the public witness network, the checkpoint MUST carry at least one Ed25519 signature by the log. The checkpoint MAY carry additional signatures of other types, by the log or otherwise.

Whether or not MTC deployments use the public witness network, I expect many of them will be "public" and thus this text would apply. That's probably a bit strong. (Also not clear whether "public witness network", without a link, refers vaguely to any public witness, or specifically https://witness-network.org/.)

Perhaps:

* Drop the MUST-level requirement on "if the log is public", so it doesn't hit new deployments with new Ed25519 legacy
* Have only SHOULD-level recommendation on algorithms for logs in general, since it depends on your ecosystem. Something like:
  * You should have an ML-DSA-44 key, unless your software isn't ready for it
  * If you need to interoperate with older witnesses, you should also have an Ed25519 key
* Say the public witness network (i.e. https://witness-network.org/) specifically has a MUST-level for... something. Not sure how far along those witnesses are in having ML-DSA-44 supported. MUST Ed25519 + SHOULD ML-DSA-44?

Thoughts?

/cc @C2SP/tlog-tiles 
/cc @C2SP/mtc-tlog

davidben:
MTI algorithms have a habit of not aging terribly well. Soften all recommendations to SHOULD. Also we're in this awkward period where we prefer ML-DSA-44 but it's very new. Try to convey that ML-DSA-44 is preferred, but maybe you don't have it, or maybe your clients don't have it, in which case the older recommendation was Ed25519.

Fixes #275

### DIFF
--- a/tlog-tiles.md
+++ b/tlog-tiles.md
@@ -47,9 +47,20 @@ with `Content-Type: text/plain; charset=utf-8`.
 This endpoint is mutable, so its headers SHOULD prevent caching beyond a few
 seconds.
 
-If the log is public, or is interacting in any way with the public witness
-network, the checkpoint MUST carry at least one Ed25519 signature by the log.
-The checkpoint MAY carry additional signatures of other types, by the log or
+If the log ecosystem does not provide its own requirements, the following
+signature algorithm recommendations apply:
+
+* If ML-DSA-44 support is available, checkpoints SHOULD carry at least one
+  ML-DSA-44 signature by the log.
+
+* Otherwise, or if compatibility with older log clients is needed, checkpoints
+  SHOULD carry at least one Ed25519 signature by the log.
+
+Checkpoints MAY carry both ML-DSA-44 and Ed25519 signatures by the log. This
+provides quantum resistance for newer log clients, while retaining compatibility
+with older log clients.
+
+Checkpoints MAY carry additional signatures of other types, by the log or
 otherwise.
 
 ### Merkle Tree


davidben:
Hmm, I guess in writing this, this does lose the note about the public witness network. Should I put that back? If so, what's the PWN's state? Ed25519 still a MUST?

(My intuition is that this is better tracked at https://witness-network.org/ or something. Algorithms required on the log for the PWN is the dual of algorithms that the PWN requires its constituent witnesses to support. Seems odd to not track that in the same place.)

jsha:
Why is this requirement in tlog-tiles instead of tlog-checkpoint? It seems to me we should delete this requirement from tlog-tiles and instead put any needed requirements (like a MUST for the public witness network) under https://github.com/C2SP/C2SP/blob/main/tlog-checkpoint.md#signatures.

davidben:
...for that matter, tlog-checkpoint _already_ has this text:

> Logs SHOULD use ML-DSA-44 cosignatures to sign the checkpoint, but MAY use any note signature algorithm based on the ecosystem they operate in. Note that the ML-DSA-44 cosignature format doesn't sign the extension lines, which SHOULD be empty.

@AlCutter and @FiloSottile, thoughts?

AlCutter:
Hmm, fair point - after all, `tlog-tiles`, specifically _isn't_ a requirement for `tlog-witness`...

davidben:
In that case, should we just do https://github.com/C2SP/C2SP/pull/293 instead? I didn't realize it was already in another doc when I wrote this.

davidben:
(Closing in favor of the other PR.)

PR #277, opened 2026-07-01
tlog-witness,tlog-mirror: switch percent-encoded origin to origin hash

FiloSottile:
Turns out we picked a bad color for the bike shed.

You _would_ think that the whole point of URL percent-encoding would be to escape special characters, but the HTTP ecosystem seems to have made a mess of it, and in many cases `/foo%2Fbar/checkpoint` will be served as `/foo/bar/checkpoint`, which defeats the point of the escaping.

* Go's net/http.FileServerFS serves based on `r.URL.Path` which is unescaped.
* CloudFront's `%2F` handling is stage- and config-dependent.
* Cloudflare unescapes `%2F` if "*[Normalize URLs to origin](https://developers.cloudflare.com/rules/normalization/)*" is enabled.

This PR switches to a hex-encoded SHA-256 hash. There was no directory listing anyway, so clients had to know the origin they were interested in already. (If we want to change this, which we might, we can add an endpoint to serve origins.) It's definitely safe and edge-case free.

Alternatives include Base64url (straight or of a hash), which is case sensitive and not the Base64 we use in checkpoints, and Base32 (straight or of a hash), which is not used anywhere else and a bit less common. I could be convinced that Base32 is better if anyone has strong opinions, but hex felt like a nicely boring answer.

/cc @C2SP/tlog-witness @C2SP/tlog-mirror

### DIFF
--- a/tlog-mirror.md
+++ b/tlog-mirror.md
@@ -11,7 +11,6 @@ signatures asserting that a mirror has done so.
 [pruning]: https://c2sp.org/tlog-tiles#pruning
 [retention policy]: https://c2sp.org/tlog-tiles#retention-policies
 [witness]: https://c2sp.org/tlog-witness
-[percent-encoded]: https://www.rfc-editor.org/rfc/rfc3986.html#section-2.1
 [subtree]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-02.html#name-subtrees
 [subtree consistency proof]: https://www.ietf.org/archive/id/draft-ietf-plants-merkle-tree-certs-02.html#name-subtree-consistency-proofs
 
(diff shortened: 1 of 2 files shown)


davidben (approved):
Sadness.

lukevalenta (approved):

AlCutter:
One of the things I really liked about the _encoded-but-more-or-less-readable_ subdirs was that it makes it relatively easy for an operator to tell that they've copied and pasted the wrong path (e.g. when they're about to irredeemably destroy data from an "obsolete" mirror).

I don't really have a better suggestion than `hex(sha256())` though, obvs `tr [badchars] _` just opens the door to collisions and confusion which is not a great trade-off.

So, I guess, sadness also.

FiloSottile:
Indeed. So @AlCutter good to merge with you?

AlCutter (approved):

PR #278, opened 2026-07-02
upki-revocation: initial specification text

cpu:
Adds initial text specifying the existing `upki` binary `index.bin` format, the Mozilla clubcard-crlite v4 serialization format, and the high-level revocation checking process used by a client consuming the `upki` revocation cache directory.

cc @djc @jschanck

### DIFF
--- a/upki-revocation.md
+++ b/upki-revocation.md
@@ -1,0 +1,779 @@
+# UPKI Revocation
+
+This document specifies the on-disk data structures and check procedure used to
+perform offline TLS certificate revocation checks against a local cache of
+Mozilla [CRLite][] filters encoded in the [clubcard][] v4 format, distributed by
+[upki][]. It is aimed at implementors of clients that consume such a cache
+in order to consult it for efficient revocation decisions during certificate
+path validation.
+
+The cache consists of one or more encoded clubcard filter files together with a
+binary index that maps `(CT log ID, SCT timestamp)` pairs to the covering
+filter files in the cache. Given an end-entity certificate and its
+embedded [SCTs][RFC 6962], an implementation can determine one of four
+outcomes: **revoked**, **not revoked**, **not enrolled**, or **not covered**
+using the available revocation data.
+
+How the cache is populated (for example production, hosting, downloading,
+verification, and incremental update of CRLite clubcard data) is out of scope
+for this document.
+
+[CRLite]: https://blog.mozilla.org/security/2020/01/09/crlite-part-2-end-to-end-design/
+[clubcard]: https://github.com/mozilla/clubcard
+[upki]: https://github.com/rustls/upki
+
+## Conventions used in this document
+
+Data structures are described using notation inspired by Section 3 of
+[RFC 8446][]. This document deviates from RFC 8446 in one respect:
+length-prefixed sequences of composite types carry an **element count** rather
+than a byte length, and are written in the form
+
+```
+    uintN count;
+    T     items[count];
+```
+
+where the leading `count` field is on the wire. `opaque x[N]` retains its RFC
+8446 meaning of a fixed-length byte string of exactly `N` bytes. `opaque
+x<A..B>` retains its RFC 8446 meaning of a variable-length byte string prefixed
+by a byte-length field wide enough to encode `B`; for opaque byte strings this
+length is equivalently a count of bytes and of elements.
+
+All integer fields are unsigned and encoded in **big-endian** order.
+
+`0x` followed by two hexadecimal characters denotes a byte value in the 0–255
+range. 
+
+`||` denotes concatenation. 
+
+`SHA-256(x)` denotes the SHA-256 digest of `x` as specified in [FIPS 180-4][],
+producing a 32-byte value.
+
+The **SPKI hash** of a certificate is the SHA-256 digest of the DER encoding of
+its `SubjectPublicKeyInfo` field, as defined in [RFC 5280][].
+
+The **serial number** of a certificate is the DER-encoded contents (that is, the
+`INTEGER` value bytes, excluding tag and length) of the end-entity certificate's
+`serialNumber` field, as specified in [RFC 5280][] §4.1.2.2. Importantly, if the
+DER encoding includes a leading `0x00` byte to disambiguate the sign, that byte
+is retained when used as a revocation input.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in [BCP 14][] [RFC 2119][] [RFC
+8174][] when, and only when, they appear in all capitals, as shown here.
+
+[RFC 8446]: https://www.rfc-editor.org/rfc/rfc8446.html
+[RFC 6962]: https://www.rfc-editor.org/rfc/rfc6962.html
+[RFC 5280]: https://www.rfc-editor.org/rfc/rfc5280.html
+[FIPS 180-4]: https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf
+[BCP 14]: https://www.rfc-editor.org/info/bcp14
+[RFC 2119]: https://www.rfc-editor.org/rfc/rfc2119.html
+[RFC 8174]: https://www.rfc-editor.org/rfc/rfc8174.html
+[Schanck 2025]: https://doi.org/10.1109/SP61157.2025.00128
+
+## Introduction
+
+CRLite is a scheme for compactly representing the set of revoked
+publicly-trusted TLS certificates so that a client can answer "is this
+certificate revoked?" locally, without contacting a certificate authority or
+OCSP responder. It is space-efficient compared to independent CRL files, leaks
+no private information to the CRL/OCSP issuer, and can support the scale of the
+Web PKI. In its current form the compact representation is a set membership data
+structure, a **clubcard filter**, combined with metadata describing the subset
+of the Web PKI it covers.
+
+A clubcard filter is a two-stage filter based on solving systems of linear
+equations over `GF(2)`. The construction, its parameters, and the query
+algorithm are specified in [Schanck 2025][] and this document does not restate
+them. Where the on-disk structures defined here carry values that parameterize
+a clubcard query, they are described as inputs to the algorithm in that paper.
+
+A single clubcard filter does not cover the whole Web PKI. Each filter is
+scoped to a specific set of CT logs and a specific window of SCT timestamps
+within those logs. In practice, a client maintains a **revocation cache
+directory** containing several filter files whose scopes together cover the
+CT logs and time ranges the client cares about. A given certificate's SCTs
+determine which filters, if any, must be consulted. Filter scopes may overlap,
+so a single certificate can be covered by more than one filter, and a conclusive
+answer can come from any covering filter, so each must be consulted.
+
+Operationally, a deserialized clubcard combines three pieces of data. An
+**approximate filter** is a compact bit matrix that, for most queries, quickly
+rules the key out of the encoded set. An **exact filter** is a second, denser
+structure consulted for keys the approximate filter does not rule out. Together
+the two structures exactly reproduce the membership of the set the filter was
+built from. Lastly, an **exception list** carries the small number of lookup
+keys that the linear-system construction could not encode, and whose membership
+is answered by direct lookup. The wire format in this document carries one
+approximate filter, one exact filter, and a per-issuer exception list.
+
+Selecting the covering filters efficiently is the job of a binary index file
+(`index.bin`) that sits alongside the filters in the cache. The index maps
+each `(CT log ID, SCT timestamp)` pair a client might see to the names of the
+filter files that cover it, so a query need not open or parse any filter file
+that cannot contribute to the answer. The client then parses each covering
+filter file, performs a clubcard query against it, and aggregates the
+per-filter results to determine whether the certificate is revoked.
+
+This document specifies:
+
+- The layout of a cache directory containing one or more clubcard filter files.
+- The version 4 wire format of clubcard-encoded CRLite filter files.
+- The version "upkiidx1" upki binary index (`index.bin`) that maps `(CT log ID,
+  SCT timestamp)` pairs to the filter files that cover them.
+- The general procedure a client follows to select the filters covering
+  a given certificate, query each of them, and combine the results, deferring
+  to [Schanck 2025][] for the query itself.
+
+Filter versions 0 through 3, and the index version "upkiidx0" predate this
+specification and are not covered here. An implementation of this specification
+MAY reject filters whose version identifier is not 4, and index files with
+a version identifier that is not "upkiidx1".
+
+## Revocation cache directory layout
+
+A conforming implementation reads revocation data from a directory (the
+**revocation cache directory**) with the following layout:
+
+```
+    <revocation-cache-dir>/
+        index.bin
+        <clubcard-filter-file-1>
+        <clubcard-filter-file-2>
+        ...
+```
+
+`index.bin` is the [Index file format](#index-file-format), described below,
+and identifies the filter files by name.
+
+Filter files are ordinary files whose bytes conform to the [Filter file
+format](#filter-file-format). Their names MUST be between 1 and 32 bytes
+long and MUST consist only of the ASCII characters `A`–`Z`, `a`–`z`,
+`0`–`9`, hyphen (`-`), period (`.`), and underscore (`_`). The names `.`
+and `..` MUST NOT be used. Beyond those restrictions, filenames are opaque
+to this specification.
+
+There MAY be additional metadata files in the cache directory that describe when
+the filters were fetched, their expected checksums, etc. These details are
+outside of the scope of this specification and are not required for efficient
+revocation checking.
+
+## Clubcard Parameters
+
+A general clubcard implementation is parameterized and for this specification
+and the CRLite context, we fix the parameters to the following concrete values:
+
+- **Hash function** `H`. We select `H = SHA-256`, producing a 32-byte digest.
+- **Log identifier length** `L_id`. We select `L_id = 32` bytes, matching the CT
+  `LogID` defined in Section 3.2 of [RFC 6962][]. Concretely this is the SHA-256
+  digest of a log's public key.
+- **Clubcard query width** `W`. We select `W = 4`. Each query is expressed as
+  a linear equation over `W` distinct 64-bit lanes (totalling 256 bits).
+
+## Filter file format
+
+A single filter file within the revocation cache directory is a self-contained
+encoding of a single clubcard together with the CRLite metadata needed to
+interpret it:
+
+* the CT logs it covers,
+* the timestamp intervals within those logs,
+* and the per-issuer parameters that let a client query it.
+
+The top-level structure is:
+
+```
+    struct {
+        uint8 version;
+        uint8 reserved0;
+        CRLiteCoverage coverage;
+        ClubcardIndex index;
+        ApproximateFilter approximate;
+        ExactFilter exact;
+    } FilterFile;
+```
+
+The `version` value MUST be 4. An implementation MAY reject a file whose
+`version` is not 4.
+
+The `reserved0` byte MUST be `0x00`. An implementation MUST reject a file whose
+`reserved0` is not zero. Future versions of the format may use a non-zero value
+to signal the presence of additional metadata (for example describing
+a partition of the filter) that is omitted in version 4.
+
+### Coverage
+
+The `CRLiteCoverage` structure enumerates the CT logs and timestamp ranges
+that were included when the filter was built. A query that is not scoped to at
+least one covered `(LogID, timestamp)` pair yields **not covered** for this
+filter (see [Revocation check procedure](#revocation-check-procedure)).
+
+``` opaque LogID[32];
+
+    struct {
+        uint64 low;                 /* inclusive, milliseconds since Unix epoch */
+        uint64 high;                /* inclusive, milliseconds since Unix epoch */
+    } TimestampInterval;
+
+    struct {
+        LogID log_id;
+        TimestampInterval interval;
+    } CoverageEntry;
+
+    struct {
+        uint16        count;
+        CoverageEntry entries[count];
+    } CRLiteCoverage;
+```
+
+Each `CoverageEntry` is exactly 48 bytes on the wire. Entries MUST have
+distinct `log_id` values. Their order is not significant.
+
+Timestamps are non-negative integers denoting milliseconds since
+`1970-01-01T00:00:00Z`, ignoring leap seconds, as used by [RFC 6962][] SCTs.
+
+### Per-issuer index
+
+The `ClubcardIndex` maps each covered certificate issuer, identified by its
+32-byte SPKI hash, to the parameters needed to query the clubcard filter for
+serial numbers issued by that issuer. An issuer absent from this index was
+not enrolled in the filter when it was built and queries for such an issuer
+yield **not enrolled** (see [Revocation check
+procedure](#revocation-check-procedure)).
+
+```
+    opaque IssuerSPKIHash[32];
+
+    opaque Exception<1..2^8-1>;
+
+    struct {
+        uint32    approx_m;            /* rows in the approximate matrix */
+        uint8     approx_rank;         /* columns in the approximate matrix */
+        uint32    approx_offset;       /* row offset into the approximate matrix */
+        uint32    exact_m;             /* rows in the exact matrix */
+        uint32    exact_offset;        /* row offset into the exact matrix */
+        uint8     inverted;            /* 0 or 1; result inversion flag */
+        uint16    exception_count;
+        Exception exceptions[exception_count];
+    } IssuerBlock;
+
+    struct {
+        IssuerSPKIHash issuer;
+        IssuerBlock    block;
+    } IssuerEntry;
+
+    struct {
+        uint32      count;
+        IssuerEntry entries[count];
+    } ClubcardIndex;
+```
+
+The `IssuerEntry` list MUST be sorted lexicographically by `issuer` and MUST
+NOT contain duplicate `issuer` values. Sorting permits a binary search when
+locating the block for a given issuer during a query, and the sorted order is
+part of the wire format. An implementation that writes filter files MUST
+sort, and an implementation that reads filter files MAY assume sorted order.
+
+The `exceptions` list contains certificate serial numbers, each as a
+variable-length opaque byte string, that a builder could not encode via the
+linear system and therefore stores explicitly. Each serial is prefixed by a
+single byte giving its length in bytes, so an individual exception serial is
+at most 255 bytes.
+
+Each field of `IssuerBlock` is an input to the clubcard query defined in
+[Schanck 2025][]. `approx_m`, `approx_rank`, and `approx_offset` parameterize
+the query into the approximate filter for this issuer's block. `exact_m` and
+`exact_offset` parameterize the query into the exact filter. `inverted` is
+a per-block polarity flag and MUST be either 0 or 1. `exceptions` is the
+block's exception list, and each entry is a variable-length certificate serial
+number. Implementations MUST reject a filter file whose `inverted` value is not
+0 or 1.
+
+### Approximate filter
+
+Conceptually, the approximate filter is a large bit matrix `X` that is
+a solution to a homogeneous matrix equation `A * X = 0` derived from the encoded
+revocation set ([Schanck 2025][]). `X` is shared by every `IssuerEntry` in the
+same `ClubcardIndex` and each issuer block's queries read the leading
+`approx_rank` columns of `X`, anchored within the `approx_m` consecutive rows
+starting at row `approx_offset` of that block.
+
+On the wire, `X` is stored as a sequence of columns:
+
+```
+    struct {
+        uint32 word_count;
+        uint64 words[word_count];   /* each word is a big-endian uint64 */
+    } BitColumn;
+
+    struct {
+        uint8     count;
+        BitColumn columns[count];
+    } ApproximateFilter;
+```
+
+Bits within a `BitColumn` are indexed from 0 starting at the least significant
+bit of `words[0]`. That is, bit `i` (`0 <= i < 64 * word_count`) lives at
+`words[i / 64] >> (i % 64) & 1`. Bits at indices beyond `64 * word_count - 1`
+are **implicitly equal to 0**. A query can depend on bits that are not encoded
+in the `BitColumn`, so an implementation MUST evaluate such bits as 0.
+
+Only the region of `X` that can hold non-zero bits is explicitly encoded.
+Column `i` of `X` is read only by blocks whose `approx_rank` is greater than
+`i`, which shapes the encoding as follows:
+
+- `count` equals the largest `approx_rank` across all issuer blocks in the
+  `ClubcardIndex`. An implementation MUST treat a filter file in which any
+  block's `approx_rank` exceeds `count` as malformed.
+
+- Column `i` explicitly encodes rows only up to the largest value of
+  `approx_offset + approx_m` among blocks with `approx_rank > i`. Builders
+  assign smaller `approx_offset` values to blocks with larger `approx_rank`,
+  so the explicitly encoded region of `X` has a staircase shape where successive
+  columns cover progressively fewer rows.
+
+- A builder MAY further omit trailing all-zero words from any column. Each
+  `BitColumn` therefore carries its own `word_count`. A reader MUST NOT
+  assume any relationship between the `word_count` values of different
+  columns, or between a column's `word_count` and any block's
+  `approx_offset + approx_m`.
+
+A block's query evaluates one 256-bit-wide linear equation (the query width
+`W = 4`, in 64-bit lanes) against each of the block's leading `approx_rank`
+columns. Each evaluation reads a rectangle of bits spanning rows
+`[s, s + 256)` of columns `0` through `approx_rank - 1`, where `s` is a row
+derived from the query key satisfying
+`approx_offset <= s < approx_offset + approx_m`. The rectangle may extend
+past the explicitly encoded words of a column and bits read there are implicitly
+zero, as described above.
+
+### Exact filter
+
+The exact filter is a single column with the same encoding as one `BitColumn`
+of the approximate filter:
+
+```
+    struct {
+        uint32 word_count;
+        uint64 words[word_count];
+    } ExactFilter;
+```
+
+Bits are indexed within `words` in the same manner as an approximate filter
+`BitColumn`, including the implicit-zero rule.
+
+## Index file format
+
+The `index.bin` file is a binary lookup table. It maps `(LogID, timestamp)`
+pairs to the filter files that cover them, storing enough metadata to select the
+covering filter files for a query without loading each filter file into memory.
+It is designed for memory-mapped or seek-based access: an implementation can
+load the header and the two tables once, then use file offsets to fetch per-log
+data on demand.
+
+### High-level structure
+
+```
+    struct {
+        Header header;
+        FilenameEntry filenames[header.num_filenames];
+        LogDirEntry   log_directory[header.num_log_ids];
+        opaque        entry_sections[];
+    } IndexFile;
+```
+
+`entry_sections` is a variable-length region containing one contiguous run of
+per-log entries for each entry in `log_directory`, addressed by the `offset`
+field of that entry. Its total length is determined implicitly by the file
+size and the offsets in the log directory.
+
+### Header
+
+```
+    struct {
+        opaque magic[8];            /* MUST be "upkiidx1" */
+        uint16 num_filenames;
+        uint32 num_log_ids;
+    } Header;
+```
+
+The `Header` is exactly 14 bytes. An implementation MUST reject a file whose
+first 8 bytes are not the literal ASCII string `upkiidx1` (`0x75 0x70 0x6b 0x69
+0x69 0x64 0x78 0x31`).
+
+The `num_filenames` field limits the cache to at most 65535 filter files
+addressable by a single index. `num_log_ids` gives the number of CT logs
+covered by this index across all filter files.
+
+### Filename table
+
+```
+    struct {
+        opaque name[32];            /* UTF-8, zero-padded to 32 bytes */
+    } FilenameEntry;
+```
+
+The filename table is `32 * num_filenames` bytes long and begins immediately
+after the header, at file offset 14.
+
+Each `name` slot holds a filter file's name, padded with `0x00` bytes to 32
+bytes total. There is no explicit length field; the actual name is the byte
+prefix up to (but not including) the first `0x00`, or all 32 bytes if none
+is present. An implementation MUST treat an index file containing a name
+that violates the restrictions in [Revocation cache directory
+layout](#revocation-cache-directory-layout) as malformed.
+
+The position of a name in this table is its **filename index**, a `uint16`
+value in the range `[0, num_filenames)`. Entry sections (defined below) refer
+to filenames by this index rather than by name.
+
+### Log directory
+
+```
+    struct {
+        LogID  log_id;
+        uint64 offset;       /* file offset, in bytes, to this log's entry section */
+        uint16 num_entries;  /* number of Entry values in that section */
+    } LogDirEntry;
+```
+
+The log directory is `42 * num_log_ids` bytes long and immediately follows
+the filename table.
+
+Entries MUST be sorted lexicographically by `log_id` as raw 32-byte
+strings, matching the clubcard file format index sort. This ordering permits
+a binary search when resolving a `LogID`. Duplicate `log_id` values MUST NOT
+appear.
+
+The `offset` field is measured from the start of the file. The referenced
+entry section MUST lie entirely within the file and MUST NOT overlap the
+header, filename table, or log directory. Overlap between distinct logs'
+entry sections is unspecified and implementations SHOULD reject files where
+it occurs.
+
+### Entry sections
+
+Each log's entry section is a packed sequence of `Entry` values:
+
+```
+    struct {
+        uint16 filter_idx;          /* index into the filename table */
+        uint64 min_timestamp;       /* inclusive, milliseconds since Unix epoch */
+        uint64 max_timestamp;       /* inclusive, milliseconds since Unix epoch */
+    } Entry;
+```
+
+An `Entry` is exactly 18 bytes. A log's entry section is `18 * num_entries`
+bytes, where `num_entries` is the value in the corresponding `LogDirEntry`.
+
+An `Entry` asserts that the filter named by `filenames[filter_idx]` covers
+the corresponding `LogID` for SCT timestamps `t` satisfying
+`min_timestamp <= t <= max_timestamp`.
+
+Multiple entries for the same log MAY describe overlapping or adjacent intervals
+covered by different filter files. When more than one entry matches a query
+timestamp, every matching entry identifies a covering filter and the check
+procedure consults all of them: a covering filter that cannot answer for a
+certificate's issuer must not mask another covering filter that can (see
+[Revocation check procedure](#revocation-check-procedure)).
+
+`filter_idx` MUST be strictly less than `num_filenames`. An implementation
+MUST treat a file containing an out-of-range `filter_idx` as malformed.
+
+## Cache atomicity
+
+Cache updates may happen concurrently with revocation checks. While cache
+population is out of scope, this section describes the contract between the
+process populating the cache and the process reading from it, so that a check in
+flight during an update doesn't encounter inconsistent results.
+
+A conforming reader relies on the following invariants from whatever process
+populates the cache:
+
+1. Each individual file in the cache directory (`index.bin` and each filter
+   file) MUST be replaced atomically. The new bytes must appear at the file's
+   path in a single filesystem operation, with no observable intermediate state
+   in which the path resolves to a truncated or partially-written file. On POSIX
+   systems this is typically achieved by writing to a temporary file in the same
+   directory and renaming after writing the data.
+
+2. Before an `index.bin` referencing a filter file is added to the cache, the
+   filter file bytes MUST be written to disk first, before any `index.bin` that
+   references it becomes visible at `index.bin`'s path.
+
+3. When a filter file is removed from the cache, its removal MUST NOT be
+   observable until after an `index.bin` that no longer references it has become
+   visible at `index.bin`'s path.
+
+Together, these invariants ensure that any `index.bin` reader may observe
+references only filter files that are present in the cache directory at the
+moment `index.bin` becomes visible.
+
+The reader relies on these invariants during a check:
+
+1. When a check begins, the reader opens `index.bin` and SHOULD hold the
+   resulting file handle open for the entire duration of the check. On POSIX
+   systems the handle keeps the underlying inode alive after a subsequent rename
+   replaces the file at `index.bin`'s path, so on-demand reads of entry sections
+   through the handle remain consistent with the header and tables the reader
+   parsed at open time. New `index.bin` contents won't be read until the handle
+   is closed and re-opened by its path for subsequent checks.
+
+2. Filter files are opened by name after index lookup and not through a handle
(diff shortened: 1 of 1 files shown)


upki-revocation.md
```diff
@@ -0,0 +1,611 @@
...
+    } IssuerEntry;
+
+    struct {
+        uint32      count;
+        IssuerEntry entries[count];
+    } ClubcardIndex;
+```
+
+The `IssuerEntry` list MUST be sorted lexicographically by `issuer` and MUST
+contain no duplicate `issuer` values. Sorting permits a binary search when
```
ctz:
```suggestion
The `IssuerEntry` list MUST be sorted lexicographically by `issuer` and MUST NOT
contain duplicate `issuer` values. Sorting permits a binary search when
```
seems more RFC2119

upki-revocation.md
```diff
@@ -0,0 +1,611 @@
...
+
+**Malformed inputs.** Filter files and `index.bin` are parsed from untrusted
+bytes if the cache authenticity check is imperfect. Implementations MUST
+validate all length prefixes, offsets, and index values against the bounds
+implied by the file size and the values already read, and MUST NOT allocate
+memory proportional to length prefixes without first bounding those lengths
+by the remaining file size. The wire format contains multiple opportunities
+for pathological input (e.g. nested variable-length fields, `filter_idx` values
+into the filename table, arbitrary `offset` values in the log directory) and
+and each MUST be checked for safe operation.
```
ctz:
```suggestion
into the filename table, arbitrary `offset` values in the log directory) and
each MUST be checked for safe operation.
```

ctz (commented):
couple of minor nits from me from an initial read

upki-revocation.md
```diff
@@ -0,0 +1,611 @@
...
+[Schanck 2025]: https://doi.org/10.1109/SP61157.2025.00128
+
+## Introduction
+
+CRLite is a scheme for compactly representing the set of revoked
+publicly-trusted TLS certificates so that a client can answer "is this
+certificate revoked?" locally, without contacting a certificate authority or
+OCSP responder. It is space efficient compared to independent CRL files and
+can support the scale of the Web PKI. In its current form the compact
+representation is a probabilistic set membership data structure, a **clubcard
```
djc: Should the privacy benefit be mentioned here explicitly?

upki-revocation.md
```diff
@@ -0,0 +1,611 @@
...
+document specifies how an implementation should notice that its cache is stale,
+or refresh the data.
+
+**"Not covered" semantics.** The **not covered** outcome asserts that this
+mechanism has no evidence about the certificate. It does not assert that
+the certificate is valid and unrevoked. Callers MUST decide the meaning of **not
+covered** in their trust model (fail-open, fail-closed, or delegate to another
+revocation mechanism). Treating **not covered** as equivalent to
+**not revoked** is a policy choice with security consequences and is NOT
+RECOMMENDED without a compensating control. Similar policy consideration
```
djc: "NOT RECOMMENDED" might be overly stern here? I think this would be a common choice since in the upki model you cannot really guarantee freshness.

djc (commented):
Mostly looks good to me. I wonder if this needs treatment of atomic operations, as in, how does an index/filter consumer correctly handle the risk of a cache update operation happening while it's making revocation checks?

upki-revocation.md
cpu: Yeah that reads better. Thank you! Applied.

upki-revocation.md
cpu: Good catch, thanks. Fixed up.

upki-revocation.md
cpu: Good point. I worked in a brief mention of this.

upki-revocation.md
cpu: Fair. I removed this line entirely. I think it's enough to say that callers have to give this thought and apply their own policy. Probably there's no one universally correct choice we should recommend (but if you disagree, happy to try and add more text here).

cpu:
>  I wonder if this needs treatment of atomic operations, as in, how does an index/filter consumer correctly handle the risk of a cache update operation happening while it's making revocation checks?

That's a good point. I added an explicit section on this point and added some cross-refs from other sections. I think I have the idea right, but if you want to spot-check it against what `upki` does and suggest tweaks to the wording I'd appreciate it.

upki-revocation.md
```diff
@@ -0,0 +1,667 @@
...
+within those logs. In practice, a client maintains a **revocation cache
+directory** containing several filter files whose scopes together cover the
+CT logs and time ranges the client cares about. A given certificate's SCTs
+determine which filter, if any, is the one to consult.
+
+Selecting that filter efficiently is the job of a binary index file
+(`index.bin`) that sits alongside the filters in the cache. The index maps
+each `(CT log ID, SCT timestamp)` pair a client might see to the covering
+filter file's name, so a query need not open or parse any filter file until
+the right one has been identified. Once identified, the client parses that
```
cpu: leaving myself a note here to iterate: John mentioned OOB it's incorrect to specify this as _one_ matching filter. In actuality there is a sequence of covering filters and you need to query all of them.

djc:
> > I wonder if this needs treatment of atomic operations, as in, how does an index/filter consumer correctly handle the risk of a cache update operation happening while it's making revocation checks?
> 
> That's a good point. I added an explicit section on this point and added some cross-refs from other sections. I think I have the idea right, but if you want to spot-check it against what `upki` does and suggest tweaks to the wording I'd appreciate it.

LGTM. Maybe good if @ctz also takes a look.

upki-revocation.md
```diff
@@ -0,0 +1,667 @@
...
+There MAY be additional metadata files in the cache directory (e.g.
+`manifest.json`) that describes when the filters were fetched, their expected
+checksums, etc. These details are outside of the scope of this specification and
+are not required for efficient revocation checking.
+
+Populating the cache directory, obtaining `index.bin` and the filter files
+it names, verifying their integrity, and deciding when and how to refresh them
+is out of scope for this document.
+
+## Cache atomicity
```
warpfork: I'm so happy this section exists.  🙏 

Making things like this work with pure posix fs atomicity (instead of daemon shenanigans) is beautiful and excellent.  (And even beyond no-daemon's obvious superiority, in practice, I can imagine the filesystem access spec will become concretely useful where it lets systems of nested virtualization use this in clear ways.)

upki-revocation.md
cpu: I've updated the text to describe that multiple filters need to be queried, and to describe the individual filter result aggregation.

cpu:
I believe I've addressed all feedback received so far. John mentioned he's out of office until next week but can give a review then. My plan is to wait for that, apply any additional updates requested, and then bug some folks for approvals to merge afterwards.

upki-revocation.md
```diff
@@ -0,0 +1,708 @@
...
+Mozilla [CRLite][] filters encoded in the [clubcard][] v4 format, distributed by
+[upki][]. It is aimed at implementors of clients that consume such a cache
+in order to consult it for efficient revocation decisions during certificate
+path validation.
+
+The cache consists of one or more encoded clubcard filter files together with a
+binary index that maps `(CT log ID, SCT timestamp)` pairs to the covering
+filter files in the cache. Given an end-entity certificate and its
+embedded [SCTs][RFC 6962], an implementation can determine one of three
+outcomes: **revoked**, **not revoked**, or **not covered** using the available
```
jschanck: There's also a **not enrolled** outcome which indicates that the issuing CA does not chain up to the Mozilla root store. UPKI [does not distinguish](https://github.com/rustls/upki/blob/30fda4a3c581fc45e185d8ad70b016280f97c36a/upki/src/revocation/index.rs#L293) between **not covered** and **not enrolled** but other clients might want to.

upki-revocation.md
```diff
@@ -0,0 +1,708 @@
...
+range. 
+
+`||` denotes concatenation. 
+
+`SHA-256(x)` denotes the SHA-256 digest of `x` as specified in [FIPS 180-4][],
+producing a 32-byte value.
+
+The **SPKI hash** of a certificate is the SHA-256 digest of the DER encoding of
+its `SubjectPublicKeyInfo` field, as defined in [RFC 5280][].
+
```
jschanck, line 55: We also need to define the serial number to make it clear that it may include a leading `00` byte from DER encoding.

> The **Serial Number** of a certificate is the DER encoding of its `Certificate Serial Number` field, as defined in [RFC5280][], with the tag and length octets removed.

(update: I see this is spelled out farther down in the document, up to you if you want to hoist it up here).

upki-revocation.md
```diff
@@ -0,0 +1,708 @@
...
+[Schanck 2025]: https://doi.org/10.1109/SP61157.2025.00128
+
+## Introduction
+
+CRLite is a scheme for compactly representing the set of revoked
+publicly-trusted TLS certificates so that a client can answer "is this
+certificate revoked?" locally, without contacting a certificate authority or
+OCSP responder. It is space efficient compared to independent CRL files, leaks
+no private information to the CRL/OCSP issuer and can support the scale of the
+Web PKI. In its current form the compact representation is a probabilistic set
```
jschanck: drop "probabilistic". The construction uses randomness but the encoding is exact.

upki-revocation.md
```diff
@@ -0,0 +1,708 @@
...
+        uint16        count;
+        CoverageEntry entries[count];
+    } CRLiteCoverage;
+```
+
+Each `CoverageEntry` is exactly 48 bytes on the wire. Entries MUST have
+distinct `log_id` values. Their order is not significant.
+
+Timestamps are non-negative integers denoting milliseconds since
+`1970-01-01T00:00:00Z`, as used by [RFC 6962][] SCTs.
```
jschanck: nit: add "ignoring leap seconds" to align with phrasing in RFC 6962.

upki-revocation.md
```diff
@@ -0,0 +1,708 @@
...
+number. Implementations MUST reject a filter file whose `inverted` value is not
+0 or 1.
+
+### Approximate filter
+
+The approximate filter is a bit matrix stored as a sequence of columns, one
+per bit of `approx_rank` across all issuer blocks. All columns have the same
+number of rows, drawn from the same underlying bit array; each block's query
+uses `approx_m` consecutive rows starting at `approx_offset` of its
+corresponding block.
```
jschanck: Based on this description you would expect each column to have the same `word_count`, which isn't the case. The diagram on page 30 of https://jmschanck.info/talks/rwc2025.pdf might be useful. The approximate filter is a solution `X` to a matrix equation `A * X = 0`. But what's encoded is just the pink staircase shape that is shown in the diagram. The number of columns is equal to the maximum value of `approx_rank` across all issuer blocks. The number of rows in column `i` is the maximum value of `approx_offset + approx_m` among issuer blocks with `approx_rank > i`.

We might also mention that a query to some `block` involves a rectangle of bits `X[s:s+256, 0:block.approx_rank]` where `s` is some value in `[block.approx_offset, block.approx_offset + block.approx_m]`. This might involve bits that are not explicitly encoded by the columns. Those bits are implicitly set to 0.

upki-revocation.md
```diff
@@ -0,0 +1,708 @@
...
+columns and ignores the rest.
+
+Each `BitColumn` carries its own `word_count`, and different columns may be of
+different lengths.
+
+Bits within a `BitColumn` are indexed from 0 starting at the least significant
+bit of `words[0]`. That is, bit `i` (`0 <= i < 64 * word_count`) lives at
+`words[i / 64] >> (i % 64) & 1`. Bit indices beyond `64 * word_count - 1` do
+not exist and MUST NOT be queried and a filter file whose block parameters would
+require reading such an index is malformed.
```
jschanck: A query can actually depend on bits that are not encoded in the `BitColumn`. The `min()` [here](https://github.com/mozilla/clubcard/blob/ecbd7413c7586bd37bc898c7096c3e5746644ff9/src/equation.rs#L88) ensures that we treat those bits as being equal to 0.
 
```suggestion
`words[i / 64] >> (i % 64) & 1`. Bits at indices beyond `64 * word_count - 1` are implicitly equal to 0.
```

upki-revocation.md
```diff
@@ -0,0 +1,708 @@
...
+The `ApproximateFilter` is shared by every `IssuerEntry` in the same
+`ClubcardIndex`, but different issuers may require different numbers of
+columns: each `IssuerEntry` specifies its own `approx_rank`, giving the number
+of columns that block's query reads. The `columns` field MUST therefore
+contain at least as many columns as the largest `approx_rank` in the index.
+A block whose `approx_rank` is smaller reads only the leading `approx_rank`
+columns and ignores the rest.
+
+Each `BitColumn` carries its own `word_count`, and different columns may be of
+different lengths.
```
jschanck: This is correct, so maybe just drop "All columns have the same number of rows, drawn from the same underlying bit array" above.

upki-revocation.md
```diff
@@ -0,0 +1,708 @@
...
+  when used as a revocation input.
+- `scts`: the list of Signed Certificate Timestamps embedded in the
+  end-entity certificate under the extension defined in Section 3.3 of [RFC
+  6962][]. Each SCT contributes a `(LogID, timestamp)` pair. Timestamps are
+  treated as `uint64` milliseconds since Unix epoch as they appear on the wire.
+
+If the certificate carries no SCTs, the implementation SHOULD return
+**not covered** without consulting the cache. Detached SCTs delivered via
+TLS handshake or OCSP in place of an extension in the end-entity certificate 
+MAY be accepted as inputs at the implementation's discretion.
```
jschanck: This reads as though an implementation cannot rely exclusively on detached SCTs. Is that intentional? If not, maybe reorder the sentences and rephrase "carries no SCTs":

```suggestion
Detached SCTs delivered via TLS handshake or OCSP in place of an extension in the end-entity certificate 
MAY be accepted as inputs at the implementation's discretion. If the certificate is not presented with any SCTs, then the implementation SHOULD return **not covered** without consulting the cache.
```

We might also want to mention that the signatures on SCTs should be validated (a MUST for  detached SCTs and a SHOULD for embedded SCTs, although an across the board MUST is also reasonable).

upki-revocation.md
```diff
@@ -0,0 +1,708 @@
...
+      `min_timestamp <= timestamp <= max_timestamp` identifies a covering
+      filter, whose filename relative to the cache directory is
+      `filenames[filter_idx]` from the index [Filename
+      table](#filename-table). Collect each such filename. An
+      implementation MUST NOT stop at the first match.
+
+3. If no SCT yields any covering filter, return **not covered**.
+
+The filenames collected across all SCTs, in the order encountered, are the
+check's **covering filters**.
```
jschanck: I would drop the "in the order encountered" clause, as it (a) doesn't matter, and (b) might not be optimal. Certificates are more likely to be revoked shortly after issuance than at any other time. So covering filters should be queried in the order that they were published.

upki-revocation.md
```diff
@@ -0,0 +1,708 @@
...
+
+With a per-filter outcome from step 2 in hand for every covering filter,
+determine the check's overall result:
+
+1. If any covering filter yielded **revoked**, return **revoked**.
+
+2. Otherwise, if any covering filter yielded **not revoked**, return
+   **not revoked**.
+
+3. Otherwise, every covering filter yielded **not covered**; return
```
jschanck: Isn't this impossible? This indicates some issue with the index.

upki-revocation.md
```diff
@@ -0,0 +1,708 @@
...
+into the filename table, arbitrary `offset` values in the log directory) and
+each MUST be checked for safe operation.
+
+**Query key collisions.** The query key is `SHA-256(issuer_spki_hash ||
+serial)`. Collisions in SHA-256 within a single issuer's serial-number space
+would produce incorrect results, but constructing such a collision is
+computationally infeasible under current assumptions without compromising the
+issuing CA. This revocation mechanism targets the Web PKI where issuance of
+duplicated serials by a single issuer is a mississuance event.
+
```
jschanck, line 765: We might want a section on correctness considerations. There are two sharp edges:
1) MMD violations. Coverage checks assume proper log operation. If a certificate is not visible at a log's read endpoint within one MMD of its SCT timestamp, then the aggregator will not actually encode its revocation status in a covering filter. This means that the covering filter can report either a false positive or a false negative.
2) Exclusive reliance on SCTs issued after revocation. Such SCTs may not assert coverage for the filter that encodes the certificate's revoked status, which leads to false negatives. See also: https://github.com/mozilla/crlite/issues/367 and https://github.com/mozilla/crlite/issues/376#issuecomment-3529074397.

upki-revocation.md
```diff
@@ -0,0 +1,708 @@
...
+also been validated in this manner is meaningless and MUST NOT be used to accept
+a connection.
+
+**SCT trust is external.** The `(LogID, timestamp)` pairs supplied by SCTs are
+used here purely as index keys for selecting covering filters. This
+specification does not verify SCT signatures, does not enforce a CT-log-trust
+policy (e.g. minimum number of SCTs, SCT diversity), and does not require that
+the logs named by the SCTs be logs the caller trusts. Callers that require SCTs
+to be authenticated, or that require a particular CT log inclusion policy, MUST
+enforce those requirements separately.
```
jschanck: I agree with what's written here in so far as we're saying it's the caller's responsibility to check SCT signatures and enforce CT policy. But we should warn callers against passing in SCTs that have not been validated.

```suggestion
the logs named by the SCTs be logs the caller trusts. Callers with specific requirements on SCTs MUST enforce those requirements separately. Callers SHOULD verify SCT signatures. Covering filters that are selected based on malicious or malformed SCTs may give inaccurate results.
```

jschanck (commented):
Awesome! Thank you for putting this together! A few minor comments below.

upki-revocation.md
cpu: Good call. I've updated the text throughout to talk about this outcome and allow for an impl to treat them equivalently if desired.

upki-revocation.md
cpu: I lifted the serial number details from lower into the definitions up-front (and kept a short call-out RE: the leading 0x00 in the original spot that now ref's the definition).

upki-revocation.md
cpu: Done.

upki-revocation.md
cpu: Done.

upki-revocation.md
cpu: I reworked this to hopefully be more accurate. Suggested tweaks welcome! :bow:  I tried to err on the side of not providing _too_ much detail, still hoping to defer the mechanics of the querying to the original paper and focusing more on the serialization/interop aspects.

WDYT? Looks better or still needs work?

upki-revocation.md
cpu: Made this more explicit.

upki-revocation.md
cpu: This fell away with the rework of the approximation matrix section.

upki-revocation.md
cpu: Great suggestion. I went with the MUST for detached and SHOULD for embedded and reworked the text. I was hesitant to use MUST for both because the rustls/upki impl would be out-of-spec and I don't think they want to have to deal with keeping an up-to-date log list w/ log metadata to do the verification right now.

upki-revocation.md
cpu: Good points. I dropped the "in the order encountered" and added a specific bit about potentially doing it in the order of covering filter publish date. That information isn't encoded in the index right now, so I don't want to make it mandatory, but I think with the `manifest.json` data the upki fetcher writes (that I've considered out-of-scope for this text), an impl could reasonably do it as an optimization (though AFAICT rustls/upki isn't doing that today).

upki-revocation.md
cpu: Another good point. I dropped the bullet and added a note about this being an error condition indicating an issue with the index.

upki-revocation.md
cpu: Ahh yes, agreed these are important details. I've added a short paragraph covering both.

upki-revocation.md
cpu: I ended up making this section point back to more details in the inputs section, but I did also add an explicit note anchoring the risk that unvalidated SCTs can steer the revocation checking to produce incorrect results.

upki-revocation.md
```diff
@@ -0,0 +1,778 @@
...
+entry sections is unspecified and implementations SHOULD reject files where
+it occurs.
+
+### Entry sections
+
+Each log's entry section is a packed sequence of `Entry` values:
+
+```
+    struct {
+        uint8  filter_idx;          /* index into the filename table */
```
jschanck: 256 entries is probably fine, but a little tight.

Mozilla's aggregator publishes a new filter every 12 hours and clears its filter cache every 47 days. So we have no more than 94 live filters at any time. This is good for client-side query efficiency but bad for bandwidth efficiency. The first filter published in a 47 day cycle is much larger than typical delta updates (it includes information that had been previously been delivered in the filters that have been evicted from the cache).

On 2029-03-15 the maximum validity period for WebPKI certificates will drop to 47 days. Shortly thereafter we will stop clearing the cache every 47 days. Instead, every 12 hours we will drop the oldest filter (which only covers expired certificates) at the same time that we publish a new filter. We'll still have no more than 94 live filters at any given time, but our bandwidth costs will decrease significantly.

On 2027-03-15 the maximum validity period for WebPKI certificates will drop to 100 days. If we were satisfied with the client-side query efficiency impact of having 200 live filters, we could switch strategies sooner.

The current maximum validity period is 200 days. Switching strategies now would blow the 256 entry limit for the index. Publishing filters more frequently could also pose problems in the future.

upki-revocation.md
```diff
@@ -0,0 +1,778 @@
...
+
+* the CT logs it covers,
+* the timestamp intervals within those logs,
+* and the per-issuer parameters that let a client query it.
+
+The top-level structure is:
+
+```
+    struct {
+        uint16 version;             /* NOTE: little-endian */
```
jschanck: I'm open to changing to big-endian here. Version 4 is already a breaking change for Firefox, so now is the time to do it.

upki-revocation.md
cpu: So maybe a `uint16` would be better :thinking: 

@djc @ctz Are you open to that? I'm assuming you have extent `index.bin` instances out there that will need to be handled, so WDYT about:

* Changing this index field (& the header filename count) to `uint16` in the spec text
* Also changing the spec to use `upkiidx1` as the described index.bin magic
* Opening a PR to rustls/upki that updates fetch's index writing to use `upkiidx1` magic & `uint16 filename_count`/`uint16 filter_idx`, and that updates the deser side to infer the `filename_count`/`filter_idx` based on the magic, (`upkiidx0` -> `uint8`, `upkiidx1` -> `uint16`).

upki-revocation.md
cpu: It would be nice to make that consistent.

AIUI the Mozilla data hasn't switched over to V4 yet, so presumably we could land a PR changing this in `mozilla/clubcard-crlite`, publish a new release of that crate, take the update in `rustls/upki`, and everything would be OK and we don't need to burn V4 and move to V5? E.g. I don't think we need to worry about pre-existing V4 filter data written to caches anywhere?

(cc @djc @ctz)

ctz: Putting some other ideas out there:
- `uint8 version; uint8 reserved0;`? 
- `uint16 version;` and the current version is 1024.

cpu: > uint8 version; uint8 reserved0;?

I kind of like that idea :+1: If we take it, we should probably specify that reserved0 MUST be 0.

jschanck: +1 for `uint8 version; uint8 reserved0;`. It would actually be nice to use the second byte to encode the type of the `PartitionMetadata` field. The current version deserializes to a clubcard with an [empty](https://github.com/mozilla/clubcard-crlite/blob/52318cba67e252b57dd64b36073e22fd8b4c8ef5/src/query.rs#L245-L248) field, so we would indicate that by setting the second byte equal to `0`.

upki-revocation.md
jschanck: Looks great!

upki-revocation.md
djc: Sounds good to me!

upki-revocation.md
djc: > I'm open to changing to big-endian here. Version 4 is already a breaking change for Firefox, so now is the time to do it.

I was under the impression that your ESR thing had already shipped so we couldn't make any more changes? Otherwise, I think there might be benefit in making some more things little-endian instead of big-endian (because I think we could do in-place decoding).

cpu: The `uint8 version; uint8 reserved0;` idea would be drop-in compatible with whatever V4 encodings have already shipped.

upki-revocation.md
cpu: > Sounds good to me!

Here's the promised PR: https://github.com/rustls/upki/pull/200

I also retroactively updated the plan above. Besides the `filter_idx` the size of the filename count in the header needs to be upgraded as well.

upki-revocation.md
cpu: Here's a PR implementing the change in `clubcard-crlite`: https://github.com/mozilla/clubcard-crlite/pull/32

While it's wire-compatible, it's an API break so I bumped the major version. If that's a blocker we could keep the public API types using `u16` and handle it all internally, but that felt a bit confusing/suboptimal.

cpu:
Pushed updated text to reflect the index.bin `u8` -> `u16` changes, and the clubcard filter version `u16` little-endian version -> `u8` version, `u8` reserved changes.

Can I ask folks to do a final review and give positive confirmation if you think this is ready to be merged? We can of course iterate further as topics come up (_thankfully this isn't an immutable IETF artifact..._), but I think this is ready as a first draft.

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+for pathological input (e.g. nested variable-length fields, `filter_idx` values
+into the filename table, arbitrary `offset` values in the log directory) and
+each MUST be checked for safe operation.
+
+**Query key collisions.** The query key is `SHA-256(issuer_spki_hash ||
+serial)`. Collisions in SHA-256 within a single issuer's serial-number space
+would produce incorrect results, but constructing such a collision is
+computationally infeasible under current assumptions without compromising the
+issuing CA. This revocation mechanism targets the Web PKI where issuance of
+duplicated serials by a single issuer is a mississuance event.
```
jschanck:
```suggestion
duplicated serials by a single issuer is a misissuance event.
```

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+        uint8     count;
+        BitColumn columns[count];
+    } ApproximateFilter;
+```
+
+Bits within a `BitColumn` are indexed from 0 starting at the least significant
+bit of `words[0]`. That is, bit `i` (`0 <= i < 64 * word_count`) lives at
+`words[i / 64] >> (i % 64) & 1`. Bits at indices beyond `64 * word_count - 1`
+are **implicitly equal to 0**. A query can depend on bits that are not encoded
+in the `BitColumn` , so an implementation MUST evaluate such bits as 0.
```
jschanck:
```suggestion
in the `BitColumn`, so an implementation MUST evaluate such bits as 0.
```

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+   because the file does not exist, the reader MUST treat this as evidence that
+   the `index.bin` it is using has been superseded by a newer one that no
+   longer references the missing filter. The reader SHOULD retry the check
+   from the beginning, reopening `index.bin` at least once before returning an
+   error. An implementation MAY bound the number of retries to avoid unbounded
+   restart under pathological update patterns.
+
+## Index file format
+
+`index.bin` file is a binary lookup table. It maps `(LogID, timestamp)` pairs to
```
jschanck:
```suggestion
The `index.bin` file is a binary lookup table. It maps `(LogID, timestamp)` pairs to
```

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+specified in [Schanck 2025][] and this document does not restate them. Where the
+on-disk structures defined here carry values that parameterize a clubcard
+query, they are described as inputs to the algorithm in that paper.
+
+Operationally, a deserialized clubcard combines three pieces of data. An
+**approximate filter** is a compact bit matrix that, for most queries, quickly
+rules the key out of the encoded set. An **exact filter** is a second, denser
+structure consulted for keys the approximate filter does not rule out. Together
+the two structures exactly reproduce the membership of the set the filter was
+built from. Lastly an **exception list** carries the small number of lookup keys
```
jschanck:
```suggestion
built from. Lastly, an **exception list** carries the small number of lookup keys
```

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+[RFC 8174]: https://www.rfc-editor.org/rfc/rfc8174.html
+[Schanck 2025]: https://doi.org/10.1109/SP61157.2025.00128
+
+## Introduction
+
+CRLite is a scheme for compactly representing the set of revoked
+publicly-trusted TLS certificates so that a client can answer "is this
+certificate revoked?" locally, without contacting a certificate authority or
+OCSP responder. It is space efficient compared to independent CRL files, leaks
+no private information to the CRL/OCSP issuer and can support the scale of the
```
jschanck:
```suggestion
no private information to the CRL/OCSP issuer, and can support the scale of the
```

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+`SHA-256(x)` denotes the SHA-256 digest of `x` as specified in [FIPS 180-4][],
+producing a 32-byte value.
+
+The **SPKI hash** of a certificate is the SHA-256 digest of the DER encoding of
+its `SubjectPublicKeyInfo` field, as defined in [RFC 5280][].
+
+The **serial number** of a certificate is the DER-encoded contents (i.e. the
+`INTEGER` value bytes, excluding tag and length) of the end-entity certificate's
+`serialNumber` field, as specified in [RFC 5280][] §4.1.2.2. Importantly, if the
+DER encoding includes a leading `0x00` byte to disambiguate sign, that byte is
```
jschanck:
```suggestion
DER encoding includes a leading `0x00` byte to disambiguate the sign, that byte is
```

jschanck (approved):
Looks good! Just a few nits below.

upki-revocation.md
cpu: Fixed. I swear I've made this exact same typo somewhere around a bajillion times in my life...

upki-revocation.md
cpu: Fixed.

upki-revocation.md
cpu: Fixed.

upki-revocation.md
cpu: Fixed.

upki-revocation.md
cpu: Fixed.

upki-revocation.md
cpu: Fixed.

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+linear system and therefore stores explicitly. Each serial is prefixed by a
+single byte giving its length in bytes, so an individual exception serial is
+at most 255 bytes.
+
+Each field of `IssuerBlock` is an input to the clubcard query defined in
+[Schanck 2025][]. `approx_m`, `approx_rank`, and `approx_offset` parameterize
+the query into the approximate filter for this issuer's block. `exact_m` and
+`exact_offset` parameterize the query into the exact filter. `inverted` is
+a per-block polarity flag and MUST be either 0 or 1. `exceptions` is the
+block's exception list, and each entry a variable-length certificate serial
```
ctz:
```suggestion
block's exception list, and each entry is a variable-length certificate serial
```

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+and identifies the filter files by name.
+
+Filter files are ordinary files whose bytes conform to the [Filter file
+format](#filter-file-format). Their names MUST be at most 32
+bytes when encoded as UTF-8 and MUST NOT contain the byte `0x00` for
+compatibility with the index file format. Beyond those restrictions, filenames
+are opaque to this specification.
+
+There MAY be additional metadata files in the cache directory (e.g.
+`manifest.json`) that describes when the filters were fetched, their expected
```
ctz:
```suggestion
`manifest.json`) that describe when the filters were fetched, their expected
```

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+
+An individual filter that yields **not covered** says nothing about the
+certificate's revocation status and a filter that cannot answer for the
+certificate's issuer or timestamps must not mask another covering filter that
+can.
+
+An individual filter that yields **not enrolled** indicates that the
+certificate's issuer was not enrolled in the revocation data even though its
+timestamps were covered. Implementations that make no policy distinction between
+the two MAY collapse them, treating it as a **not covered** result. Like **not
```
ctz:
```suggestion
timestamps were covered. Implementations that make no policy distinction between
the two MAY collapse them, treating them as a **not covered** result. Like **not
```

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+1. If any covering filter yielded **revoked**, return **revoked**.
+
+2. Otherwise, if any covering filter yielded **not revoked**, return
+   **not revoked**.
+
+3. Otherwise, if every covering filter yielded **not enrolled**, return
+   **not enrolled**.
+
+An individual filter that yields **not covered** says nothing about the
+certificate's revocation status and a filter that cannot answer for the
```
ctz:
```suggestion
certificate's revocation status, and a filter that cannot answer for the
```

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+   `CRLiteCoverage`. A `(log_id, timestamp)` pair is covered when a
+   `CoverageEntry` exists whose `log_id` equals `log_id` and whose interval
+   satisfies `low <= timestamp <= high`. If no SCT is covered, the filter
+   yields **not covered**.
+
+2. Locate the `IssuerEntry` in the filter's `ClubcardIndex` whose `issuer`
+   equals `issuer_spki_hash`. If none exists, the filter yields
+   **not enrolled**: the filter covers a relevant timestamp window but was
+   built without data for this issuer. Note that because the coverage check
+   precedes the issuer lookup **not enrolled** is only reported by a filter that
```
ctz:
```suggestion
   built without data for this issuer. Note that, because the coverage check
   precedes the issuer lookup, **not enrolled** is only reported by a filter that
```

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+  to [Schanck 2025][] for the query itself.
+
+Filter versions 0 through 3 predate this specification and are not covered
+here. An implementation of this specification MAY reject filters whose
+version identifier is not 4.
+
+## Clubcard Parameters
+
+A general clubcard implementation is parameterized and for this specification
+and the CRLite context we fix the parameters to the following concrete values:
```
ctz:
```suggestion
A general clubcard implementation is parameterized, and for this specification
and the CRLite context, we fix the parameters to the following concrete values:
```

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+are not required for efficient revocation checking.
+
+Populating the cache directory, obtaining `index.bin` and the filter files
+it names, verifying their integrity, and deciding when and how to refresh them
+is out of scope for this document.
+
+## Cache atomicity
+
+Cache updates may happen concurrently with revocation checks. While cache
+population is out of scope this section describes the contract between the
```
ctz:
```suggestion
population is out of scope, this section describes the contract between the
```

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+
+### Inputs
+
+The implementation extracts the following from the end-entity certificate under
+evaluation and its issuing certificate:
+
+- `issuer_spki_hash`: the [SPKI hash](#conventions-used-in-this-document) of
+  the certificate that issued the end-entity certificate. This is a 32-byte
+  value.
+- `serial`: the DER encoded [serial number](#conventions-used-in-this-document)
```
ctz:
```suggestion
- `serial`: the DER-encoded [serial number](#conventions-used-in-this-document)
```

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+        uint16        count;
+        CoverageEntry entries[count];
+    } CRLiteCoverage;
+```
+
+Each `CoverageEntry` is exactly 48 bytes on the wire. Entries MUST have
+distinct `log_id` values. Their order is not significant.
+
+Timestamps are non-negative integers denoting milliseconds since
+`1970-01-01T00:00:00Z` ignoring leap seconds, as used by [RFC 6962][] SCTs.
```
ctz:
```suggestion
`1970-01-01T00:00:00Z`, ignoring leap seconds, as used by [RFC 6962][] SCTs.
```

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+[RFC 2119]: https://www.rfc-editor.org/rfc/rfc2119.html
+[RFC 8174]: https://www.rfc-editor.org/rfc/rfc8174.html
+[Schanck 2025]: https://doi.org/10.1109/SP61157.2025.00128
+
+## Introduction
+
+CRLite is a scheme for compactly representing the set of revoked
+publicly-trusted TLS certificates so that a client can answer "is this
+certificate revoked?" locally, without contacting a certificate authority or
+OCSP responder. It is space efficient compared to independent CRL files, leaks
```
ctz:
```suggestion
OCSP responder. It is space-efficient compared to independent CRL files, leaks
```

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+**Cache authenticity.** Nothing in the on-disk format signs or otherwise
+authenticates the filter files or `index.bin`. An implementation MUST
+obtain the cache contents through an authenticated channel and MUST verify
+its integrity before use. The mechanism for doing so is out of scope.
+
+**Freshness.** A filter reflects the set of revocations known at the time it
+was built. A recently issued and recently revoked certificate may not yet be
+represented in any cached filter, and the check procedure will return
+**not covered** or **not revoked** for such a certificate. Callers that need
+a hard freshness guarantee MUST enforce it externally and nothing in this
```
ctz:
```suggestion
a hard freshness guarantee MUST enforce it externally, and nothing in this
```

ctz (approved):
As you can tell, I'm scraping the bottom of the barrel here.

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+path validation.
+
+The cache consists of one or more encoded clubcard filter files together with a
+binary index that maps `(CT log ID, SCT timestamp)` pairs to the covering
+filter files in the cache. Given an end-entity certificate and its
+embedded [SCTs][RFC 6962], an implementation can determine one of four
+outcomes: **revoked**, **not revoked**, **not enrolled**, or **not covered**
+using the available revocation data.
+
+How the cache is populated (e.g. production, hosting, downloading, verification,
```
djc: As mentioned before elsewhere, I'd find reformulations for all uses of "e.g." and "i.e.".

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+
+## Introduction
+
+CRLite is a scheme for compactly representing the set of revoked
+publicly-trusted TLS certificates so that a client can answer "is this
+certificate revoked?" locally, without contacting a certificate authority or
+OCSP responder. It is space efficient compared to independent CRL files, leaks
+no private information to the CRL/OCSP issuer, and can support the scale of the
+Web PKI. In its current form the compact representation is a set membership data
+structure, a **clubcard filter**, combined with metadata describing the subset
```
djc, line 84: "clubcard filter" isn't really explained here. Maybe it's worth moving the paragraph from below up to here, or referencing it?

(In general, it feels like this section bounces a little bit between top-down and bottom-up and it might be worth being more strict about which order you want to aim for?)

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+- The version "upkiidx1" upki binary index (`index.bin`) that maps `(CT log ID,
+  SCT timestamp)` pairs to the filter files that cover them.
+- The general procedure a client follows to select the filters covering
+  a given certificate, query each of them, and combine the results, deferring
+  to [Schanck 2025][] for the query itself.
+
+Filter versions 0 through 3 predate this specification and are not covered
+here. An implementation of this specification MAY reject filters whose
+version identifier is not 4.
+
```
djc: Should there be a similar note about `upkiidx0`?

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+    } FilterFile;
+```
+
+The `version` value MUST be 4. An implementation MAY reject a file whose
+`version` is not 4.
+
+The `reserved0` byte MUST be `0x00`. An implementation MUST reject a file whose
+`reserved0` is not zero. Future versions of the format may use a non-zero value
+to signal the presence of additional metadata (e.g. describing a partition of
+the filter) that is omitted in version 4.
```
djc: What is the difference here between a non-4 version and a non-0 `reserved0`, which get respectively "MAY reject" and "MUST reject"?

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+`index.bin` is the [Index file format](#index-file-format), described below,
+and identifies the filter files by name.
+
+Filter files are ordinary files whose bytes conform to the [Filter file
+format](#filter-file-format). Their names MUST be at most 32
+bytes when encoded as UTF-8 and MUST NOT contain the byte `0x00` for
+compatibility with the index file format. Beyond those restrictions, filenames
+are opaque to this specification.
+
+There MAY be additional metadata files in the cache directory (e.g.
```
djc: I don't see much value in specifically naming `manifest.json` here?

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+`manifest.json`) that describes when the filters were fetched, their expected
+checksums, etc. These details are outside of the scope of this specification and
+are not required for efficient revocation checking.
+
+Populating the cache directory, obtaining `index.bin` and the filter files
+it names, verifying their integrity, and deciding when and how to refresh them
+is out of scope for this document.
+
```
djc: This is also mentioned in the initial section (before the intro). Is it useful to keep both instances?

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+2. Before an `index.bin` referencing a filter file is added to the cache, the
+   filter file bytes MUST be written to disk first, before any `index.bin` that
+   references it becomes visible at `index.bin`'s path.
+
+3. When a filter file is removed from the cache, its removal MUST NOT be
+   observable until after an `index.bin` that no longer references it has become
+   visible at `index.bin`'s path.
+
+Together, these invariants ensure that any `index.bin` a reader may observe
```
djc: "any `index.bin` a reader" I think can drop the "a"?

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+   referenced filter file was present at that moment.
+
+3. If opening a filter file identified in step 1 of the check procedure fails
+   because the file does not exist, the reader MUST treat this as evidence that
+   the `index.bin` it is using has been superseded by a newer one that no
+   longer references the missing filter. The reader SHOULD retry the check
+   from the beginning, reopening `index.bin` at least once before returning an
+   error. An implementation MAY bound the number of retries to avoid unbounded
+   restart under pathological update patterns.
+
```
djc: Some applications might want to hold all of the data in memory, but I also wonder if there's a middle ground where processes will want to hold only the index in memory. Should we cover if/how that can be supported?

cpu:
> As you can tell, I'm scraping the bottom of the barrel here.

Thanks! Fixed those nits throughout. I think this is good signal we're ready to merge this once djc approves.

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+```
+    struct {
+        opaque name[32];            /* UTF-8, zero-padded to 32 bytes */
+    } FilenameEntry;
+```
+
+The filename table is `32 * num_filenames` bytes long and begins immediately
+after the header, at file offset 14.
+
+Each `name` slot holds a filter file's name as UTF-8, padded with `0x00`
```
djc: I forget if we discussed this previously, should we strict filter file names to be ASCII?

upki-revocation.md
```diff
@@ -0,0 +1,776 @@
...
+can.
+
+An individual filter that yields **not enrolled** indicates that the
+certificate's issuer was not enrolled in the revocation data even though its
+timestamps were covered. Implementations that make no policy distinction between
+the two MAY collapse them, treating it as a **not covered** result. Like **not
+covered**, this should not preclude checking other covering filters.
+
+Because **revoked** takes precedence over every other outcome, an
+implementation MAY interleave steps 2 and 3, returning **revoked** as soon as
```
djc: I found this confusing. I'm not sure "interleaving" steps 2 and 3 is a good way to explain this when the important part is that all of step 1 must be done first and the natural implementation in my mind is to track all three steps in a single loop and only return early for the revoked case?

(The steps themselves are very clearly laid out, BTW.)

djc (approved):
Looks great to me overall, here are some last notes.

In addition to the ordering concerns re the Introduction listed below, I found that there's a bit of bouncing around in the document in general in terms of top-down vs bottom-up?

upki-revocation.md
cpu: My thinking here is that there are extant pre-V4 filter files that an implementation MAY want to accept (e.g. rustls/upki happily consumes the bincode serialized filter files). 

In contrast there aren't any existing V4 filter files that set reserved0 to anything other than 0 so it seems OK to say you MUST reject that. If we decide to make `reserved0` communicate something in the future it will require spec updates (e.g. to change the field name, and indicate with more clarity what the non-zero value means) and we would change the MUST to a MAY/SHOULD/whatever at that time.

upki-revocation.md
cpu: I'm not sure I follow what you're asking 100%, but my inclination is to not try to describe things that aren't implemented today.

upki-revocation.md
cpu: On the one hand I can't think of a strong reason why we would need UTF-8 filenames, on the other I don't know that it adds much complexity in this instance.

 @ctz do you have opinions?

upki-revocation.md
cpu: Replaced throughout.

upki-revocation.md
cpu: I shifted around the order of the paragraphs a little bit and I think it reads better now.

upki-revocation.md
cpu: Added.

upki-revocation.md
cpu: Removed.

upki-revocation.md
cpu: Fair point, removed here.

upki-revocation.md
cpu: Good catch, fixed.

upki-revocation.md
cpu: I reworded this.

cpu:
> In addition to the ordering concerns re the Introduction listed below, I found that there's a bit of bouncing around in the document in general in terms of top-down vs bottom-up?

Fair point. I reordered some sections slightly to hopefully make the narrative a bit more consistent in ordering.

upki-revocation.md
ctz: I have opinions but suspect they're not very good. These are: all strings are utf-8, filenames are strings, systems where these are not true are rubbish (and also ubiquitous!).

Seriously though, I think if we start restricting the bytes in `name` I would prefer to eliminate `/`, `\\` and `:` rather than the top bit.

cpu: >  I think if we start restricting the bytes in name I would prefer to eliminate /, \\ and : rather than the top bit.

I pushed some updated text that indicates the names in the index

> MUST be between 1 and 32 bytes long and MUST consist only of the ASCII characters `A`–`Z`, `a`–`z`, `0`–`9`, hyphen (`-`), period (`.`), and underscore (`_`). The names `.` and `..` MUST NOT be used.

I'll give some time for you folks to object if you dislike that update and will merge before EOD otherwise.

djc (approved):

cpu:
Thanks for all the input everyone! I'll follow the [CONTRIBUTOR.md docs](https://github.com/C2SP/C2SP/blob/main/.github/CONTRIBUTING.md#tagging-new-versions) to tag a v1.0.0, and we'll get a C2SP steward to add ctz to the maintainer team in https://github.com/C2SP/C2SP/issues/303

PR #281, opened 2026-07-04
tlog-mirror: allow timestamp refresh

FiloSottile:
With upload_start = upload_end = mirror checkpoint size.

/cc @C2SP/tlog-mirror

### DIFF
--- a/tlog-mirror.md
+++ b/tlog-mirror.md
@@ -146,8 +146,7 @@ previously valid pending checkpoint. `ticket` is an opaque value from the
 mirror, or the empty string, to help the mirror recover past pending
 checkpoints.
 
-`upload_start` MUST be less or equal to `upload_end`. It MUST also be less or
-equal to the mirror's next entry for the origin.
+`upload_start` MUST be less or equal to `upload_end`.
 
 The request body uploads log entries from the interval
 `[upload_start, upload_end)`. The interval determines a fixed *canonical
@@ -204,13 +203,20 @@ First, the mirror reads `log_origin`, `upload_start`, `upload_end`, and
 * If `log_origin` is not a known log, the mirror MUST respond with a
   "404 Not Found" HTTP status code.
 
-* If `upload_end` is not the tree size of a known pending checkpoint value, or
-  if it is less than the mirror checkpoint's tree size, the mirror MUST respond
-  with a "409 Conflict" HTTP status code.
+* If `upload_end` is neither a known pending checkpoint value nor the mirror
+  checkpoint's tree size, the mirror MUST respond with a "409 Conflict" HTTP
+  status code. As described below, the mirror can maintain state or use `ticket`
+  to determine known pending checkpoint values.
 
-* If `upload_start` is greater than the mirror's next entry, or too far below
-  the mirror's next entry, the mirror MUST also respond with a "409 Conflict"
-  HTTP status code.
+* If `upload_end` is less than the mirror checkpoint's tree size, the mirror
+  MUST respond with a "409 Conflict" HTTP status code.
+
+* If `upload_start` is greater than the mirror's next entry, the mirror MUST
+  respond with a "409 Conflict" HTTP status code.
+
+* Let `excess_entries = min(upload_end, next_entry) - upload_start`, where
+  `next_entry` is the mirror's next entry. If `excess_entries` is too large, the
+  mirror MUST respond with a "409 Conflict" HTTP status code.
 
 The mirror SHOULD send these error responses without waiting for the entire
 request body to be available. Conversely, the client SHOULD be prepared to
@@ -238,10 +244,10 @@ committed partially. The client retries with `upload_start` set to the
 advertised next entry value, repeating until the mirror has received all
 packages.
 
-If no entry package was authenticated and saved before the body ended (for
-example, the request header itself was malformed, or the first package's
-bytes were truncated mid-package), the mirror MUST respond with a
-"400 Bad Request" HTTP status code.
+If the canonical sequence is not empty, and no entry package was authenticated
+and saved before the body ended (for example, the request header itself was
+malformed, or the first package's bytes were truncated mid-package), the mirror
+MUST respond with a "400 Bad Request" HTTP status code.
 
 When sending a "409 Conflict" or "202 Accepted" response, the response body
 MUST have a `Content-Type` of `text/x.tlog.mirror-info` and consist of three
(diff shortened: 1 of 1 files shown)


FiloSottile:
Hmm, this breaks the upload_start = upload_end = 0 bootstrap when the mirror checkpoint is at 0, which is especially annoying if the pending checkpoint is not at 0.

tlog-mirror.md
```diff
@@ -204,6 +204,11 @@ First, the mirror reads `log_origin`, `upload_start`, `upload_end`, and
 * If `log_origin` is not a known log, the mirror MUST respond with a
   "404 Not Found" HTTP status code.
 
+* If `upload_start` and `upload_end` are both equal to the mirror checkpoint's
+  tree size, the mirror MUST directly respond with a "200 Success" HTTP status
+  code and a response body containing cosignatures on the mirror checkpoint, as
+  described below, with a fresh timestamp.
```
davidben: It is a little confusing that we say "directly respond ..." but then write the rest of the text as if they also apply. (But some of it still doesn't apply like `upload_start`.) Perhaps we could unify it with something like...

> * If `upload_end` is neither a known pending checkpoint value nor the mirror checkpoint's tree size, the mirror MUST respond with a "409 Conflict" HTTP status code. As described below, the mirror can maintain state or use `ticket` to determine known pending checkpoint values.
>
> * If `upload_end` is less than the mirror checkpoint's tree size, the mirror MUST respond with a "409 Conflict" HTTP status code.
>
> * If `upload_start` is greater than `upload_end`, or greater than the mirror's next entry, the mirror MUST respond with a "409 Conflict" HTTP status code.
>
> * Let `excess_entries = min(upload_end, next_entry) - upload_start`, where `next_entry` is the mirror's next entry. If `excess_entries` is too large, the mirror MUST respond with a "409 Conflict" HTTP status code.
>
> The mirror SHOULD send these error responses without waiting for the entire request body to be available. Conversely, the client SHOULD be prepared to receive an error response before the request body is fully sent.

This has another side effect, but I think it's a good one. Let's say the mirror is at 100. The pending checkpoint transited through 200 and then 300. The mirror never saw any entries. The 300 client tries to upload entries targeting 300 and we get all the way to, I dunno, 250, but then we stop. Then the 200 client  tries to upload targeting 200. If next entry is 250 and the mirror believes 50 extra entries is too many, the current rules make it impossible to mint checkpoint 200. There are no valid values of `upload_start`.

The unified version generalizes the `upload_start` special case to say that, `upload_start <<< next entry` is OK if `upload_start = upload_end <<< next entry`.  What we're actually controlling for is wasted upload and there's nothing to be uploaded.

tlog-mirror.md
davidben: Think the second sentence here needs to be updated.

davidben (approved):

FiloSottile:
> Hmm, this breaks the upload_start = upload_end = 0 bootstrap when the mirror checkpoint is at 0, which is especially annoying if the pending checkpoint is not at 0.

@davidben I like your version, but it still has this problem, right? I can’t think of a clean way around it.

davidben:
I must be extra thick today... I'm not seeing how it's broken? We have:

* `upload_end` (0) is the mirror checkpoint's tree size (0), so that's OK
* `upload_end` (0) is not less than the mirror checkpoint's tree size (0), so that's OK
* `upload_start` (0) is neither greater than `upload_end` (0) nor `next_entry` (whatever value)
* `excess_entries` is 0 which cannot be too large by any reasonable threshold

FiloSottile:
Right, but then you get a 200, not a 409, and you don't learn the pending checkpoint size and the next entry.

Say mirror checkpoint is 0, next entry is 1000, pending checkpoint is 2000. How do you _start_ an upload from no state? If you send upload_start = upload_end = 0 hoping to get a 409, instead you get a 200.

You can learn the pending checkpoint size using the witness's read endpoint if there is one, and then use upload_start and either succeed or get a 409, but that's not mandated to exist currently.

This rhymes with #221 and #223.

davidben:
Oh hmm. I assumed any client with no state would begin with an `add-checkpoint` because, for all you know, the mirror's pending checkpoint is behind your target position. And then you would learn a legitimate `add-checkpoint` position either because you successfully set it, or because you got an `add-checkpoint` conflict and the conflict told you a position you were happy with. And then you can `add-entries` from there.

Put another way, I was assuming a mirror client's goal isn't "catch the mirror up to whatever it happens to believe is the pending checkpoint" but "catch the mirror up to my concrete position X".

FiloSottile:
Ok so the bootstrap is

1. `add-checkpoint` and get a 200 or 409, learning the pending checkpoint size
2. `add-entries` with upload_start = 0, potentially getting a 409 if the next entry is too far ahead

I'm happy with that answer, I'll remove the upload_start = upload_end = 0 bootstrap hint from the spec in this PR.

davidben:
Ahhh, I see. The interesting thing is learning `next_entry` rather than a pending checkpoint. And thus #221, because you don't have a good way to learn `next_entry` and (2) requires you to have already started streaming in a bunch of data. (Am re-remembering some of this.) Admittedly, #221 wouldn't full solve it because you don't learn it from add-checkpoint 409. But I guess you can always turn an add-checkpoint 409 into an add-checkpoint 200 by requesting a no-op update.

But I think that's fine? If you're bootstrapping, you can always query the mirror checkpoint and start your `next_entry` at that. That is unlikely to be super wasteful.

tlog-mirror.md
FiloSottile: Removed. It's part of processing, not part of the body validity.

FiloSottile:
> I guess you can always turn an add-checkpoint 409 into an add-checkpoint 200 by requesting a no-op update.

That might livelock if you are chasing, right?

I think the guaranteed bootstrap is:
- `add-checkpoint` and get a 200 or 409, giving the pending checkpoint size
- fetch `checkpoint` and get the mirror checkpoint size
- `add-entries` with mirror checkpoint size and pending checkpoint size, handle 409 as needed

It's not the most efficient but whatever, it's simple and it doesn't have to be efficient, it only happens rarely if ever.

Added the following text:

> If a client doesn't have information on the mirror, it MAY initially make an `add-checkpoint` request to obtain a pending checkpoint size and fetch a checkpoint from the monitoring prefix; those can become stale before the `add-entries` request, but are a reasonable starting point for `upload_end` and `upload_start`, respectively.

tlog-mirror.md
```diff
@@ -204,13 +203,20 @@ First, the mirror reads `log_origin`, `upload_start`, `upload_end`, and
...
+  status code. As described below, the mirror can maintain state or use `ticket`
+  to determine known pending checkpoint values.
 
-* If `upload_start` is greater than the mirror's next entry, or too far below
-  the mirror's next entry, the mirror MUST also respond with a "409 Conflict"
-  HTTP status code.
+* If `upload_end` is less than the mirror checkpoint's tree size, the mirror
+  MUST respond with a "409 Conflict" HTTP status code.
+
+* If `upload_start` is greater than `upload_end`, or greater than the mirror's
```
davidben: (My fault) I guess `upload_start > upload_end` is probably a 400 Bad Request, not a 409 Conflict.

davidben (approved):
LGTM

tlog-mirror.md
FiloSottile: Yep, removed that part, it's already a MUST in the body description.

PR #293, opened 2026-07-11
tlog-tiles: remove MUST for Ed25519 signature.

jsha:
As of #237, logs SHOULD use ML-DSA-44 signatures.


FiloSottile:
Is this a dup of #276?

jsha:
Yep! I'll comment on #276.

davidben:
Reopening since I think it's actually preferable to mine given there is text in tlog checkpoint already. @FiloSottile WDYT about just merging this one?

FiloSottile (approved):
Yeah, this text didn't age well, removing it seems like the best solution.

mhutchinson (approved):

PR #326, opened 2026-08-25
tlog-mirror: clarify past pending checkpoints

jsha:
The langauge around past pending checkpoints was a little inconsistent (e.g. "known pending checkpoint"), so I made "past pending checkpoint" a defined term early on and endeavored to use it consistently throughout. I also expanded some of the other defined terms into their own bullet points.

An important functionality change: previously the request validation for add-entries allowed upload_end == mirror checkpoint tree size. Carrying through this special case everywhere made the langauge pretty messy ("The pending checkpoint, or a past pending checkpoint, or the mirror checkpoint"), so I dropped that, but can do some more wordsmithing to add it back - or perhaps add a special case during validation and say "if you send upload_end == mirror checkpoint, you just get the signature right away with no further processing".

### DIFF
--- a/tlog-mirror.md
+++ b/tlog-mirror.md
@@ -73,20 +73,25 @@ prefix MUST include a [cosignature][] from the mirror.
 The mirror update process is designed to be safely interruptible, while avoiding
 large atomic operations. For each origin log, a mirror maintains the following:
 
-* A copy of the log, [pruned][pruning] to its minimum index. The latest
-  checkpoint of this log copy is known as the *mirror checkpoint*.
+* A copy of the log, [pruned][pruning] to its minimum index.
+
+* A *mirror checkpoint* representing the latest state of the log for which the
+  mirror has a full copy.
 
 * A *pending checkpoint*, which is at or ahead of the mirror checkpoint. If
   ahead of the mirror checkpoint, the pending checkpoint describes entries that
   have yet to be incorporated into the mirror.
 
-* A list of *pending entries* that have yet to be incorporated into the mirror
-  checkpoint. The mirror's *next entry* is the log index of the first entry,
-  greater or equal to the minimum index, that is not in either the log or
-  pending entry list.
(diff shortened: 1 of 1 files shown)


FiloSottile:
> An important functionality change: previously the request validation for add-entries allowed upload_end == mirror checkpoint tree size. Carrying through this special case everywhere made the langauge pretty messy ("The pending checkpoint, or a past pending checkpoint, or the mirror checkpoint"), so I dropped that, but can do some more wordsmithing to add it back - or perhaps add a special case during validation and say "if you send upload_end == mirror checkpoint, you just get the signature right away with no further processing".

This is necessary to allow timestamp refresh, see #281.

jsha:
I'm missing some context here. By timestamp refresh, presumably you mean that the client wants the mirror to produce a cosignature with an updated timestamp, even though the tree size hasn't changed? Two questions:

 - Under what conditions does the client need to request that?
 - Why implement that as part of add-entries rather than as a separate endpoint?
 - The mirror isn't required to produce a fresh signature with the current timestamp, right?

FiloSottile:
> By timestamp refresh, presumably you mean that the client wants the mirror to produce a cosignature with an updated timestamp, even though the tree size hasn't changed?

Correct.

> * Under what conditions does the client need to request that?

If it needs to prove to someone the tree size is fresh, e.g. to show a monitor it's not being slow-rolled, or for logs where recency is important, like vindex output logs, where past leaves are stale maps.

> * Why implement that as part of add-entries rather than as a separate endpoint?

We already have this endpoint that returns a signature over the mirror checkpoint, and I think the language https://github.com/C2SP/C2SP/pull/281 landed on works well, and I like how Sunlight's implementation looks. (I haven't reviewed this PR yet.)

> * The mirror isn't required to produce a fresh signature with the current timestamp, right?

Ultimately that's between a mirror and its clients I guess, because it's hard to make precise rules about what counts as now or how much caching is allowed, but I think it's good for the spec to have a way to do it.

PR #327, opened 2026-08-26
tlog-checkpoint: Don't allow duplicates in {key name,key id}

niels-moller:
Duplicate signature lines, i.e., same key name, same key id, same underlying cryptographic key (barring key id collisions), does not add anything useful.

But worse, for cosignatures, such duplicate lines can carry distinct timestamps. By allowing duplicates, any implementation doing useful things with the cosignature timestamps must be prepared to handle the obscure corner case of multiple timestamps by the same key (and with poor guidance from specs on what is the proper behavior). Rejecting duplicates makes things simpler.

The corresponding change has been discussed in some more detail in the Sigsum context, see
https://git.glasklar.is/sigsum/project/documentation/-/blob/main/proposals/2026-05-29-no-duplicate-cosignature-keys.md and
https://git.glasklar.is/sigsum/project/documentation/-/blob/main/archive/2026-06-02--meeting-minutes.md

### DIFF
--- a/tlog-checkpoint.md
+++ b/tlog-checkpoint.md
@@ -80,6 +80,10 @@ any note signature algorithm based on the ecosystem they operate in. Note that
 the ML-DSA-44 cosignature format doesn't sign the extension lines, which SHOULD
 be empty.
 
+There MAY be multiple signature lines with the same key name. However,
+there MUST NOT be multiple signature lines with both the same key
+name and the same key id.
+
 According to the note specification, clients MUST ignore unknown signatures.
 This enables, for example, log key rotation, and witness cosigning.
 


tlog-checkpoint.md
```diff
@@ -80,6 +80,10 @@ any note signature algorithm based on the ecosystem they operate in. Note that
 the ML-DSA-44 cosignature format doesn't sign the extension lines, which SHOULD
 be empty.
 
+There MAY be multiple signature lines with the same key name. However,
+there MUST NOT be multiple signature lines with both the same key
```
niels-moller, line 84: Note that this MUST NOT is intended to apply to parties that produce and publish checkpoints.

Does the spec needs any explicit wording on receiver behavior? My expectation is that a checkpoint parser SHOULD reject input with duplicates, and as far as I remember, the meeting note comment 

    filippo: I think this sounds right, a "should" seems right

referred to such receiver behavior.

PR #346, opened 2026-09-18
Cut the subtree/checkpoint timestamp knot

### LINKED ISSUE #345: tlog-cosignature: Zero timestamps for subtrees are still a little odd
@phbnf and @AlCutter noticed some odd things about subtree signatures. Tugging on that string, there seems to be quite a lot to clean up.

**Background**

To catch everyone up, with the ML-DSA signature scheme, we tried to do a couple things:
* Have a single signature format that met the needs of tlog and MTCs.
* Unify subtree signatures and checkpoint signatures because it's just a special case where `start` is zero.

This leads to a conflict around timestamps, however. Checkpoint signatures need a very strong precondition on timestamps. It's not just that it's the signing time, but also:

> If the timestamp is not zero, it is also a statement that, as of the specified time, this is the largest consistent tree the cosigner has observed for the log.

That is, you're not allowed to go back and sign an smaller checkpoint than the one you've locked onto. Your signer must be monotonic.

This precondition is incompatible with subtrees, where you inherently sign a subtree that is smaller than the largest thing you've seen. The solution was to say the timestamp could be zero to make no statement and, depending on the context, you either require zero or non-zero timestamps. Specifically:

* Checkpoint signatures require a non-zero timestamp because they need the strong precondition.
* Subtree signatures require a zero timestamp because they don't need and cannot provide the strong precondition.

This works just fine for the signature input, but we neglected to think about the representation of the signature itself...

**Issues**

So, first, tlog-cosignature is actually not self-consistent. When describing `timestamped_signature`, we say:

> timestamp is the time at which the cosignature was generated, as a POSIX timestamp. It MUST NOT exceed 2^63 - 1, and verifiers MAY reject cosignatures with timestamps in the future.

But then in the ML-DSA-44 bits we say:

> timestamp is timestamped_signature.timestamp. These two values MAY be zero if the cosigner doesn't make any statement as to the tree being the largest observed at time of signing.

That's probably easy to fix, but it then reveals another mishap: this means that a subtree signature, if represented as a `timestamped_signature` always has a leading 8 zero bytes. This means:

* The `sign-subtree` example in tlog-witness is impossible because that doesn't begin with a bunch of zeros.
* MTC doesn't use those leading 8 bytes. That means when we say they're compatible, that's not quite right.

Now, we _could_ fix this by patching the MTC draft and https://c2sp.org/mtc-log to explicitly say you trim off those 8 bytes. That's easy enough.

But given that the timestamp is _always_ redundant in sign-subtree, should just change sign-subtree to not use timestamped_signature at all? That might reduce the need to patch all the other bits, and remove a source of redundancy to boot.

Thoughts?

/cc @C2SP/tlog-cosignature 
/cc @C2SP/tlog-witness 
/cc @C2SP/mtc-tlog

davidben:
This is one possible way to fix #345. It opts to:

- More explicitly define subtree and checkpoint signatures as closely related, but different animals.

- Reorganize tlog-cosignature to define, semi-ortogonally, the two formats of cosignatures, and the two cosignature algorithms.

- Since a subtree cosignature gains nothing from the timestamp and, in context, you already have to know which you're working with, simply remove the timestamp from subtree cosignatures at all. This shaves 8 bytes, and avoids the need to have an transform when passing those cosignatures from the tlog universe to the MTC universe.

Note this PR makes what is arguably a compatibility break to sign-subtree. sign-subtree no longer returns a fixed 8-byte zero prefix on the signature value. I say arguably because it is not entirely clear to me which was the original definition. The text was a bit self-contradictory.

Thoughts? This is what seems cleanest to me, but it does mean diverging the note signature lines between subtrees and checkpoints. If note text existed for subtrees, we would already have to diverge those anyway, so ISTM this should be fine, but I don't know how signed notes are meant to work. :-)

/cc @C2SP/tlog-cosignature
/cc @C2SP/tlog-witness
/cc @C2SP/mtc-tlog


lukevalenta (approved):
Not a maintainer of this spec, but the new text LGTM.

